PR #1, opened 2018-08-22
initial check-in of basic skeleton/BL

craigmcnally:
See [MODORDERS-21](https://issues.folio.org/projects/MODORDERS/issues/MODORDERS-21) for details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-orders</artifactId>
-  <version>1.0.1-SNAPSHOT</version>
+  <version>1.0.0-SNAPSHOT</version>
 
   <name>Orders Business Logic</name>
   <description>Business logic to manage orders in FOLIO</description>
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-orders</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.0.1-SNAPSHOT</version>
 
   <name>Orders Business Logic</name>
   <description>Business logic to manage orders in FOLIO</description>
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -0,0 +1,133 @@
...
+              futures.add(polFut);
+            } catch (Exception e) {
+              logger.error("Exception calling POST /po_line", e);
+              throw new CompletionException(e);
+            }
+          });
+
+          VertxCompletableFuture.allOf(ctx, futures.toArray(new CompletableFuture[futures.size()]))
+            .thenAccept(v -> {
+              compPO.setPoLines(lines);
```
mreno-EBSCO: We should determine if the order of the returned PO lines, in multi line POs, matters, i.e. should match what is sent in the POST?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -0,0 +1,133 @@
...
+          String poId = po.getId();
+          compPO.setPurchaseOrder(po);
+
+          List<PoLine> lines = new ArrayList<>(compPO.getPoLines().size());
+          List<CompletableFuture<Void>> futures = new ArrayList<>();
+          compPO.getPoLines().forEach(line -> {
+            line.setPurchaseOrderId(poId);
+            try {
+              Buffer polBuf = JsonObject.mapFrom(line).toBuffer();
+              CompletableFuture<Void> polFut = httpClient.request(HttpMethod.POST, polBuf, "/po_line", null)
```
mreno-EBSCO: The Okapi headers need to be passed in here instead of `null`.

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-orders</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.0.1-SNAPSHOT</version>
```
mreno-EBSCO: NEWS.md should be updated to reflect 1.0.1 is the unreleased version.

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
craigmcnally: I'll bring this up at the next meeting w/ Stacks

PR #3, opened 2018-08-31
MODORDERS-24: Add support for composite PO line

mreno-EBSCO:
The composite_purchase_order schema was updated to refer to a new composite_po_line schema that combines the po_line schema and all of its UUID references as schema "refs".

### DIFF
--- a/src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
+++ b/src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
@@ -126,8 +126,6 @@ public class OrdersResourceImplTest {
 
     String body = getMockData(poCreationFailurePath);
 
-    final Async asyncLocal = ctx.async();
-
     final Errors errors = RestAssured
       .given()
         .header(X_OKAPI_URL)
@@ -153,8 +151,6 @@ public class OrdersResourceImplTest {
     ctx.assertNotNull(errors.getErrors().get(0).getParameters().get(0));
     ctx.assertEquals("purchaseOrder.poNumber", errors.getErrors().get(0).getParameters().get(0).getKey());
     ctx.assertEquals("123", errors.getErrors().get(0).getParameters().get(0).getValue());
-
-    asyncLocal.complete();
   }
 
   @Test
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
```diff
@@ -125,23 +126,35 @@ public void testPoCreationFailure(TestContext ctx) throws Exception {
...
-      .header(X_OKAPI_TENANT)
-      .contentType(APPLICATION_JSON)
-      .body(body)
-      .post(rootPath)
-      .then()
-      .contentType(TEXT_PLAIN)
-      .statusCode(400)
-      .extract()
-      .response();
+    final Async asyncLocal = ctx.async();
```
craigmcnally: Is this needed?  If the rest assured call is synchronous (I think they are), then which part of this is asynchronous?

src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
```diff
@@ -125,23 +126,35 @@ public void testPoCreationFailure(TestContext ctx) throws Exception {
...
+        .header(X_OKAPI_TENANT)
+        .contentType(APPLICATION_JSON)
+        .body(body)
+      .when()
+        .post(rootPath)
+      .then()
+        .log()
+          .all()
+        .contentType(APPLICATION_JSON)
+        .statusCode(422)
```
craigmcnally: It's pretty cool that you get the 422/Unprocessible Entity for free... I was unaware of that and was confused when I saw this until I looked at the RMB code.

src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
mreno-EBSCO: Sorry, habit when dealing with vertx tests. :) Agreed, while it doesn't hurt, in this case it is not needed. I'll remove it.

src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
mreno-EBSCO: Yeah, pretty much anything that causes schema validation errors is reported as a 422 by Okapi. Sorry for the confusion.

PR #4, opened 2018-09-07
generate poNumber and polNumber

craigmcnally:
see [MODORDERS-29](https://issues.folio.org/browse/MODORDERS-29)

### PO Number ###
The algorithm I chose is based on a hex representation of milliseconds since an arbitrary epoch (9/1/2018), with 1-2 hex digits at the end (obtained from nanotime) for entropy.  The entire string is upper-cased.

### POL Number ###
Append a dash followed by a sequential number (1-999)

**Example:**

* PONumber: `2274C0AF60`
* POLNumbers:  `2274C0AF60-1`, `2274C0AF60-2`, `2274C0AF60-3`

If we need to change the algorithm later, we can.  I just wanted to get _something_ in place so we can move forward.

### DIFF
--- a/src/main/java/org/folio/rest/impl/PostOrdersHelper.java
+++ b/src/main/java/org/folio/rest/impl/PostOrdersHelper.java
@@ -27,6 +27,10 @@ public class PostOrdersHelper {
 
   private static final Logger logger = Logger.getLogger(PostOrdersHelper.class);
 
+  // epoch time @ 9/1/2018.
+  // if you change this, you run the risk of duplicate poNumbers
+  private static final long PO_NUMBER_EPOCH = 1535774400000L;
+
   private final HttpClientInterface httpClient;
   private final Context ctx;
   private final Handler<AsyncResult<javax.ws.rs.core.Response>> asyncResultHandler;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -133,4 +141,11 @@ public Void handleError(Throwable throwable) {
 
     return null;
   }
+
```
mtraneis: Although there is no hard requirement on the alpha text, do we want to have a consistent alpha-numeric naming convention? Just to keep this in mind because right now you just want to have something.

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -49,13 +53,17 @@ public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okap
         .thenApply(OrdersResourceImpl::verifyAndExtractBody)
         .thenAccept(poBody -> {
           PurchaseOrder po = poBody.mapTo(PurchaseOrder.class);
+          String poNumber = generatePoNumber();
```
mtraneis: This works for the new PO and new PO lines case.  Do we need to take into account existing PO or PO lines, or a combination of both (exiting PO with new PO lines or additional (new) POL's)?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -133,4 +141,11 @@ public Void handleError(Throwable throwable) {
 
     return null;
   }
+
+  public static String generatePoNumber() {
+    return (Long.toHexString(System.currentTimeMillis() - PO_NUMBER_EPOCH) +
+        Long.toHexString(System.nanoTime() % 100))
+          .toUpperCase();
+  }
```
mtraneis: This algorithm works, no need to account for race conditions, but a different algorithm for a particular naming scheme may require race conditions management.

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
craigmcnally: I think we might need to do this later in order to handle the existing order case.  For now I think it's ok to punt on this just so we can move forward.

PR #5, opened 2018-09-12
MODORDERS-27 MODORDERS-28

craigmcnally:
see [MODORDERS-27](https://issues.folio.org/browse/MODORDERS-27) and [MODORDERS-28](https://issues.folio.org/browse/MODORDERS-28)

I actually went as far as implementing this for:

- cost
- details
- eresource
- location
- vendor_details

NOTE:  the POLine <-> SubObject relationships are doubly linked in the schemas, i.e. the POLine references the SubObject by Id and the SubObject references POLine by Id.  For now I just reference the SubObject from POLine.

We can later add additional calls to add the backwards reference.

### DIFF
--- a/src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
+++ b/src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
@@ -107,6 +107,7 @@ public class OrdersResourceImplTest {
     // logger.info(respBody);
 
     JsonObject json = new JsonObject(respBody);
+    logger.info(json.encodePrettily());
 
     String poId = json.getJsonObject("purchase_order").getString("id");
     assertNotNull(poId);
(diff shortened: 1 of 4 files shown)


src/test/java/org/folio/rest/impl/OrdersResourceImplTest.java
```diff
@@ -207,6 +206,11 @@ protected Router defineRoutes() {
       router.route().handler(BodyHandler.create());
       router.route(HttpMethod.POST, "/purchase_order").handler(this::handlePostPurchaseOrder);
       router.route(HttpMethod.POST, "/po_line").handler(this::handlePostPOLine);
+      router.route(HttpMethod.POST, "/details").handler(this::handlePostAssignId);
```
mtraneis: Will these router calls exercise all the helper methods that were defined?

craigmcnally: When we run these tests, they makes calls to the /orders API, providing an X-Okapi-Url header pointing to localhost, where we have this mock server running.  The orders code will run and make calls to ${X-Okapi-Url}/cost, ${X-Okapi-Url}/details, etc.  For now the the mock server just returns enough dummy data to get by (essentially the request body with a populated id/UUID).  Eventually, in the interest of completeness I/we should add some negative test cases here.  

If it makes you feel any better, parts of the exceptional path(s) involved here are shared, e.g. verifyAndExtractBody(...)/handleError(...) and are already covered, but we can and should do better.

craigmcnally: I'll create a JIRA ticket for improving the unit tests, including additional negative test cases

mtraneis: Sounds good!

craigmcnally: FYI https://issues.folio.org/browse/MODORDERS-34

PR #6, opened 2018-09-17
MODORDERS-34

craigmcnally:
Better Unit Tests

* Failure to create one or more POLine sub-objects (e.g. Cost/Details/etc.)
* More assertions... check po_line_number, po_number, etc.


craigmcnally:
Discussed with Martin and we walked through the code together.  We're on the same page now.

PR #9, opened 2018-10-04
MODORDERS-45 Create mock data JSON records

AndreiShumski2:
Added 15 JSON files with sample orders. Records conform to the JSON schemas in folio-org/acq-models.
+ simple readme file


craigmcnally:
Looks like the build failure is caused by a unit test failing... The root cause is related to the adjustment field...

`16:40:08 ERROR GetOrderHelper       Failed to read mock data
java.lang.IllegalArgumentException: Cannot construct instance of `org.folio.rest.jaxrs.model.Adjustment` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('')
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.folio.rest.jaxrs.model.CompositePurchaseOrder["po_lines"]->java.util.ArrayList[0]->org.folio.rest.jaxrs.model.PoLine["adjustment"])
`

craigmcnally:
see https://github.com/folio-org/acq-models/blob/0060041ce5a9eaecde75c8a24d72a80a01cd63ef/composite_po_line.json#L40

I think that's supposed to be an "Adjustment" object, not just a string.

craigmcnally:
Andrei, it looks like it's expecting the data/time format to be:

`2018-06-29T03:00:00.000+0000`
instead of:
`2018-06-29T03:00:00Z`

craigmcnally:
It also looks like the order of the fields isn't exactly the same in all cases causing the assertion to fail.  I'll have to adjust how we compare the expected w/ actual.

craigmcnally:
I fixed the assertions and the date format in `1ab7ef6a-d1d4-4a4f-90a2-882aed18af14.json`.  The build is now passing.  Please updated the date formats in the remaining json files.

src/main/resources/mockdata/README
```diff
@@ -1 +1,15 @@
+00ed10af-fac2-46ad-9f94-a298358646a2.json - A simple placeholder record.
```
craigmcnally: I think you missed the point...  The idea is that since we chose to use the PO UUID for the json file names, they don't really provide any context.  To help clarify what each json file represents, we list the files in this README and add a short description/explanation.  The values in the third row ("Purpose") of the excel spreadsheet should suffice.... e.g. "One-time ebook" or "Ongoing Print Media" etc.

src/main/resources/mockdata/1ab7ef6a-d1d4-4a4f-90a2-882aed18af14.json
```diff
@@ -1,7 +1,154 @@
...
+    "transmission_date": "",
+    "transmission_method": "",
+    "po_workflow_status_id": "Open",
+    "po_receipt_status": "not received",
+    "po_payment_status": ""
   },
-  "po_lines": []
+  "po_lines": [
+    {
+      "id": "",
```
craigmcnally: just assign a random UUID.  I use https://www.uuidgenerator.net/ for things like this.

src/main/resources/mockdata/1ab7ef6a-d1d4-4a4f-90a2-882aed18af14.json
```diff
@@ -1,7 +1,154 @@
...
+      "workflow_status": {
+        "id": null,
+        "code": null,
+        "description": "Approved"
+      },
+      "selector": "Stevens",
+      "requester": "",
+      "account_number": "",
+      "donor": ""
+    }
```
craigmcnally: This is only the first po_line.  Examples 1 and 3 have additional PO Lines defined in the "Multiline Examples" tab.  I was confused by this at first too.

src/main/resources/mockdata/00ed10af-fac2-46ad-9f94-a298358646a2.json
```diff
@@ -0,0 +1,154 @@
...
+        "id": null,
+        "code": null,
+        "description": "Approved"
+      },
+      "selector": "Hartman",
+      "requester": "",
+      "account_number": "",
+      "donor": ""
+    }
+  ]
```
craigmcnally: This is only the first po_line.  See the "Multiline Examples" tab of the spreadsheet.

src/main/resources/mockdata/1ab7ef6a-d1d4-4a4f-90a2-882aed18af14.json
craigmcnally: This goes for all of the "id" fields

src/main/resources/mockdata/1ab7ef6a-d1d4-4a4f-90a2-882aed18af14.json
craigmcnally: So I still only see the one po line here...  This example should have 3.

craigmcnally: My bad, was looking at the wrong file...

PR #12, opened 2018-10-25
MODORDERS-53 Update mock data, code and tests to work with latest schema changes

craigmcnally:
The latest schema changes include pulling the purchase_order fields up to the top-level of composite_purchase_order.

See [MODORDERS-53](https://issues.folio.org/browse/MODORDERS-53) for additional details.

Due to a discrepancy between composite_purchase_order and purchase_order schemas, I added a bit of code to strip off any adjustment fields in orders during placement.  The top/po-level adjustments are calculated based on the po_line adjustment data (summation).  I added a rudimentary impl of this too while I was at it.

*NOTE* the acq-models submodule was updated to point to the MODORDERS-53 branch of that repo.  Once https://github.com/folio-org/acq-models/pull/38 is merged, I'll update it to be the latest commit.  There shouldn't be any schemas differences, just the submodule branch/commit.

### DIFF
--- a/src/main/java/org/folio/rest/impl/PostOrdersHelper.java
+++ b/src/main/java/org/folio/rest/impl/PostOrdersHelper.java
@@ -8,13 +8,13 @@ import java.util.concurrent.CompletionException;
 
 import org.apache.log4j.Logger;
 import org.folio.orders.rest.exceptions.HttpException;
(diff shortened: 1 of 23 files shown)


craigmcnally:
Thanks for the review guys.  Just updated the acq-modules submodule to point to head of master (after the MODORDERS-53 changes to acq-models were merged)

PR #13, opened 2018-11-03
MODORDERS-39 - Implement GET /orders/id

craigmcnally:
Implement the GET /orders/id endpoint.  See [MODORDERS-39](https://issues.folio.org/browse/MODORDERS-39) and [MODORDERS-41](https://issues.folio.org/browse/MODORDERS-41)

Note:  this work also fixes [MODORDERS-54](https://issues.folio.org/browse/MODORDERS-54)


craigmcnally:
This should bring our test coverage back above 80%.  I would like to spend some time improving the unit test assertions, but don't want to hold up getting this in.  If I have time I'll get some of that into this PR.  If not I'll create a follow-up user story

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
```diff
@@ -0,0 +1,250 @@
...
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import me.escoffier.vertx.completablefuture.VertxCompletableFuture;
+
+public class GetOrdersByIdHelper {
+
+  private static final Logger logger = Logger.getLogger(GetOrdersByIdHelper.class);
+
+  public static final String BASE_MOCK_DATA_PATH = "mockdata/";
```
KVupp: Are we mocking data in implementation class?

KVupp: This variable needs to be removed?

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
```diff
@@ -0,0 +1,250 @@
...
+    this.ctx = ctx;
+    this.asyncResultHandler = asyncResultHandler;
+  }
+
+  public CompletableFuture<CompositePurchaseOrder> getOrder(String id, String lang) {
+    CompletableFuture<CompositePurchaseOrder> future = new VertxCompletableFuture<>(ctx);
+
+    getCompositePurchaseOrderById(id, lang)
+      .thenAccept(orders -> {
+        logger.info("Returning mock data: " + JsonObject.mapFrom(orders).encodePrettily());
```
KVupp: The info message must be changed

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
```diff
@@ -0,0 +1,250 @@
...
+    return future;
+  }
+
+  private CompletableFuture<CompositePurchaseOrder> getCompositePurchaseOrderById(String id, String lang) {
+    CompletableFuture<CompositePurchaseOrder> future = new VertxCompletableFuture<>(ctx);
+
+    getPurchaseOrder(id, lang)
+      .thenAccept(po -> {
+        logger.info("got: " + po.encodePrettily());
+        po.remove("adjustment");
```
KVupp: Isn't the adjustment required to be returned for this API?

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
craigmcnally: good call

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
```diff
@@ -0,0 +1,250 @@
...
+  public CompletableFuture<CompositePurchaseOrder> getOrder(String id, String lang) {
+    CompletableFuture<CompositePurchaseOrder> future = new VertxCompletableFuture<>(ctx);
+
+    getCompositePurchaseOrderById(id, lang)
+      .thenAccept(orders -> {
+        logger.info("Returning mock data: " + JsonObject.mapFrom(orders).encodePrettily());
+        future.complete(orders);
+      })
+      .exceptionally(t -> {
+        logger.error("Error getting orders", t);
```
KVupp: probably trivial  but wouldn't it be easier to debug if we specify the error message that  we are trying to get order by id

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
KVupp: don't we need the adjustment at the PO level?

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
craigmcnally: You get that for free if using the default log4j.properties:

`16:29:56 ERROR GetOrdersByIdHelper  Error getting orders
java.util.concurrent.CompletionException: org.folio.orders.rest.exceptions.HttpException: non-existent-po-id`

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -23,7 +31,47 @@ public static String getMockData(String path) throws IOException {
...
+
+    return response.getBody();
+  }
+
+  public static Adjustment calculateAdjustment(List<PoLine> lines) {
+    Adjustment ret = new Adjustment();
+    for (int i = 0; i < lines.size(); i++) {
+      Adjustment a = lines.get(i).getAdjustment();
+      if (a != null) {
+        if (ret == null) {
```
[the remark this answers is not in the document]
craigmcnally: another good catch.

src/main/java/org/folio/rest/impl/GetOrdersByIdHelper.java
```diff
@@ -0,0 +1,250 @@
...
+
+  private CompletableFuture<PoLine> resolvePoLine(JsonObject line) {
+    CompletableFuture<PoLine> future = new VertxCompletableFuture<>(ctx);
+
+    List<CompletableFuture<Void>> futures = new ArrayList<>();
+    futures.add(resolveSubObjIfPresent(line, "adjustment", "/adjustment/"));
+    futures.add(resolveSubObjIfPresent(line, "cost", "/cost/"));
+    futures.add(resolveSubObjIfPresent(line, "details", "/details/"));
+    futures.add(resolveSubObjIfPresent(line, "eresource", "/eresource/"));
+    futures.add(resolveSubObjIfPresent(line, "location", "/location/"));
```
KVupp: Isn't this URL /locations? https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/location.raml , or Am I looking the wrong API?

craigmcnally: nope I think you're correct.  I had a feeling I was going to get burned by the competing plural/singular api conventions eventually

PR #20, opened 2018-11-29
MODORDERS - PUT /orders/id

javalkar-varun:
Jira ticket - https://issues.folio.org/browse/MODORDERS-60

### DIFF
--- a/src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
@@ -30,27 +30,28 @@ public class PutOrdersByIdHelper {
 
   public PutOrdersByIdHelper(HttpClientInterface httpClient, Map<String, String> okapiHeaders,
       Handler<AsyncResult<Response>> asyncResultHandler, Context ctx) {
-    Map<String,String> customHeader=new HashMap<>();
+    Map<String, String> customHeader = new HashMap<>();
     customHeader.put(HttpHeaders.ACCEPT.toString(), "application/json, text/plain");
     httpClient.setDefaultHeaders(customHeader);
-    
+
     this.httpClient = httpClient;
     this.okapiHeaders = okapiHeaders;
     this.ctx = ctx;
     this.asyncResultHandler = asyncResultHandler;
   }
   
-  public CompletableFuture<Void> updateOrder(String id, String lang, CompositePurchaseOrder compPO, Context vertxContext) {
+  public CompletableFuture<Void> updateOrder(String id, String lang, CompositePurchaseOrder compPO,
+      Context vertxContext) {
     CompletableFuture<Void> future = new VertxCompletableFuture<>(ctx);
-    
+
     PostOrdersHelper postHelper = new PostOrdersHelper(httpClient, okapiHeaders, asyncResultHandler, vertxContext);
-    DeleteOrdersByIdHelper delHelper = new DeleteOrdersByIdHelper(httpClient, okapiHeaders, asyncResultHandler, vertxContext);
-    
-    delHelper.deleteOrder(id,lang)
+    DeleteOrdersByIdHelper delHelper = new DeleteOrdersByIdHelper(httpClient, okapiHeaders, asyncResultHandler,
+        vertxContext);
+
+    delHelper.deleteOrder(id, lang)
     .thenRun(() -> {
       try {
         compPO.setId(id);
-        logger.info("POST id is: -------------------------> "+ id);
         postHelper.createPOandPOLines(compPO)
         .thenAccept(withCompPO -> {
           logger.info("Successfully Placed Order: " + JsonObject.mapFrom(withCompPO).encodePrettily());
@@ -63,8 +64,7 @@ public class PutOrdersByIdHelper {
       } catch (Exception e) {
         logger.error(e);
       }
-    })
-    .exceptionally(delHelper::handleError);
+    }).exceptionally(delHelper::handleError);
     return future;
   }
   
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,103 @@
...
+
+    PostOrdersHelper postHelper = new PostOrdersHelper(httpClient, okapiHeaders, asyncResultHandler, vertxContext);
+    DeleteOrdersByIdHelper delHelper = new DeleteOrdersByIdHelper(httpClient, okapiHeaders, asyncResultHandler,
+        vertxContext);
+
+    delHelper.deleteOrder(id, lang)
+    .thenRun(() -> {
+      try {
+        compPO.setId(id);
+        postHelper.createPOandPOLines(compPO)
```
KVupp: I think it would be better to also add the place holders that Post had for "apply funds" and "update inventory" here, so that we need not do extra work when those are implemented

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,103 @@
...
+        logger.error("Failed to create order compPO", e);
+      }
+    }).exceptionally(delHelper::handleError);
+    return future;
+  }
+  
+  public Void handleError(Throwable throwable) {
+    final Future<javax.ws.rs.core.Response> result;
+
+    logger.error("Exception querying for orders", throwable.getCause());
```
KVupp: The error message needs to be changed to reflect the put operation

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
javalkar-varun: Agree

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,115 @@
...
+              javax.ws.rs.core.Response response = PutOrdersByIdResponse.withNoContent();
+              AsyncResult<javax.ws.rs.core.Response> result = Future.succeededFuture(response);
+              asyncResultHandler.handle(result);
+            })
+            .exceptionally(postHelper::handleError);
+          })
+          .exceptionally(postHelper::handleError);
+        })
+        .exceptionally(postHelper::handleError);
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
javalkar-varun: I had it there to catch any exception that happens in between thenAccept() calls. After reading online, it seems that it is handled by exceptionally() in java8. Will remove try-catch

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,113 @@
...
+            postHelper.updateInventory(withFunds)
+            .thenAccept(withInventory -> {
+              
+              logger.info("Successfully Placed Order: " + JsonObject.mapFrom(withCompPO).encodePrettily());
+              httpClient.closeClient();
+              javax.ws.rs.core.Response response = PutOrdersByIdResponse.withNoContent();
+              AsyncResult<javax.ws.rs.core.Response> result = Future.succeededFuture(response);
+              asyncResultHandler.handle(result);
+            })
+            .exceptionally(postHelper::handleError);
```
craigmcnally: you don't want to call the postHelper's handleError method, you want to call this helper's handleError method.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,113 @@
...
+          })
+          .exceptionally(postHelper::handleError);
```
craigmcnally: same as above

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,113 @@
...
+        })
+        .exceptionally(postHelper::handleError);
```
craigmcnally: Same as above

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    })
+    .exceptionally(delHelper::handleError);
```
craigmcnally: you don't want to call the delHelper's handleError method, you want to call this helper's handleError method.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -0,0 +1,113 @@
...
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.json.JsonObject;
+import me.escoffier.vertx.completablefuture.VertxCompletableFuture;
+
+public class PutOrdersByIdHelper {
+  private static final Logger logger = Logger.getLogger(DeleteOrdersByIdHelper.class);
```
craigmcnally: For consistency probably want to use the vertx logger like the other helpers were recently switched to... e.g. `LoggerFactory.getLogger(...)` see: 
https://github.com/folio-org/mod-orders/blob/2d80a99b2696b6a608cc84bfe9d6f57e9dbe614c/src/main/java/org/folio/rest/impl/PostOrdersHelper.java#L30
and
https://github.com/folio-org/mod-orders/blob/2d80a99b2696b6a608cc84bfe9d6f57e9dbe614c/src/main/java/org/folio/rest/impl/PostOrdersHelper.java#L36

javalkar-varun:
@craigmcnally I created a follow-up PR because it gave a bunch RAML updates conflict so this is the continuation PR which was merged - https://github.com/folio-org/mod-orders/pull/24

Closing this unmerged PR

PR #25, opened 2018-12-03
MODORDERS-46: @Validate Annotation

KVupp:
Changes to add @Validate Annotation to the APIs


src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+
+   
+   
+   logger.info("=== Test validation on invalid lang query parameter ===");
+   RestAssured
+   .with()
+     .header(X_OKAPI_URL)
+     .header(X_OKAPI_TENANT)
+     .contentType(APPLICATION_JSON)
+   .get(rootPath+INVALID_LANG)
```
javalkar-varun: Test says `testValidationDelete` but doing .get() ?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+  public void testValidationOnPut() throws Exception {
+    logger.info("=== Test validation Annotation on PUT API ===");
+
+   logger.info("=== Test validation on empty body ===");
+   RestAssured
+      .with()
+        .header(X_OKAPI_URL)
+        .header(X_OKAPI_TENANT)
+        .contentType(APPLICATION_JSON)
+      .post(rootPath)
```
javalkar-varun: Test says `testValidationOnPut` but doing .post() ?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+       .body(containsString("'lang' parameter is incorrect. parameter value {english} is not valid: must match \"[a-zA-Z]{2}\""));
+
+     
+  }
+  
+  @Test
+  public void testValidationOnPut() throws Exception {
+    logger.info("=== Test validation Annotation on PUT API ===");
+
+   logger.info("=== Test validation on empty body ===");
```
javalkar-varun: You are testing for an empty body`{}` but not passing an empty body? I think you have done it correctly on line 524 but not here.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+     .header(X_OKAPI_TENANT)
+   .get(rootPath+INVALID_LIMIT)
+     .then()
+       .statusCode(400)
+       .body(containsString("'limit' parameter is incorrect. parameter value {-1} is not valid: must be greater than or equal to 0"));
+     
+  }
+  
+  @Test
+  public void testValidationOnGetById(TestContext ctx) throws Exception {
```
javalkar-varun: You are not using `TestContext ctx` anywhere. Maybe just remove it?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+   .get(rootPath+INVALID_LANG)
+     .then()
+       .statusCode(400)
+       .body(containsString("'lang' parameter is incorrect. parameter value {english} is not valid: must match \"[a-zA-Z]{2}\""));
+
+     
   }
   
+  @Test
+  public void testValidationDelete(TestContext ctx) throws Exception {
```
javalkar-varun: Same here. Maybe just remove `TestContext ctx`?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -385,8 +392,156 @@ public void putOrdersById(TestContext ctx) throws Exception {
...
+    .body("{}")
+   .post(rootPath+INVALID_LANG)
+     .then()
+       .statusCode(400)
+       .body(containsString("'lang' parameter is incorrect. parameter value {english} is not valid: must match \"[a-zA-Z]{2}\""));
+     
+  }
+  
+  @Test
+  public void testValidationOnGet(TestContext ctx) throws Exception {
```
javalkar-varun: Same comment as below for `TestContext ctx` and all other places.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
KVupp: Thank you, corrected the info, the test was to be tested with no body sent on a post event,

PR #26, opened 2018-12-05
MODORDERS-82-Purchase-Order-set-createdOn-by-system

pautau-s:
- ignoring created date at post and put requests
- added additional get call in putOrders to retrieve createdDate before deleting

### DIFF
--- a/src/test/java/org/folio/rest/impl/OrdersImplTest.java
+++ b/src/test/java/org/folio/rest/impl/OrdersImplTest.java
@@ -123,17 +123,17 @@ public class OrdersImplTest {
 
     final CompositePurchaseOrder resp = RestAssured
       .with()
-      .header(X_OKAPI_URL)
-      .header(X_OKAPI_TENANT)
-      .contentType(APPLICATION_JSON)
-      .body(body)
+        .header(X_OKAPI_URL)
+        .header(X_OKAPI_TENANT)
+        .contentType(APPLICATION_JSON)
+        .body(body)
       .post(rootPath)
-      .then()
-      .contentType(APPLICATION_JSON)
-      .statusCode(201)
-      .extract()
-      .response()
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -113,17 +123,17 @@ public void testListedPrintMonograph(TestContext ctx) throws Exception {
 
     final CompositePurchaseOrder resp = RestAssured
       .with()
-        .header(X_OKAPI_URL)
-        .header(X_OKAPI_TENANT)
-        .contentType(APPLICATION_JSON)
-        .body(body)
+      .header(X_OKAPI_URL)
```
craigmcnally: I know it's a pain but please retain the indentation here..

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -158,17 +168,17 @@ public void testPlaceOrderMinimal(TestContext ctx) throws Exception {
...
```
craigmcnally: same as above... indentation

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -199,20 +209,20 @@ public void testPoCreationFailure(TestContext ctx) throws Exception {
 
     final Errors errors = RestAssured
       .given()
-        .header(X_OKAPI_URL)
-        .header(X_OKAPI_TENANT)
-        .contentType(APPLICATION_JSON)
-        .body(body)
+      .header(X_OKAPI_URL)
+      .header(X_OKAPI_TENANT)
```
craigmcnally: indentation

PR #27, opened 2018-12-05
MODORDERS-82-Purchase-Order-set-createdOn-by-system

pautau-s:
- ignoring created date at post and put requests
- added additional get call in putOrders to retrieve createdDate before deleting


craigmcnally:
Looks like a duplicate of the other PR with the same name... let me know if I'm missing something here...

PR #31, opened 2018-12-06
Modorders 78 Get PO line endpoint implementation

grabsefx:
Implementation of  "GET /orders/id/lines/id" endpoint


src/main/java/org/folio/rest/impl/GetPOLineByIdHelper.java
```diff
@@ -0,0 +1,114 @@
...
+    this.httpClient = httpClient;
+    this.okapiHeaders = okapiHeaders;
+    this.ctx = ctx;
+    this.asyncResultHandler = asyncResultHandler;
+  }
+
+  public CompletableFuture<PoLine> getCompositePOLineByPOLineId(String orderId, String polineId, String lang) {
+    CompletableFuture<PoLine> future = new VertxCompletableFuture<>(ctx);
+
+    HelperUtils.getCompositePoLineById(polineId, lang, httpClient, ctx, okapiHeaders, logger)
```
piotr-kalashuk: Instead of calling `getCompositePoLineById` probably it makes sense to call ` getPoLineById` and validate order id match. So in this case no extra calls will be done to storage for such case.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -720,6 +723,92 @@ public void testDeleteOrderLineByIdNotCorrespondingToOrderId() {
...
 
+  @Test
+  public void testGetOrderLineById(TestContext ctx) {
+    logger.info("=== Test Get Orderline By Id ===");
+
+    String orderId = "d79b0bcc-DcAD-1E4E-Abb7-DbFcaD5BB3bb";
+
+    final CompositePoLine resp = RestAssured
+      .with()
+      .header(X_OKAPI_URL)
```
piotr-kalashuk: Please use indentation to separate logical steps

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -720,6 +723,92 @@ public void testDeleteOrderLineByIdNotCorrespondingToOrderId() {
...
+      .header(X_OKAPI_TENANT)
+      .get(rootPath + "/" + orderId + "/lines/" + ANOTHER_PO_LINE_ID_FOR_SUCCESS_CASE)
```
piotr-kalashuk: Please use `String.format(LINE_BY_ID_PATH, orderId, ANOTHER_PO_LINE_ID_FOR_SUCCESS_CASE)` so the URL path is defined in one place

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -720,6 +723,92 @@ public void testDeleteOrderLineByIdNotCorrespondingToOrderId() {
...
+      .with()
+      .header(X_OKAPI_URL)
+      .header(X_OKAPI_TENANT)
+      .get(rootPath + "/" + orderId + "/lines/" + lineId)
+      .then()
+      .statusCode(404)
+      .extract()
+      .response();
+
+    assertEquals(lineId, resp.getBody().print());
```
piotr-kalashuk: Better to use `.asString()` instead of `.print()`

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -45,12 +47,15 @@
...
-    subObjectApis.put("alerts", "/alerts/");
-    subObjectApis.put("claims", "/claims/");
+    subObjectApis.put("alerts", "/alert/");
+    subObjectApis.put("claims", "/claim/");
+    subObjectApis.put("reporting_codes", "/reporting_code/");
     subObjectApis.put("fund_distribution", "/fund_distribution/");
     subObjectApis.put(PO_LINES, "/po_line/");
   }
 
+  private static final String GET_POLINE_EXCEPTION = "Exception calling GET /po_line/";
```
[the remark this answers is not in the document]
grabsefx: thanks for advice. I will fix it

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -272,6 +278,11 @@ private static double accumulate(Double a, Double b) {
     logger.info("Calling {} {}", operation, url);
 
     try {
+      // TODO: remove this mock after implementing [MODORDERS-79](https://issues.folio.org/browse/MODORDERS-79)
```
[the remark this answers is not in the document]
grabsefx: You are right. This new method will help

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -329,6 +340,27 @@ private static double accumulate(Double a, Double b) {
       logger.error(EXCEPTION_CALLING_ENDPOINT_MSG, e, HttpMethod.GET, endpoint);
       future.completeExceptionally(e);
     }
+
+    return future;
+  }
+
+  public static CompletableFuture<PoLine> getCompositePoLineById(String lineId, String lang, HttpClientInterface httpClient, Context ctx, Map<String, String> okapiHeaders, Logger logger) {
```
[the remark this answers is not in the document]
grabsefx: Already changed it. Thanks

src/main/java/org/folio/rest/impl/GetPOLineByIdHelper.java
grabsefx: Fixed. Thanks for feedback

src/main/java/org/folio/rest/impl/GetPOLineByIdHelper.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+
+public class GetPOLineByIdHelper {
+
+  private static final Logger logger = LoggerFactory.getLogger(GetOrdersByIdHelper.class);
+
+  public static final String MOCK_DATA_PATH = "mockdata/lines/c0d08448-347b-418a-8c2f-5fb50248d67e.json";
+  public static final String ORDER_REFERENCE_ERROR_MESSAGE = "Current order do not referenced to requested PO line";
```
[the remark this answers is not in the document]
grabsefx: Fixed

PR #32, opened 2018-12-06
Modorders 79 delete po line

piotr-kalashuk:
- Implementing delete order line business logic
- Updating order.raml file to return 422 response (validation issue)
- Handling the case when after unsuccessful deletion of the PO line, some sub-objects already deleted but order line is still referring to them. Allow retrieval of such line and subsequent attempt to delete
- Adding test for GET PO by ID simulating the case when after unsuccessful deletion of the PO line, source sub-object already deleted but order line is still referring to it.
- Enable debug logging for unit tests


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -197,7 +187,7 @@ private static double accumulate(Double a, Double b) {
               .thenAccept(lines::add));
           }
 
-          VertxCompletableFuture.allOf(ctx, futures.toArray(new CompletableFuture[futures.size()]))
+          VertxCompletableFuture.allOf(ctx, futures.toArray(new CompletableFuture[0]))
```
craigmcnally: I had to look into why this change was made... just learned something new!  

https://shipilev.net/blog/2016/arrays-wisdom-ancients/

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -267,26 +269,66 @@ private static double accumulate(Double a, Double b) {
...
     } catch (Exception e) {
-      logger.error(String.format("Exception performing http request %s %s %s", operation.toString(), url, e));
+      logger.error("Exception performing http request {} {}", e, operation, url);
       future.completeExceptionally(e);
     }
 
     return future;
   }
+
+  private static CompletableFuture<JsonObject> handleGetRequest(String endpoint, HttpClientInterface httpClient, Context ctx, Map<String, String> okapiHeaders,
```
craigmcnally: good idea.

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -70,10 +76,25 @@ public static JsonObject verifyAndExtractBody(Response response) {
...
+    }
+    return verifyAndExtractBody(response);
+  }
+
   public static Adjustment calculateAdjustment(List<PoLine> lines) {
     Adjustment ret = null;
-    for (int i = 0; i < lines.size(); i++) {
-      Adjustment a = lines.get(i).getAdjustment();
+    for (PoLine line : lines) {
+      Adjustment a = line.getAdjustment();
```
KVupp: shouldn't we check for null here? a NPE be thrown if line is null?

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -70,10 +76,25 @@ public static JsonObject verifyAndExtractBody(Response response) {
...
+  /**
+   * If response http code is {@code 404}, empty {@link JsonObject} is returned. Otherwise verifies if response is successful and extracts body
+   * In case there was failed attempt to delete PO or particular PO line, the sub-objects might be already partially deleted.
+   * This check allows user to retry DELETE operation
+   *
+   * @param response response to verify
+   * @return empty {@link JsonObject} if response http code is {@code 404}, otherwise verifies if response is successful and extracts body
+   */
+  private static JsonObject verifyAndExtractBodyIfFound(Response response) {
+    if (response.getCode() == 404) {
```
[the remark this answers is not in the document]
piotr-kalashuk: Yes, you are right. The DELETE request to storage always return 204. However in case most of the data of the PO line already deleted but something failed, the PO line won't be deleted from storage. And it would still keep references to deleted sub-objects. In such case I believe some popup with error will be shown to the end user on UI. There are 2 cases possible:
1. User just tries to click delete again, the logic will just retrieve PO line again and all DELETE requests would produce 204. We are ok.
2. User decides to refresh the screen and UI app will send GET PO line request to mod-orders. This will result to attempt to GET all the sub-objects and for those already deleted `404` would be returned. In this case logic will not return composite PO line back to client and user won't be able to retry deletion.

So the check above is basically for the second case. Probably the java doc I added is not so good :)

src/main/java/org/folio/orders/utils/HelperUtils.java
piotr-kalashuk: As far as I can see this method is called either from POST or GET order. Looking at the logic of those operations the PO line cannot be added to the list as null. So I guess there is no need to add this check.

PR #40, opened 2018-12-11
MODORDERS 33 - Implemented sub-object creation - fund_distribution

grabsefx:
* Added required permissions for integration with mod-orders-storage module.
* Fixed unit tests according to new endpoint
* **Should be merged together** with https://github.com/folio-org/mod-orders-storage/pull/43


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -48,7 +48,7 @@
     subObjectApis.put("alerts", "/alert/");
     subObjectApis.put("claims", "/claim/");
     subObjectApis.put("reporting_codes", "/reporting_code/");
-    subObjectApis.put("fund_distribution", "/fund_distribution/");
+    subObjectApis.put("fund_distribution", "/orders-storage/fund_distribution/");
```
craigmcnally: I'm not sure this endpoint exists yet... I believe it's currently defind as just `/fund_distribution`.  However, eventually this will be adjusted per [MODORDSTOR-18](https://issues.folio.org/browse/MODORDSTOR-18)

grabsefx: Yes, it is really confusing. But unfortunatly endpoint **/fund_distribution** is already defined in mod-finance-storage. You can find it here https://github.com/folio-org/mod-finance-storage/blob/master/ramls/fund_distribution.raml
So I decided to define endpoint as /orders-storage/fund_distribution. And also implemented it so according to [MODORDSTOR-19](https://issues.folio.org/browse/MODORDSTOR-19)

NEWS.md
```diff
@@ -12,6 +12,7 @@
  * [MODORDERS-45](https://issues.folio.org/browse/MODORDERS-45) 
  * [MODORDERS-37](https://issues.folio.org/browse/MODORDERS-37)
  * [MODORDERS-34](https://issues.folio.org/browse/MODORDERS-34)
+ * [MODORDERS-33](https://issues.folio.org/browse/MODORDERS-33)
```
piotr-kalashuk: Please move this from  `## 1.0.2` section because this is already released

src/main/java/org/folio/orders/utils/HelperUtils.java
craigmcnally: OK, we might want to hold off on merging this until https://github.com/folio-org/mod-orders-storage/pull/43 is merged.  Also, be sure to update the path here once changing the endpoint to be plural... `/orders-storage/fund_distributions`

NEWS.md
grabsefx: fixed

PR #41, opened 2018-12-11
ModORDERS-32 PO_Line SubObject Creation 

KVupp:
Add implementation for
* alerts
* claims
* reporting_codes
* source
* renewal


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -333,4 +333,8 @@ private static double accumulate(Double a, Double b) {
     }
     return future;
   }
+
+  public static Map<String,String> getSubobjectapis() {
```
craigmcnally: Why can't we just access the map directly?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -280,6 +295,114 @@ public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okap
...
+                }
+              }))
+      );
+
+    return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()]))
+        .thenAccept(t -> {
+          line.put(REPORTING_CODES, reportingIds);
+          compPOL.setReportingCodes(reportingCodes);
+        });
+
```
javalkar-varun: Why aren't catching an exception here?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -280,6 +295,114 @@ public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okap
...
+    if(null!=claims)
+     claims
+      .forEach(claimObject -> 
+        futures.add(createSubObjIfPresent(line, claimObject, CLAIMS, HelperUtils.getSubobjectapis().get(CLAIMS))
+              .thenAccept(id -> {
+                if (id != null) {
+                  claimObject.setId(id);
+                  claimsIds.add(id);
+                }
+              }))
```
javalkar-varun: catch exception here?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -280,6 +295,114 @@ public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okap
...
+  
+  private CompletableFuture<Void> createReportingCodes(PoLine compPOL, JsonObject line,
+      List<ReportingCode> reportingCodes) {
+
+    List<String> reportingIds = new ArrayList<>();
+    List<CompletableFuture<Void>> futures = new ArrayList<>();
+    if(null!=reportingCodes)
+      reportingCodes
+      .forEach(reportingObject -> 
+        futures.add(createSubObjIfPresent(line, reportingObject, REPORTING_CODES, HelperUtils.getSubobjectapis().get(REPORTING_CODES))
```
piotr-kalashuk: The endpoints specified in the `HelperUtils` have trailing slash however POST operations are expected against endpoints without it. As far as I know `okapi` on the discovery step will not be able to find module serving such path. Could you please check?

KVupp: Thanks Piotr

piotr-kalashuk:
I've merged this PR to allow to create PR for [MODORDERS-33](https://issues.folio.org/browse/MODORDERS-33) as this refactoring is required to be in master

PR #43, opened 2018-12-13
MODORDERS-91  Settings: Limit number of Purchase Order Lines per purchase order

pautau-s:
- retrieving configs from mod-config
- fail the call to POST /orders if number of polines more than limit
- fail the call to POST /order/{orderId}/lines if order already reached the limit


src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -0,0 +1,262 @@
...
+import org.folio.rest.tools.client.interfaces.HttpClientInterface;
+
+import javax.ws.rs.core.Response;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+
+public class PostOrderLineHelper extends AbstractOrderLineHelper {
```
epam-pavelk: Good idea to move it to move post po line to a separate helper.

PR #45, opened 2018-12-18
MODORDERS-77 PUT order line

piotr-kalashuk:
## Purpose
[MODORDERS-77](https://issues.folio.org/browse/MODORDERS-77) Implement PUT /orders/id/lines/id with new approach trying to improve approach described in the PR #42

## Approach
- Retrieve PO Line data from storage; then
- Validate that PO id of the PO Line from storage corresponds to order id in the path; then
- Depending on the content in the storage the sub-object updates are following:
  - In case the sub-object is available in the storage and the updated PO Line data also contains this, the `PUT` request is being sent to the storage to corresponding endpoint.
  - In case the sub-object is available in the storage but the updated PO Line data does not contain this anymore, the `DELETE` request is being sent to the storage to corresponding endpoint.
  - In case the sub-object is not available in the storage but the updated PO Line data contains this, the `POST` request is being sent to the storage to corresponding endpoint.
- In case there is an error happened on sub-object operation, the flow continues but happened error is remembered. Once PO Line summary is persisted in the storage, the response with `500` status and with set of happened errors as `JSON` body sent to client


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -299,14 +255,19 @@ private static double accumulate(Double a, Double b) {
...
     CompletableFuture<JsonObject> future = new VertxCompletableFuture<>(ctx);
 
     logger.info("Calling {} {}", operation, url);
 
     try {
-      httpClient.request(operation, url, okapiHeaders)
+      httpClient.request(operation, (body != null ? body.toBuffer() : null), url, okapiHeaders)
         // In case there was failed attempt to delete order or particular PO line, the sub-objects might be already partially deleted.
-        .thenApply(HelperUtils::verifyAndExtractBodyIfFound)
+        .thenApply(response -> verifyAndExtractBody(operation, response))
```
craigmcnally: This is fairly minor, but I think this can be a little misleading... if I'm looking at the logs and run into a GET/404, I'm going to see that I got back and empty response, but in reality we probably got back a 404 with a text/plain response.  I think I'd rather see the GET/404 specific logic handled here instead of in a new variant of `verifyAndExtractBody(...)`.  You can then log an appropriate message saying you got back a 404 but you're ignoring it and continuing.

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -110,13 +56,13 @@ public static JsonObject verifyAndExtractBody(Response response) {
...
-   * This check allows user to retrieve order/line again and retry DELETE operation
+   * This check allows user to retrieve order/line again and retry DELETE operation if required
    *
    * @param response response to verify
    * @return empty {@link JsonObject} if response http code is {@code 404}, otherwise calls {@link #verifyAndExtractBody(Response)}
    */
-  private static JsonObject verifyAndExtractBodyIfFound(Response response) {
-    if (response.getCode() == 404) {
+  private static JsonObject verifyAndExtractBody(HttpMethod operation, Response response) {
+    if (response.getCode() == 404 && operation == HttpMethod.GET) {
```
craigmcnally: See my comment in `operateOnSubObj(operation, response)`

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -137,9 +134,6 @@ public Void handleError(Throwable throwable) {
       case 400:
         result = Future.succeededFuture(PostOrdersResponse.respond400WithTextPlain(message));
         break;
-      case 500:
```
craigmcnally: We'll never serve a 500?  What am I missing?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -127,15 +126,19 @@
   private static final String MOCK_DATA_PATH = BASE_MOCK_DATA_PATH + "getOrders.json";
   private static final String PO_LINES_MOCK_DATA_PATH = BASE_MOCK_DATA_PATH + "lines/";
   private static final String COMP_PO_LINES_MOCK_DATA_PATH = BASE_MOCK_DATA_PATH + "compositeLines/";
-  private final String mockDataRootPath = "src/test/resources/";
-  private final String listedPrintMonographPath = mockDataRootPath + "/po_listed_print_monograph.json";
-  private final String minimalOrderPath = mockDataRootPath + "/minimal_order.json";
-  private final String existedOrder = mockDataRootPath + "/existed_order.json";
-  private final String poCreationFailurePath = mockDataRootPath + "/po_creation_failure.json";
-  private final String poLineCreationFailurePath = mockDataRootPath + "/po_line_creation_failure.json";
+  private static final String MOCK_DATA_ROOT_PATH = "src/test/resources/";
```
craigmcnally: thanks for cleaning this up.

src/main/java/org/folio/orders/utils/SubObjects.java
```diff
@@ -0,0 +1,68 @@
...
+
+    SUB_OBJECT_COLLECTION_APIS = Collections.unmodifiableMap(apis);
+    SUB_OBJECT_ITEM_APIS = Collections.unmodifiableMap(
+      apis.entrySet()
+          .stream()
+          .collect(Collectors.toMap(Map.Entry::getKey, v -> v.getValue() + "/"))
+    );
+  }
+
+  public static String collectionPath(String field) {
```
craigmcnally: Sorry to be a PITA... I appreciate what you're doing here, but the use of "collection" and "item" may lead to confusion... Those terms have specific meanings within FOLIO, and misuse has led to confusion in the past.  While it should be clear given the context, we should probably avoid overloading these terms whenever possible....  maybe "resourcesPath" and "resourceByIdPath" ???

src/main/java/org/folio/orders/utils/HelperUtils.java
piotr-kalashuk: Actually I was trying to handle `404` response a few lines below knowing that empty json object is for such case:
```java
if (json != null) {
  if (json.isEmpty()) {
    logger.warn("The {} {} operation completed with empty response body", operation, url);
```

I will try to make it more clear...

src/main/java/org/folio/orders/utils/HelperUtils.java
piotr-kalashuk: The method is removed and the logic moved to `operateOnSubObj`

src/main/java/org/folio/orders/utils/SubObjects.java
piotr-kalashuk: The methods renamed as suggested to avoid confusions...

PR #51, opened 2018-12-21
MODORDERS-69 Create Instance Record in inventory for items ordered th…

epam-pavelk:
## Purpose
The purpose is to add business logic for creating Instance Record (if one does not exist yet) when placing an Order in Open status or updating order's status to Open.

## Approach
- add flag to PostOrderLineHelper's createPoLineWithAutoGeneratedData method to update Inventory (create Instance Record for now) so that caller can control whether Inventory should be updated or not
- the logic first checks if Instance Record already exists by sending GET request to Inventory endpoint and if it does not, creates a new one by sending POST request
- several auxiliary GET calls to Inventory endpoints are made in order to retrieve data required to create an Instance Record

#### TODOS and Open Questions
- Unit tests still need to be updated to cover newly added logic


src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
+          return httpClient.request(HttpMethod.POST, polBuf, resourcesPath(PO_LINES), okapiHeaders)
             .thenApply(HelperUtils::verifyAndExtractBody)
-            .thenAccept(body -> {
+            .thenApply(body -> {
               logger.info("response from /po_line: " + body.encodePrettily());
 
               compPOL.setId(body.getString("id"));
-              future.complete(compPOL);
-            })
-            .exceptionally(t -> {
```
craigmcnally: Don't we still need this exceptionally block?

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
-              future.completeExceptionally(t.getCause());
-              return null;
+              return compPOL;
             });
         } catch (Exception e) {
           logger.error("Exception calling POST /po_line", e);
-          future.completeExceptionally(e.getCause());
+          throw new CompletionException(e);
         }
+      });
```
craigmcnally: I think we might need an exceptionally(...) block here too.

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
+    return handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger);
+  }
+
+  private CompletableFuture<JsonObject> getStatus(String statusCode) {
+    String endpoint = String.format("/instance-statuses?query=code=%s", statusCode);
+    return handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger);
+  }
+
+  private String buildProductIdQuery(ProductId productId, Map<String, String> productTypes) {
+    return String.format("(identifiers adj \"\\\"identifierTypeId\\\": \\\"%s\\\"\" " +
```
craigmcnally: I think for identifiers and types we should be using "==" instead of "adj".  We want exact matches in both cases

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
+      });
+  }
+
+  private CompletableFuture<JsonObject> getInstanceType(String typeName) {
+    String endpoint = String.format("/instance-types?query=code=%s", typeName);
+    return handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger);
+  }
+
+  private CompletableFuture<JsonObject> getStatus(String statusCode) {
+    String endpoint = String.format("/instance-statuses?query=code=%s", statusCode);
```
craigmcnally: "==" instead of "=" ?

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
+              return location.substring(location.lastIndexOf('/'));
+            });
+        } catch (Exception e) {
+          throw new CompletionException(e);
+        }
+      });
+  }
+
+  private CompletableFuture<JsonObject> getInstanceType(String typeName) {
+    String endpoint = String.format("/instance-types?query=code=%s", typeName);
```
craigmcnally: "==" instead of "=" ?

src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -78,13 +79,14 @@ public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okap
...
           List<CompletableFuture<Void>> futures = new ArrayList<>();
+          boolean updateInventory = compPO.getWorkflowStatus() == WorkflowStatus.OPEN;
           for (int i = 0; i < compPO.getPoLines().size(); i++) {
             PoLine compPOL = compPO.getPoLines().get(i);
             compPOL.setPurchaseOrderId(poId);
             compPOL.setPoLineNumber(poNumber + "-" + (i + 1));
 
             futures.add(new PostOrderLineHelper(httpClient, okapiHeaders, asyncResultHandler, ctx)
-              .createPoLineWithAutoGeneratedData(compPOL).thenAccept(lines::add));
+              .createPoLineWithAutoGeneratedData(compPOL, updateInventory).thenAccept(lines::add));
```
craigmcnally: can we move the .thenAccept(...) to the next line for consistency?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -1512,6 +1521,68 @@ Router defineRoutes() {
       return router;
     }
 
+    private void handlePostInstanceRecord(RoutingContext ctx) {
+      logger.info("got: " + ctx.getBodyAsString());
+      /*JsonObject body = ctx.getBodyAsJson();
```
craigmcnally: remove this or explain in the comment block why this is here

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -56,34 +86,179 @@
...
+   * @param compPOL the PO line to retrieve product type details for
+   * @return product types map
+   */
+  private CompletableFuture<Map<String, String>> getProductTypesMap(PoLine compPOL) {
+    // do not fail if no productId is provided, should be enforced on schema level if it's required
+    if (compPOL.getDetails() == null || compPOL.getDetails().getProductIds().isEmpty()) {
+      return completedFuture(Collections.emptyMap());
+    }
+
+    String endpoint = compPOL.getDetails().getProductIds().stream()
```
craigmcnally: Damn... I meant to bring this up sooner.  I think we're eventually going to want to store the identifier UUID and not have our own productIdType enumeration which is redundant and will inevitably become confusing at some point.  For now I think what you have here is fine, but we'll have to change this later on...  I think ideally the UI would get the list of available identifierTypes and populate a drop-down.  When making the API call, it would use the UUID, not the name/value.  This is similar to how we plan on handling things like locations, assignedTo, etc.

epam-pavelk: Yes, we actually discussed this earlier and decided to not change schema in scope of this story and make this change separately. But in short term, implement the logic above.

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
epam-pavelk: Agree, changed to ==

src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
epam-pavelk: Agree, changed to ==

epam-pavelk:
> Approving but please update ModuleDescriptor-template.json defining required interface and permission dependencies.
Good catch, thanks. The Module descriptor is updated.

> Please take a look at your asynchronous exception handling and CQL queries.
Asynch exception handling is clarified. Please let me no if you still have questions/concerns
CQL queries are fixed.

PR #53, opened 2018-12-27
MODORDSTOR-35 get purchase_order and po_line do not have created by in response

pautau-s:
## Purpose
[MODORDSTOR-35](https://issues.folio.org/browse/MODORDSTOR-35) fix bug with missing created_by in GET /purchase-order and /po_lines responses

## Approach
Use the [metadata.schema](https://github.com/folio-org/raml/blob/master/schemas/metadata.schema) instead of the created and created_by fields.
Changes:
- updated acq-models
- removed created and created_by processing from helpers
- removed test related to created and created_by
- updated samples


src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -48,26 +46,23 @@
...
+            compPOL.setInstanceId(instanceId);
+          }
+          return createPoLine(compPOL);
+        });
     }
     return createPoLine(compPOL);
   }
 
   private CompletableFuture<PoLine> createPoLine(PoLine compPOL) {
-    return createPoLine(compPOL, null);
```
epam-pavelk: I would probably leave these piece, so that: 
- all PO line creation logic is in one place (createPoLine method), rather than InstanceId set in one place and the rest of data in another place
- make the calling logic cleaner by not requiring manipulations with InstanceId, e.g.:
    ...
    .thenCompose(instanceId -> createPoLine(compPOL, instanceId));
}
return createPoLine(compPOL);

PR #55, opened 2019-01-04
Modorders 89- /orders/po_number/validate endpoint

KVupp:
https://issues.folio.org/browse/MODORDERS-89
## Purpose
Adding a new end point to validate a PO Number to ensure a user entered PO Number is unique and matches the pattern 5-16 alphanumeric characters

## Approach
Using the RMB specified @validate annotation to ensure it follows the pattern specified in the acq models
A call to the /orders-storage/purchase-orders endpoint to check if the entered PO number is distinct.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -231,6 +231,14 @@
             "orders-storage.sources.item.delete",
             "orders-storage.vendor_details.item.delete"
           ]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/orders/po_number/validate",
+          "permissionsRequired": ["orders.po_number.item.post"],
+          "modulePermissions": [
+            "orders-storage.purchase_orders.item.get"
+          ]
         }
       ]
     }
@@ -351,6 +359,11 @@
       "description": "Delete an existing PO line"
     },
     {
+      "permissionName": "orders.po_number.item.post",
+      "displayName": "Orders - validate a PO Number",
+      "description": "Validate a PO Number"
+    },
+    {
       "permissionName": "orders.all",
       "displayName": "orders - all permissions",
       "description": "Entire set of permissions needed to use orders",
@@ -362,7 +375,8 @@
         "orders.po_lines.item.post",
         "orders.po_lines.item.get",
         "orders.po_lines.item.put",
-        "orders.po_lines.item.delete"
+        "orders.po_lines.item.delete",
+        "orders.po_number.item.post"
       ]
     }
   ],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -234,8 +234,8 @@
         },
         {
           "methods": ["POST"],
-          "pathPattern": "/orders/po_number/validate",
-          "permissionsRequired": ["orders.po_number.item.post"],
+          "pathPattern": "/orders/po-number/validate",
+          "permissionsRequired": ["orders.po-number.item.post"],
           "modulePermissions": [
             "orders-storage.purchase_orders.item.get"
           ]
@@ -359,7 +359,7 @@
       "description": "Delete an existing PO line"
     },
     {
-      "permissionName": "orders.po_number.item.post",
+      "permissionName": "orders.po-number.item.post",
       "displayName": "Orders - validate a PO Number",
       "description": "Validate a PO Number"
     },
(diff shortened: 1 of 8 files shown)


ramls/order.raml
```diff
@@ -51,7 +52,33 @@ resourceTypes:
...
+        description: validate if the PO Number is unique and matches the pattern specified
+        is: [validate, auth]
+        body:
+          application/json:
+            type: po_number
+        responses:
+          204:
+            description: "Valid PO_Number"
+          400:
+            description: "Bad request, e.g. malformed request body or query parameter. Details of the error (e.g. name of the parameter or line/character number with malformed data) provided in the response."
```
craigmcnally: Maybe update this description as it seems to be a little misleading to me.  I think a malformed request would actually result in a 422 most of the time.  A 400 would occur when the po_number provided already exists.

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -29,12 +30,14 @@
...
 public class HelperUtils {
 
+  private static final Pattern PONUMBER_PATTERN = Pattern.compile("^[a-zA-Z0-9]{5,16}$");
   public static final String DEFAULT_POLINE_LIMIT = "500";
   public static final String OKAPI_URL = "X-Okapi-Url";
   public static final String PO_LINES_LIMIT_PROPERTY = "poLines-limit";
   public static final String URL_WITH_LANG_PARAM = "%s?lang=%s";
   public static final String GET_ALL_POLINES_QUERY_WITH_LIMIT = resourcesPath(PO_LINES)+"?limit=%s&query=purchase_order_id==%s&lang=%s";
-  public static final String GET_ALL_PURCHASE_ORDER_QUERY = resourceByIdPath(PURCHASE_ORDER)+URL_WITH_LANG_PARAM;
+  public static final String GET_PURCHASE_ORDER_BYID = resourceByIdPath(PURCHASE_ORDER)+URL_WITH_LANG_PARAM;
```
javalkar-varun: Where is `PURCHASE_ORDER` being defined or how will it get its value?

KVupp: It is defined in the SubObjects.java. It is accessed via a static import

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -222,6 +225,18 @@ public void putOrdersLinesByIdAndLineId(String orderId, String lineId, String la
     }
   }
 
+  @Override
+  @Validate
```
piotr-kalashuk: The `po_number.json` schema defines `"pattern": "^[a-zA-Z0-9]{5,16}$"` and `@Validate` annotation should let the RMB to validate `PoNumber` content. Looks like `HelperUtils.isPOValid` is not required in this case...

KVupp: Yea that method is not being used anywhere currently, I just left it in case if has to be used by the other APIs

ramls/order.raml
```diff
@@ -51,7 +52,29 @@ resourceTypes:
         example:
           strict: false
           value: !include acq-models/examples/composite_purchase_order.sample
-
+  /po_number/validate:
```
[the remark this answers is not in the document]
KVupp: Makes sense, Changes are in

descriptors/ModuleDescriptor-template.json
```diff
@@ -350,6 +358,11 @@
       "displayName": "Orders - delete an existing PO line",
       "description": "Delete an existing PO line"
     },
+    {
+      "permissionName": "orders.po_number.item.post",
```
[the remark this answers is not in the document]
KVupp: Makes sense, Changes are in

ramls/order.raml
KVupp: Added valid description

descriptors/ModuleDescriptor-template.json
```diff
@@ -231,6 +231,14 @@
             "orders-storage.sources.item.delete",
             "orders-storage.vendor_details.item.delete"
           ]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/orders/po-number/validate",
+          "permissionsRequired": ["orders.po-number.item.post"],
+          "modulePermissions": [
+            "orders-storage.purchase_orders.item.get"
```
piotr-kalashuk: Please change this permission to `orders-storage.purchase_orders.collection.get`

KVupp: Good find. Thank you

PR #56, opened 2019-01-04
Modorders 94 improve put orders endpoint

pautau-s:
## Purpose
[MODORDERS-94](https://issues.folio.org/browse/MODORDERS-94) Update PUT /orders/id with new approach

Extra work:
- Updated module descriptor to fix bug with token length on PUT orders/id
- If the poLines limit is exceeded, then return the specific code required for UI

## Approach
- The updated approach is following:
   - if the provided composite_purchase_order doesn't contain the "po_lines" field, just update the purchase_order information via the /orders-storage/purchase_orders endpoint.
   - if the "po_lines" field is present:
       - Get the current list of po_lines for this order
       - Compare the provided list of po_lines with what's in the database..
           - if the po_line already exists, update it and any sub-objects (eventually most of these will be collapsed w/ po_line)
           - else if the po_line exists in the database, but not in the request, remove the po_line
           - else (the po_line exists in the request but not in the database), create the po_line.


src/main/java/org/folio/rest/impl/PostOrderLineHelper.java
```diff
@@ -533,6 +561,9 @@ private JsonObject buildInstanceRecordJsonObject(PoLine compPOL, Map<String, Str
   Response buildErrorResponse(int code, String message) {
     final Response result;
     switch (code) {
+      case LIMIT_INTERNAL_HTTP_CODE:
```
craigmcnally: I realized this wasn't introduced in this PR, and might not even be code that you wrote, but the strategy of defining and passing along these codes seems a little kludgy to me.  Shouldn't we be creating Exception classes, e.g. "InvalidLimitException", or "LimitExceededException", and then just passing the Exception to the buildErrorResponse method?

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -4,109 +4,151 @@
...
+    List<CompletableFuture<?>> futures = new ArrayList<>();
+    List<PoLine> linesToCreate = new ArrayList<>();
+    if (poLinesFromStorage.isEmpty()) {
+      linesToCreate.addAll(compOrder.getPoLines());
+    } else {
+      futures.addAll(filterPoLinesAndUpdateMatching(compOrder, poLinesFromStorage, linesToCreate));
+    }
+    poLinesFromStorage.stream().forEach(poLine ->futures.add(deletePoLine((JsonObject) poLine, httpClient, ctx, okapiHeaders, logger)));
+    boolean updateInventory = compOrder.getWorkflowStatus() == CompositePurchaseOrder.WorkflowStatus.OPEN;
+    linesToCreate.forEach(poLine -> futures.add(new PostOrderLineHelper(httpClient, okapiHeaders, asyncResultHandler, ctx)
```
piotr-kalashuk: Probably for consistency it makes sense to create instance of the `PostOrderLineHelper` in the constructor like other helpers...

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -4,109 +4,151 @@
...
+          poLinesFromStorage.remove(i);
           break;
-        default:
-          result = Future.succeededFuture(PutOrdersByIdResponse.respond500WithTextPlain(message));
+        }
+      }
+      if (existingLine == null ) {
+        linesToCreate.add(line);
+      } else {
+        futures.add(putLineHelper.updateOrderLine(line, JsonObject.mapFrom(existingLine)));
```
piotr-kalashuk: As far as I can see `existingLine` is `JsonObject`. Is `JsonObject.mapFrom` still required?

src/main/java/org/folio/rest/impl/DeleteOrdersByIdHelper.java
```diff
@@ -1,48 +1,27 @@
...
       Handler<AsyncResult<Response>> asyncResultHandler, Context ctx) {
-    Map<String, String> customHeader = new HashMap<>();
-    customHeader.put(HttpHeaders.ACCEPT.toString(), "application/json, text/plain");
-    httpClient.setDefaultHeaders(customHeader);
-
-    this.httpClient = httpClient;
-    this.okapiHeaders = okapiHeaders;
-    this.ctx = ctx;
-    this.asyncResultHandler = asyncResultHandler;
+    super(AbstractHelper.getHttpClient(okapiHeaders), okapiHeaders, asyncResultHandler, ctx);
```
epam-pavelk: same as above

PR #61, opened 2019-01-14
MODORDERS-72 Receiving endpoints

piotr-kalashuk:
## Purpose
[MODORDERS-72](https://issues.folio.org/browse/MODORDERS-72) Define receiving endpoints

## Approach
- Updating to latest acq-models
- Adding receiving endpoints
  - `POST /orders/receive` to submit `receivingCollection`. The `receivingResults` response is expected in the success case
  - `POST /orders/check-in` to submit `checkinCollection`. The `receivingResults` response is expected in the success case
  - `GET /orders/receiving-history`. The `receiving_history_collection` response is expected in the success case. Once agreed what kind of extra data is required, the new schema will be defined
- Adding `post-json-body-200.raml` which accepts `application/json` body and on success returns `200` with `application/json` to be used for `/orders/receive` and `/orders/check-in`


ramls/order.raml
```diff
@@ -33,16 +38,18 @@ traits:
 
 resourceTypes:
   collection-item: !include raml-util/rtypes/item-collection.raml
+  collection-get: !include raml-util/rtypes/collection-get.raml
   # The post-empty-body.raml does not declare request body but it can be defined when this resource type is used
   post-item: !include raml-util/rtypes/post-empty-body.raml
+  post-with-200: !include rtypes/post-json-body-200.raml
```
epam-pavelk: minor proposal... I would name the raml and type simply "post-body.raml"

PR #62, opened 2019-01-14
Modorders 109 update put /orders/id

pautau-s:
## Purpose
[MODORDERS-96](https://issues.folio.org/browse/MODORDERS-96) Update Put  /orders/<id>

## Approach

- Validate if the resulting PO number is unique on PUT order operation
- Update PO lines numbers once user edits PO number of the pending order.

### DIFF
--- a/src/main/java/org/folio/rest/impl/OrdersImpl.java
+++ b/src/main/java/org/folio/rest/impl/OrdersImpl.java
@@ -82,35 +82,32 @@ public class OrdersImpl implements Orders {
     final HttpClientInterface httpClient = AbstractHelper.getHttpClient(okapiHeaders);
     PostOrdersHelper helper = new PostOrdersHelper(httpClient, okapiHeaders, asyncResultHandler, vertxContext, lang);
     loadConfiguration(okapiHeaders, vertxContext, logger).thenAccept(config -> {
-      int limit = getPoLineLimit(config);
-      if (compPO.getPoLines().size() <= limit) {
-        logger.info("Creating PO and POLines...");
-        helper.createPurchaseOrder(compPO)
-          .thenAccept(withIds -> {
+      validatePoLinesQuantity(compPO, config);
+      logger.info("Creating PO and POLines...");
+      helper.createPurchaseOrder(compPO)
+        .thenAccept(withIds -> {
 
-            logger.info("Applying Funds...");
-            helper.applyFunds(withIds)
-              .thenAccept(withFunds -> {
+          logger.info("Applying Funds...");
+          helper.applyFunds(withIds)
+            .thenAccept(withFunds -> {
 
-                logger.info("Updating Inventory...");
-                helper.updateInventory(withFunds)
-                  .thenAccept(withInventory -> {
+              logger.info("Updating Inventory...");
+              helper.updateInventory(withFunds)
+                .thenAccept(withInventory -> {
 
-                    logger.info("Successfully Placed Order: " + JsonObject.mapFrom(withInventory).encodePrettily());
-                    httpClient.closeClient();
-                    javax.ws.rs.core.Response response = PostOrdersResponse.respond201WithApplicationJson(withInventory,
-                      PostOrdersResponse.headersFor201().withLocation(ORDERS_LOCATION_PREFIX + withInventory.getId()));
-                    AsyncResult<javax.ws.rs.core.Response> result = Future.succeededFuture(response);
-                    asyncResultHandler.handle(result);
-                  })
-                  .exceptionally(helper::handleError);
-              })
-              .exceptionally(helper::handleError);
-          })
-          .exceptionally(helper::handleError);
-      } else {
-        throw new ValidationException(String.format(OVER_LIMIT_ERROR_MESSAGE, limit), LINES_LIMIT_ERROR_CODE);
-      }
+                  logger.info("Successfully Placed Order: " + JsonObject.mapFrom(withInventory).encodePrettily());
+                  httpClient.closeClient();
+                  javax.ws.rs.core.Response response = PostOrdersResponse.respond201WithApplicationJson(withInventory,
+                    PostOrdersResponse.headersFor201().withLocation(ORDERS_LOCATION_PREFIX + withInventory.getId()));
+                  AsyncResult<javax.ws.rs.core.Response> result = Future.succeededFuture(response);
+                  asyncResultHandler.handle(result);
+                })
+                .exceptionally(helper::handleError);
+            })
+            .exceptionally(helper::handleError);
+        })
+        .exceptionally(helper::handleError);
+
     }).exceptionally(helper::handleError);
   }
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -51,7 +50,10 @@ public PutOrderLineByIdHelper(Map<String, String> okapiHeaders, Handler<AsyncRes
    */
   public void updateOrderLine(String orderId, PoLine compOrderLine) {
     getPoLineByIdAndValidate(orderId, compOrderLine.getId())
-      .thenCompose(lineFromStorage -> updateOrderLine(compOrderLine, lineFromStorage))
+      .thenCompose(lineFromStorage -> {
+        compOrderLine.setPoLineNumber(lineFromStorage.getString(PO_LINE_NUMBER));
```
epam-pavelk: Please add a comment to clarify that this is to override PO line number in the request with one from the storage, because it's not allowed to change it during PO line update.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -29,37 +8,96 @@
...
                              Handler<AsyncResult<Response>> asyncResultHandler, Context ctx, String lang) {
     super(getHttpClient(okapiHeaders), okapiHeaders, asyncResultHandler, ctx, lang);
     setDefaultHeaders(httpClient);
     postHelper = new PostOrdersHelper(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
     putLineHelper = new PutOrderLineByIdHelper(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
     postOrderLineHelper = new PostOrderLineHelper(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
+    validationHelper = new ValidationHelper(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
+  }
+
+  public void updateOrderWithPoLines(String orderId, CompositePurchaseOrder compPO) {
```
epam-pavelk: I would propose to change the method name to "updateOrder", because it does not always update PO lines. Also it will be more concise.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -29,37 +8,96 @@
...
+    getPurchaseOrderById(orderId, lang, httpClient, ctx, okapiHeaders, logger)
+      .thenAccept(poFromStorage -> {
+        String oldPoNumber = poFromStorage.getString(PO_NUMBER);
+        boolean isPoNumberChanged = !StringUtils.equalsIgnoreCase(oldPoNumber, compPO.getPoNumber());
+        boolean isRequestContainsLines = !CollectionUtils.isEmpty(compPO.getPoLines());
+
+        if (isPoNumberChanged) {
```
epam-pavelk: The logic of this method became a bit complicated because of many 'if' statements. Probably worth thinking how it could be simplified.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -82,7 +120,16 @@ public void updateOrderWithPoLines(String id, CompositePurchaseOrder compPO) {
...
 
-    return future;
+  }
+
+  private CompletableFuture<Void> updatePoLinesNumber(CompositePurchaseOrder compOrder, JsonArray poLinesFromStorage) {
+    List<CompletableFuture<?>> futures = poLinesFromStorage.stream().map(o -> {
+      JsonObject lineFromStorage = (JsonObject) o;
+      lineFromStorage.put(PO_LINE_NUMBER, buildNewPoLineNumber(lineFromStorage, compOrder.getPoNumber()));
+      String endpoint = String.format(URL_WITH_LANG_PARAM, resourceByIdPath(PO_LINES, lineFromStorage.getString(ID)), lang);
+      return operateOnSubObj(HttpMethod.PUT, endpoint, lineFromStorage, httpClient, ctx, okapiHeaders, logger);
```
epam-pavelk: There are 2 methods to update PO line:
- handlePoLines - this method uses PutOrderLineByIdHelper to update PO lines
- updatePoLinesNumber - this method uses badly named operateOnSubObj static method to update PO lines
I would propose to introduce another method in PutOrderLineByIdHelper such as 'updateOrderLineSummary' that updates only PO line summary w/o sub-objects and use this method here.

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
pautau-s: done

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
pautau-s: Renamed

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
pautau-s: created updateOrderLineSummary method in PutOrderLineByIdHelper

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
pautau-s: simplified a bit

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -119,26 +116,27 @@ public void postOrders(String lang, CompositePurchaseOrder compPO, Map<String, S
   public void putOrdersById(String orderId, String lang, CompositePurchaseOrder compPO, Map<String, String> okapiHeaders,
       Handler<AsyncResult<javax.ws.rs.core.Response>> asyncResultHandler, Context vertxContext) {
     PutOrdersByIdHelper putHelper = new PutOrdersByIdHelper(okapiHeaders, asyncResultHandler, vertxContext, lang);
-    if (CollectionUtils.isEmpty(compPO.getPoLines())) {
-      putHelper.updateOrder(orderId, compPO);
+    if (org.apache.commons.lang.StringUtils.isEmpty(compPO.getPoNumber())) {
```
KVupp: This line can be removed as it will not be reachable. PO Number pattern validation will be handled by  @Pattern(regexp = "^[a-zA-Z0-9]{5,16}$") using @validate

pautau-s: Since the number is an optional field, it may not be in the request. In this case, the user can expect the number to be deleted.

KVupp: ah good point , Thanks.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -114,28 +166,31 @@ public void updateOrderWithPoLines(String id, CompositePurchaseOrder compPO) {
...
+  }
+
+  private String buildNewPoLineNumber(JsonObject poLineFromStorage, String poNumber) {
+    String oldPoLineNumber = poLineFromStorage.getString(PO_LINE_NUMBER);
+    Matcher matcher = PO_LINE_NUMBER_PATTERN.matcher(oldPoLineNumber);
+    if (matcher.find()) {
+      return poNumber + matcher.group(2);
     }
-    return futures;
+    return oldPoLineNumber;
```
craigmcnally: This should never really happen, but if it does I'm not sure leaving the poLineNumber as-is (which doesn't match the new PO Number) is the right thing to do.  At the absolute bare minimum we want to log an error.  Ideally we'd simply assign the line a new, valid number using the poNumber and the next poLine sequence number (which isn't implemented yet).  So for now let's just log an error and add a TODO here so we don't lose track of this and can circle back once the POLine sequence/API is in place.

PR #66, opened 2019-01-16
MODORDERS-87 - Create GET /orders/po_number endpoint for generating a po_number

khandramai:
## Purpose

As per [MODORDERS-87](https://issues.folio.org/browse/MODORDERS-87) in order to centralize generation of po_number one need an API endpoint _GET /orders/po_number_ that calls _/orders-storage/po_number_ and returns a generated **po_number**.

## Approach
- order.raml: orders/po_number endpoint introduced.
- ModuleDescriptor-template.json updated.
- "hardcoded" po_number generation removed from PostOrdersHelper.
- ValidationHelper renamed to PoNumberHelper and extended by including logic for calling _/orders-storage/po_number_ and returning generated po_number.
- Unit tests updated (positive + negative scenarious created).


NEWS.md
```diff
@@ -1,4 +1,5 @@
 ## 1.1.0 - unreleased
+ * [MODORDERS-87](https://issues.folio.org/browse/MODORDERS-87) - Implemented GET PO Number (PO Number generation)
```
craigmcnally: This will be in v3.0.0... Also, try to keep these bullets in descending order.

PR #69, opened 2019-01-17
MODORDERS-121 Create Instance Record in inventory when Order status i…

epam-pavelk:
## Purpose
* When Order status is changed to 'Open' corresponding Inventory records (currently only Instance records) must be created and associated with PO lines. 
* The status of the order must be changed to Open at the very end, when the rest is updated (including Inventory records), otherwise leave it in current status. 
As a special case, if Order is created in 'Open' status it should initially be created in Pending status and transitioned to Open only when the rest is created (incl PO lines, Inventory Records)

## Approach
- "update Inventory" logic is moved from PostOrderLineHelper to PutOrderLineByIdHelper, because no Inventory should be created when PO lines is created, but rather during update when status is changed to Open
- PostOrderHelper is updated
-- if Order is created in Open status, it is initially created in Pending status and then transition to Open status is triggered (which assumes to first update Inventory). The status is changed to Open at the last step
-- logic to create PO lines is moved to separate method
- PutOrdersByIdHelper is updated
-- when Order status is changed to Open corresponding Inventory records are created and associated with PO lines
-- the code is refactored for better readability

## Notes
- the changes are still being tested locally, so minor additions may appear if an issue is found


src/main/java/org/folio/rest/impl/PostOrdersHelper.java
```diff
@@ -32,11 +35,13 @@
   private static final long PO_NUMBER_EPOCH = 1535774400000L;
 
   private final ValidationHelper validationHelper;
+  private final PutOrdersByIdHelper putOrderHelper;
 
   public PostOrdersHelper(HttpClientInterface httpClient, Map<String, String> okapiHeaders,
       Handler<AsyncResult<javax.ws.rs.core.Response>> asyncResultHandler, Context ctx, String lang) {
     super(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
     this.validationHelper = new ValidationHelper(httpClient, okapiHeaders, asyncResultHandler, ctx, lang);
+    this.putOrderHelper = new PutOrdersByIdHelper(okapiHeaders, asyncResultHandler, ctx, lang);
```
craigmcnally: It feels a little weird to have a putOrderHelper as a member of the postOrderHelper.  I think the approach we've tried to take in situations like this is to split code that's shared between helpers in a common/shared location, e.g. HelperUtils rather than having the helpers reference one another.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -47,92 +53,122 @@
...
-              })
-              .exceptionally(this::handleError);
-          })
-          .exceptionally(this::handleError);
+    return operateOnSubObj(HttpMethod.PUT, resourceByIdPath(PURCHASE_ORDER, compPO.getId()), purchaseOrder, httpClient, ctx, okapiHeaders, logger)
+      .thenAccept(json -> {});
+  }
+
+  private CompletableFuture<Void> updatePoLinesNumber(CompositePurchaseOrder compOrder, JsonArray poLinesFromStorage) {
+    CompletableFuture[] futures = poLinesFromStorage.stream()
```
craigmcnally: `CompletableFuture<?>[]`

PR #73, opened 2019-01-21
MODORDERS-124 GET orders collection bug

grabsefx:
## Purpose
The purpose is to fix a bug related to not provided endpoint `GET /orders/composite-orders` by module
## Approach
Adding GET /orders/composite-orders to module descriptor, adding required permissions for the endpoint

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -7,6 +7,14 @@
       "version": "2.0",
       "handlers": [
         {
+          "methods": ["GET"],
+          "pathPattern": "/orders/composite-orders",
+          "permissionsRequired": ["orders.collection.get"],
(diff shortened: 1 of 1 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -6,6 +6,14 @@
       "id": "orders",
       "version": "2.0",
       "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/composite-orders",
+          "permissionsRequired": ["orders.collection.get"],
+          "modulePermissions": [
+            "orders-storage.purchase_orders.collection.get"
```
craigmcnally: depending on what happens in https://github.com/folio-org/mod-orders-storage/pull/63 you might need to change this.

PR #77, opened 2019-01-23
MODORDERS-67 Create Item Record in inventory

piotr-kalashuk:
## Purpose
[MODORDERS-67](https://issues.folio.org/browse/MODORDERS-67) Create Item Record in inventory
[MODORDERS-117](https://issues.folio.org/browse/MODORDERS-117) Business Logic: e-Resources, 'Create Item' Flag

## Approach
- creating `org.folio.rest.impl.InventoryHelper` which contains all the logic specific to Inventory interaction
- `org.folio.rest.impl.PutOrderLineByIdHelper` updates:
  - existing methods related to Instance have been moved to `InventoryHelper`.
  - `updateOrderLine(PoLine, JsonObject)` method updated so now it checks if PO Line update successful or not (it guarantees that any client which calls this method will have correct result of the operation without extra checks)
- `HelperUtils#calculateInventoryItemsQuantity(PoLine)` added with the logic described in [MODORDERS-117](https://issues.folio.org/browse/MODORDERS-117)
- The `OrdersImplTest` updated with additional tests related to Inventory Items
- README updated with the summary information about Inventory interaction
- some mock files updated with valid UUID's

## TODO
- [ ] Uncomment lines in code which deals with `purchaseOrderLineIdentifier` property in the item record once the changes of [MODINVSTOR-245](https://issues.folio.org/browse/MODINVSTOR-245) merged to master and updated inventory components available. All such places have TODO comments so they can be easily found
- [ ] Remove TODO comment in the `InventoryHelper#handleHoldingRecord(PoLine)` in scope of [MODORDERS-67](https://issues.folio.org/browse/MODORDERS-67)


src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -0,0 +1,429 @@
...
+    CompletableFuture<List<String>> result = new VertxCompletableFuture<>(ctx);
+    logger.debug("Creating {} items for PO Line with '{}' id", count, compPOL.getId());
+
+    buildItemRecordJsonObject(compPOL, holdingId)
+      .thenCompose(itemData -> createItemRecords(itemData, count))
+      .thenAccept(createdItemIds -> {
+        // In case no items created at all, throw an exception because nothing can be done at this stage
+        if (createdItemIds.isEmpty()) {
+          throw new IllegalStateException(String.format("No items created for PO Line with %s id", compPOL.getId()));
+        } else if (count != createdItemIds.size()) {
```
epam-pavelk: Wouldn't it be better to throw an exception and stop the processing (namely, creating piece records for succeeded items) if not all items are created, than just logging an error and returning succeded ones?
@craigmcnally , what do you think?

src/test/resources/README
```diff
@@ -0,0 +1,18 @@
+Describes files and how they can be used
```
epam-pavelk: Like. I think it will be useful given the number of mock files we've got so far.

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    String endpoint = compPOL.getDetails().getProductIds().stream()
+      .map(productId -> String.format("name==%s", productId.getProductIdType().toString()))
+      .collect(joining(" or ", "/identifier-types?query=", ""));
+
+    return handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger)
+      .thenApply(productTypes -> {
+        if (productTypes.getJsonArray("identifierTypes").size() != compPOL.getDetails().getProductIds().size()) {
+          throw new CompletionException(new HttpException(422,
```
craigmcnally: I personally think this is a misuse of the HttpException class, which was intended to encapsulate an HTTP 4xx/5xx response.  Here, we never got a 422, with this message.  While this may be convenient, it could be confusing to someone debugging a problem in logs, etc.   

I would prefer to see a new Exception class introduced for this specific purpose, e.g. `InvalidProductTypeException`

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -0,0 +1,429 @@
...
+
+  private String buildProductIdQuery(ProductId productId, Map<String, String> productTypes) {
+    return String.format("(identifiers adj \"\\\"identifierTypeId\\\": \\\"%s\\\"\" " +
+        "and identifiers adj \"\\\"value\\\": \\\"%s\\\"\")",
+      productTypes.get(productId.getProductIdType().toString()),
+      productId.getProductId());
+  }
+
+  private JsonObject buildInstanceRecordJsonObject(PoLine compPOL, Map<String, String> productTypes, JsonObject lookupObj) {
+    JsonObject instance = new JsonObject();
```
craigmcnally: It's a bummer we're forced to work with JsonObject here...  -- No action required --

src/main/java/org/folio/rest/impl/InventoryHelper.java
craigmcnally: The operation would need to be retried either way.

That said, on one hand, fast-failing is nice and clean and keeps the code simple.  On the other hand, it's probably cleaner on the storage end to create piece records for the item records that were successfully created.  Is everything in place to fully transition the po_line to Open? No, but at least the records that were created have the appropriate linkage (assuming there aren't also failures while creating the piece records).  

A few other minor factors include:
* the decision we make here could have some affect record metadata...  e.g. creation dates, etc.  I'm not sure it really matters though
* Are the advantages of refactoring this worth the time it's going to take making the changes, getting them reviewed again, holding up other tasks - possibly pushing them into next sprint?

I think I'm in favor of just keeping it as currently implemented.  My decision really comes down to the first and last points I make.

PR #82, opened 2019-01-25
Modorders 99 purchase order limit set system default to 1

pautau-s:
## Purpose
[MODORDERS-99](https://issues.folio.org/browse/MODORDERS-99) Purchase Order Limit: Set system default to 1

## Approach
- hardcoded default limit of po_lines to 1
- updated unit test and mocks in accordance with the new default limit value


src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -843,6 +845,32 @@ public void testPoUpdateWithOverLimitPOLines(TestContext ctx) throws Exception {
     ctx.assertEquals(LINES_LIMIT_ERROR_CODE, errors.getErrors().get(0).getCode());
   }
 
+  @Test
+  public void testPoUpdateWithOverLimitPOLinesWithDefaultLimit(TestContext ctx) throws Exception {
+    logger.info("=== Test PUT PO, with over limit lines quantity with default limit ===");
+
+    String body = getMockData(listedPrintMonographPath);
+    final Errors errors = RestAssured
+      .with()
```
piotr-kalashuk: Better to reformat it to improve readability

PR #88, opened 2019-01-28
Addings holdings creation

KVupp:
https://issues.folio.org/browse/MODORDERS-66
## Purpose
To create a holdings record if one doesn't exist

## Approach
Check for a holdings record based on an instanceId and locationId. If more than one holdings is returned pick the first one. 
If it doesn't exist create a holdings record

#### TODOS and Open Questions
Assuming that inventory will be able to move items from one holding to another. 
This implementation will be further modified to include creation of multiple holding records for multiple locations

### DIFF
--- a/src/main/java/org/folio/rest/impl/InventoryHelper.java
+++ b/src/main/java/org/folio/rest/impl/InventoryHelper.java
@@ -56,6 +56,8 @@ public class InventoryHelper {
   private static final String CREATE_INSTANCE_ENDPOINT = "/inventory/instances?lang=%s";
   private static final String LOOKUP_ITEMS_ENDPOINT = "/item-storage/items?query=purchaseOrderLineIdentifier==%s&limit=%d&lang=%s";
   private static final String CREATE_ITEM_ENDPOINT = "/item-storage/items?lang=%s";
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -77,12 +79,27 @@
...
-      .thenApply(this::extractId);
+          .thenCompose(holdings -> {
+            if (!holdings.getJsonArray(HOLDINGS_RECORDS).isEmpty()) {
+              return completedFuture(extractId(getFirstObjectFromResponse(holdings, HOLDINGS_RECORDS)));
+            }
+            return createHoldingsrecord(compPOL);
+          });
+  }
+
+  private CompletableFuture<String> createHoldingsrecord(PoLine compPOL) {
```
pautau-s: nitpick: : missed camelCase

KVupp:
> Update NEWS.md?

Recently when I release mod-gobi  I came across that, the NEWS.md will have to be updated during the release with major changes. Hence I refrained from modifying it
(https://dev.folio.org/guidelines/release-procedures/)

PR #89, opened 2019-01-28
MODORDERS-100 Create Piece records

javalkar-varun:
## Purpose

https://issues.folio.org/browse/MODORDERS-100 Create Piece records for items quantity when one doesn't exist

## Approach

1. Check that `itemIds` list is not empty
2. Get already existing pieces
3. Check that `total_records` != `expectedItemsQuantity`
4. Filter `itemIds` based on response data
5. Create piece object using `org.folio.rest.acq.model.Piece`
6. Update unit tests


src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -161,14 +162,95 @@ public boolean isUpdateSuccessful() {
...
+		}
+		List<CompletableFuture<Void>> futuresList = new ArrayList<>();
+		String poLineId = compPOL.getId();		
+		String endpoint = String.format(PIECES_ENDPOINT, poLineId);
+		
+		handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger)
+		    .thenAccept(body -> {
+		    	int i=0;
+		      // No Pieces exists
+			    if (body.getInteger("total_records") == 0) {
```
piotr-kalashuk: At the moment there is no filtering logic and basically we cannot rely on the order of the item identifiers (because it is fully depends on Inventory storage implementation). So the logic might always filter items based on existing piece records and create piece records for remaining item id's. Something like:
```java
          PieceCollection pieces = body.mapTo(PieceCollection.class);
          List<String> existingItemIds = pieces.getPieces()
                                               .stream()
                                               .map(Piece::getItemId)
                                               .filter(StringUtils::isNotEmpty)
          itemIds.stream()
                 .filter(id -> !existingItemIds.contains(id))
                 .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
```

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -161,14 +162,95 @@ public boolean isUpdateSuccessful() {
...
+   * Construct Piece object associated with PO Line
+   *
+   * @param compPOL Composite PO line to update Inventory for
+   * @param itemIds List of itemIds that were created
+   * @param index index associated with each Piece to count total Pieces created
+   * @return CompletableFuture with new Piece record.
+   */
+  private CompletableFuture<Void> createPiece(String poLineId, List<String> itemIds, int index) {
+		// Construct Piece object
+		JsonObject pieceObj = new JsonObject();
```
piotr-kalashuk: My personal preference is to use models generated based on json schemas. This allows to avoid mistakes and guaranty that any change in the schema results to compilation error rather then runtime exception. So the change is like following:
```java
    Piece pieceObj = new Piece();
		pieceObj.setPoLineId(poLineId);
		pieceObj.setItemId(itemId);
		pieceObj.setReceivingStatus(Piece.ReceivingStatus.EXPECTED);
...
			operateOnSubObj(HttpMethod.POST, resourcesPath(PIECES), JsonObject.mapFrom(pieceObj), httpClient, ctx, okapiHeaders, logger)
```

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
javalkar-varun: Done. Nice recommendation! Thanks @piotr-kalashuk

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
javalkar-varun: Agree. Have made changes.

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -161,14 +168,91 @@ public boolean isUpdateSuccessful() {
...
+
+			    logger.info("response from GET /pieces: " + body.encodePrettily());
+
+			    CompletableFuture.allOf(futuresList.toArray(new CompletableFuture[0]))
+			    .thenAccept(v -> future.complete(null))
+			    .exceptionally(t -> {
+				    future.completeExceptionally(t);
+				    return null;
+			    });
+		    });
```
piotr-kalashuk: Only more note. The `future` is being created in the beginning and expected to be completed manually. In case `handleGetRequest` completes exceptionally, the `future` will never be completed

PR #90, opened 2019-01-29
Modorders 150 edit po withot status

piotr-kalashuk:
## Purpose
[MODORDERS-150](https://issues.folio.org/browse/MODORDERS-150) - handling the case when the order's workflow status is not set in the storage.

## Approach
- The main issue resolved by [MODORDERS-145](https://issues.folio.org/browse/MODORDERS-145) where latest acq-models schemas added and default value for `workflow_status` set to `Pending` for composite purchase order. In this case if UI does not send workflow status for order, the default one will be always used.
- The second case when some 3rd party service sends requests directly to storage endpoints. In this case there is potentially might happen that workflow status is not defined. Using `Pending` as a default one in the business logic.
- Adding unit test to verify this case

### DIFF
--- a/src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
@@ -109,13 +109,7 @@ public class PutOrdersByIdHelper extends AbstractHelper {
   }
 
   private boolean isTransitionToOpen(CompositePurchaseOrder compPO, JsonObject poFromStorage) {
-    // MODORDERS-150 Workflow status might be null. Use default as 'Pending' in such case
-    WorkflowStatus currentStatus;
-    if (StringUtils.isNotEmpty(poFromStorage.getString("workflow_status"))) {
-      currentStatus = WorkflowStatus.fromValue(poFromStorage.getString("workflow_status"));
-    } else {
-      currentStatus = PENDING;
-    }
+    WorkflowStatus currentStatus = WorkflowStatus.fromValue(poFromStorage.getString("workflow_status"));
     return currentStatus == PENDING && compPO.getWorkflowStatus() == OPEN;
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -109,7 +109,13 @@ public void updateOrder(String orderId, CompositePurchaseOrder compPO) {
   }
 
   private boolean isTransitionToOpen(CompositePurchaseOrder compPO, JsonObject poFromStorage) {
-    WorkflowStatus currentStatus = WorkflowStatus.fromValue(poFromStorage.getString("workflow_status"));
+    // MODORDERS-150 Workflow status might be null. Use default as 'Pending' in such case
+    WorkflowStatus currentStatus;
+    if (StringUtils.isNotEmpty(poFromStorage.getString("workflow_status"))) {
```
pautau-s: Wouldn't it be better to set the default "Pending" for "workflow_status" in the "purchase_order" scheme instead of this check?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -722,6 +724,34 @@ private void verifyPoWithPoLinesUpdate(JsonObject reqData, JsonObject storageDat
     assertEquals(MockServer.serverRqRs.get(PO_LINES, HttpMethod.PUT).size(), sameLinesCount);
   }
 
+  @Test
+  public void testUpdateOrderWithStatus_MODORDERS150() throws Exception {
+    logger.info("=== Test Put Order By Id - MODORDERS-150 ===");
```
KVupp: A description of the test would be easier to debug later here instead of the JIRA ticket number.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
piotr-kalashuk: Thanks! The https://github.com/folio-org/acq-models/pull/85 created.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
piotr-kalashuk: The jira reference replaced by description.

PR #97, opened 2019-02-01
 MODORDERS-149 Add PO Line identifier to item record

piotr-kalashuk:
## Purpose
[MODORDERS-149](https://issues.folio.org/browse/MODORDERS-149) Add PO Line identifier to item record
The https://github.com/folio-org/mod-inventory-storage/pull/240 has been already merged and updated inventory storage is already available in testing

## Approach
- populating `purchaseOrderLineIdentifier` item's property
- required Inventory module permissions added to module descriptor
- update `item-storage`'s interface version to 7.2
- in case number of items created less than expected, create piece records for successfully created items and return exception after this

## Pre-Merge Checklist:

- Does this PR meet or exceed the expected quality standards ([sonar PR#97](https://sonarcloud.io/project/issues?id=org.folio%3Amod-orders&pullRequest=97&resolved=false))?

## API Tests
Order transition from Pending to Open

image

Open Order creation

image

### DIFF
--- a/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
@@ -42,6 +42,7 @@ import java.util.stream.Collectors;
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.lang3.StringUtils;
+import org.folio.orders.rest.exceptions.InventoryException;
 import org.folio.orders.rest.exceptions.ValidationException;
 import org.folio.orders.rest.exceptions.HttpException;
 import org.folio.rest.acq.model.Piece;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -170,13 +171,18 @@ public boolean isUpdateSuccessful() {
...
-        return completedFuture(itemIds);
-      })
-      .thenCompose(itemIds -> createPieces(compPOL, itemIds));
+        int itemsSize = itemIds.size();
+
+        // Create piece records for successfully created items
+        return createPieces(compPOL, itemIds)
+          .thenAccept(v -> {
+            if (itemsSize != expectedItemsQuantity) {
+              String message = String.format("The issue happened creating items for PO Line with '%s' id. Expected %d but %d created",
```
piotr-kalashuk: This block can be covered by unit tests only in scope of [MODORDERS-154](https://issues.folio.org/browse/MODORDERS-154). At the moment all items being created for POL have absolutely the same content. So there is no way to emulate partial items creation. In scope of [MODORDERS-154](https://issues.folio.org/browse/MODORDERS-154) items for POL would have different holding id's so partial items failure will be emulated as well.

craigmcnally: If this is unreachable code, we should consider commenting it out with a TODO and uncommenting it as part of [MODORDERS-154](https://issues.folio.org/browse/MODORDERS-154).  This would satisfy the code coverage requirement w/o compromising functionality

craigmcnally: I must be missing something... if we're testing against mocks, why can't we emulate this?

piotr-kalashuk: The new test added to cover this case.

PR #98, opened 2019-02-04
MODORDERS-146 Return application/json (error schema) for all errors

grabsefx:
## Purpose
https://issues.folio.org/browse/MODORDERS-146
## Approach
* Update raml file to use `application/json` for any error response
* Added a dictionary for keeping error codes with their descriptions
* Update unit test

#### TODOS and Open Questions
* Check if API tests successfully completed.

## Pre-Merge Checklist:


src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -0,0 +1,28 @@
+package org.folio.orders.utils;
+
+public enum ErrorCodes {
+
+  PO_NUMBER_ALREADY_EXISTS("poNumberNotUnique", "PO Number already exists"),
+  UNDEFINED_ERROR("9999", "Undefined error");
```
craigmcnally: Why use an arbitrary numeric code here?

src/main/java/org/folio/orders/rest/exceptions/CustomHttpException.java
```diff
@@ -0,0 +1,23 @@
+package org.folio.orders.rest.exceptions;
+
+import org.folio.orders.utils.ErrorCodes;
+
+public class CustomHttpException extends Exception {
```
craigmcnally: It may be easier to expand HttpException to accept an ErrorCode than create a whole new class, but it's up to you.  

Other things to consider if you decide to keep this class:
* if you're creating a specialized variant of HttpException, why not extend HttpException?
* it's generally good practice to call the super constructor
* it would be nice to provide a toString() implementation

src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -0,0 +1,28 @@
+package org.folio.orders.utils;
+
+public enum ErrorCodes {
+
+  PO_NUMBER_ALREADY_EXISTS("poNumberNotUnique", "PO Number already exists"),
```
craigmcnally: Why stop here?  Are there other error codes that should be added here?

craigmcnally:
Please discuss this with the front-end devs as these might be breaking changes for ui-orders, e.g. if the UI is expecting a plain text message, but gets json...

src/main/java/org/folio/orders/utils/ErrorCodes.java
grabsefx: We discussed it with UI guys, and agreed that the story was created to provide "poNumberNotUnique" code only at this point. Other codes will be added as necessary.

src/main/java/org/folio/orders/utils/ErrorCodes.java
grabsefx: It's just a draft code not used anywhere. Removed it from here until we reach agreement about list of codes with ui guys

ramls/order.raml
```diff
@@ -31,19 +31,12 @@ traits:
...
-        body:
-          text/plain:
-            example: "unable to create <<resourcePathName|!singularize>> -- unauthorized"
 
 resourceTypes:
-  collection: !include raml-util/rtypes/collection.raml
-  collection-item: !include raml-util/rtypes/item-collection.raml
-  collection-get: !include raml-util/rtypes/collection-get.raml
-  post-with-200: !include rtypes/post-json-body-200.raml
+  collection: !include rtypes/collection-with-json-response.raml
```
piotr-kalashuk: The [RMB's RestVerticle](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/RestVerticle.java#L832) returns `application/json` error response only for 422 status. In case `RestVerticle` returns the response without invocation of the module's business logic methods, the `text/plain` will be returned back to the caller.

craigmcnally: @piotr-kalashuk I didn't realize this.  I feel as though this makes the API a little confusing... Now clients will need to know how to handle an application/json response as well as as text/plain response.  I guess it is what it is for now.  I'll take this up with the RMB team

PR #101, opened 2019-02-06
MODORDERS-100 Create piece on the basis of total quantity

javalkar-varun:
https://issues.folio.org/browse/MODORDERS-100

## Purpose

- Create piece records, even if not creating item records

## Approach

- Calculate total quantity using `compPOL.getCost().getQuantityPhysical()` and `compPOL.getCost().getQuantityElectronic()` as total items and then subtract inventory quantity
- Update tests

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -199,19 +200,27 @@ public boolean isUpdateSuccessful() {
...
         // Extract item Id's which already associated with piece records
         List<String> existingItemIds = pieces.getPieces()
             .stream()
             .map(Piece::getItemId)
             .filter(StringUtils::isNotEmpty)
             .collect(Collectors.toList());
-
+        
+        // Create pieces for successfully created item Ids
+        if(!itemIds.isEmpty()) {
```
piotr-kalashuk: This check does not cover P/E Mix order format when some electronic resources are not expected to be registered in the inventory. The items will be created for physical resources only and the logic will not create piece records for electronic resources.
I propose to enhance the logic and calculate sum of `compPOL.getCost().getQuantityPhysical()` and `compPOL.getCost().getQuantityElectronic()` as total items and then subtract inventory quantity `HelperUtils#calculateInventoryItemsQuantity(pol)`
Ideally the logic should prepare POL content setting `compPOL.getCost().setQuantity(..)` in the beginning of the POL processing so this is always up to date...

javalkar-varun: Thanks for your insight @piotr-kalashuk. Just curious, why should we `getQuantityPhysical()` and `getQuantityElectronic()` from `cost`?

piotr-kalashuk: Please have a look at https://issues.folio.org/browse/UIOR-121. The Physical and/or Electronic quantities are specified in Cost details and locations' should be based on these figures.
I just checked and see that there is no `compPOL.getCost().setQuantity(..)` according to schema. So the sum should be used all the time to get overall quantity

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -218,6 +219,11 @@ public boolean isUpdateSuccessful() {
                .filter(id -> !existingItemIds.contains(id))
                .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
 
+        // Create pieces when item record does not exists
```
piotr-kalashuk: The `createPieces` method will be invoked only if any inventory updates required. Please add additional conditions in the begining of the `updateInventory` method to cover this case.

javalkar-varun: Currently, only `updateInventory` is invoking `createPieces` so why do we need a conditional check? No one else is calling `createPieces`?

piotr-kalashuk: I mean this check:
```java
    int expectedItemsQuantity = calculateInventoryItemsQuantity(compPOL);
    if (expectedItemsQuantity == 0) {
      logger.debug("PO Line with '{}' id does not require inventory updates", compPOL.getId());
      return completedFuture(null);
    }
```

If nothing required to be created in inventory, the `createPieces`will not be called at all.

javalkar-varun:
@piotr-kalashuk - This branch was far behind the master and I was behaving funky so I have created a new PR https://github.com/folio-org/mod-orders/pull/103. Thank you for your feedback.

PR #102, opened 2019-02-08
MODORDERS-146 error code is absent in error message (BUG)

grabsefx:
## Purpose
Error code is absent in error response in case of handling HttpException.
(example: POST purchase order with the existing `po_number`)

## Approach
* adding error code for error response mapping
* adding default error code for other exceptions

### DIFF
--- a/src/main/java/org/folio/rest/impl/AbstractHelper.java
+++ b/src/main/java/org/folio/rest/impl/AbstractHelper.java
@@ -93,8 +93,9 @@ public abstract class AbstractHelper {
     final Throwable t = throwable.getCause();
     final String message = t.getMessage();
     if (t instanceof HttpException) {
-      final int code = ((HttpException) t).getCode();
-      result = succeededFuture(buildErrorResponse(code, new Error().withMessage(message)));
+      final int httpCode = ((HttpException) t).getCode();
+      final String errorCode = ((HttpException) t).getErrorCode();
+      result = succeededFuture(buildErrorResponse(httpCode, new Error().withMessage(message).withCode(errorCode)));
     } else if (t instanceof ValidationException) {
       result = succeededFuture(buildErrorResponse(422, ((ValidationException) t).getError()));
     } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/AbstractHelper.java
```diff
@@ -93,8 +93,9 @@ protected Void handleError(Throwable throwable) {
     final Throwable t = throwable.getCause();
     final String message = t.getMessage();
     if (t instanceof HttpException) {
-      final int code = ((HttpException) t).getCode();
-      result = succeededFuture(buildErrorResponse(code, new Error().withMessage(message)));
+      final int httpCode = ((HttpException) t).getCode();
+      final String errorCode = ((HttpException) t).getErrorCode();
+      result = succeededFuture(buildErrorResponse(httpCode, new Error().withMessage(message).withCode(errorCode)));
     } else if (t instanceof ValidationException) {
```
pautau-s: It seems that now it is possible to replace the `ValidationException `with `HttpException`.

craigmcnally:
_In some cases error code is absent in error message_

I've looked at the changes, and at the JIRA.  It's still unclear to me what those some cases are.  Please add details to the JIRA and/or here.

craigmcnally:
Also, code coverage on new code is low (60%)

grabsefx:
> _In some cases error code is absent in error message_
> 
> I've looked at the changes, and at the JIRA. It's still unclear to me what those some cases are. Please add details to the JIRA and/or here.

PR description updated

grabsefx:
> Also, code coverage on new code is low (60%)

Coverage of new code is so low because of adding lines to several `HttpException `constructors that never used. I suppose, these constructors were created previously for future usage, but at this moment they don't used anywhere.

craigmcnally:
> > Also, code coverage on new code is low (60%)
> 
> Coverage on new code is so low because of adding required lines to several `HttpException `constructors that never used. I suppose, these constructors were created previously for future usage, but at this moment they don't used anywhere.

Sorry to be a pain...  Can we just add tests that exercise that code, or remove the unused code?

PR #103, opened 2019-02-11
Create pieces when item record does not exists

javalkar-varun:
https://issues.folio.org/browse/MODORDERS-100

## Purpose

- Create piece records, even if not creating item records

## Approach

- Calculate total quantity using `compPOL.getCost().getQuantityPhysical()` and `compPOL.getCost().getQuantityElectronic()`
- Subtract inventory items from total quantity to create remaining pieces
- Update tests

Extra:
Update `dateOrdered` tests to verify against month and year to avoid time zone issues
 
## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
@@ -207,7 +207,7 @@ public class PutOrderLineByIdHelper extends AbstractHelper {
       .thenAccept(body -> {
         PieceCollection pieces = body.mapTo(PieceCollection.class);
         int remainingPiecesToCreate = 0;
-        int existingPieces = 0;
+        int existingPieces = pieces.getPieces().size();
         
         if(itemIds!=null) {
           // 1. Get list of item Id's which are already associated with piece records
@@ -227,12 +227,10 @@ public class PutOrderLineByIdHelper extends AbstractHelper {
 	        // 4. Calculate count for remaining pieces to create when Physical and Electronic resources exists
 	        // Scenario: PO Line of "P/E Mix" format with 3 Physical resources and 2 Electronic resources with `create_inventory` set to `false`.
 	        // Create 2 piece records for Electronic resources
-          existingPieces = pieces.getPieces().size();
 	        remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - piecesForPhysicalResources.size() - existingPieces);   
         }
         else {
           // Calculate count for remaining pieces to create on the basis of the total quantity of resources(Physical and Electronic) 
-        	existingPieces = pieces.getPieces().size();
         	remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - existingPieces);
         }
         
--- a/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
@@ -202,30 +202,41 @@ public class PutOrderLineByIdHelper extends AbstractHelper {
 		List<CompletableFuture<Void>> futuresList = new ArrayList<>();
 		String poLineId = compPOL.getId();
 		String endpoint = String.format(PIECES_ENDPOINT, poLineId);
-
+		
 		handleGetRequest(endpoint, httpClient, ctx, okapiHeaders, logger)
       .thenAccept(body -> {
         PieceCollection pieces = body.mapTo(PieceCollection.class);
-
+        int remainingPiecesToCreate = 0;
+        
         if(itemIds!=null) {
-          // Extract item Id's which are already associated with piece records
+          // 1. Get list of item Id's which are already associated with piece records
           List<String> existingItemIds = pieces.getPieces()
 																	             .stream()
 																	             .map(Piece::getItemId)
 																	             .filter(StringUtils::isNotEmpty)
 																	             .collect(Collectors.toList());
 
-          // Create piece records for item id's which do not have piece records yet     
-	        itemIds.stream()
-	               .filter(id -> !existingItemIds.contains(id))
-	               .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
+          // 2. Get list of item id's which do not have piece records yet
+          List<String> piecesForPhysicalResources = itemIds.stream()
+          .filter(id -> !existingItemIds.contains(id)).collect(Collectors.toList());
+
+          // 3. Create piece records     
+          piecesForPhysicalResources.stream().forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));  
+	        
+	        // 4. Calculate count for remaining pieces to create when Physical and Electronic resources exists
+	        // Scenario: PO Line of "P/E Mix" format with 3 Physical resources and 2 Electronic resources with `create_inventory` set to `false`.
+	        // Create 2 piece records for Electronic resources
+	        remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - piecesForPhysicalResources.size());   
         }
         else {
-          // Create pieces on the basis of the total quantity of resources
+          // Calculate count for remaining pieces to create on the basis of the total quantity of resources(Physical and Electronic) 
         	int existingPieces = pieces.getPieces().size();
-        	int remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - existingPieces);
-          IntStream.range(0, remainingPiecesToCreate).forEach(i -> futuresList.add(createPiece(poLineId, null)));
+        	remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - existingPieces);
         }
+        
+        // Create pieces for Electronic resources
+        IntStream.range(0, remainingPiecesToCreate).forEach(i -> futuresList.add(createPiece(poLineId, null)));
+        
         allOf(futuresList.toArray(new CompletableFuture[0]))
           .thenAccept(v -> future.complete(null))
           .exceptionally(t -> {
@@ -242,7 +253,7 @@ public class PutOrderLineByIdHelper extends AbstractHelper {
 		return future;
   }
 
-	/**
+	 /**
    * Construct Piece object associated with PO Line and create in the storage
    *
    * @param poLineId PO line identifier
--- a/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
+++ b/src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
@@ -165,9 +165,9 @@ public class PutOrderLineByIdHelper extends AbstractHelper {
     if (expectedItemsQuantity == 0) {
     	// Create pieces if items does not exists
     	return createPieces(compPOL, null)
-    	.thenRun(() -> {
-    		logger.info("Created pieces for PO Line with '{}' id where inventory updates are not required", compPOL.getId());
-    	});
+    	.thenRun(() -> 
+    		logger.info("Create pieces for PO Line with '{}' id where inventory updates are not required", compPOL.getId())
+    	);
     }
     
     return inventoryHelper.handleInstanceRecord(compPOL)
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -162,9 +163,14 @@ public boolean isUpdateSuccessful() {
     // Check if any item should be created
     int expectedItemsQuantity = calculateInventoryItemsQuantity(compPOL);
     if (expectedItemsQuantity == 0) {
-      logger.debug("PO Line with '{}' id does not require inventory updates", compPOL.getId());
-      return completedFuture(null);
+    	// Create pieces if items does not exists
+    	return createPieces(compPOL, null)
+    	.thenRun(() -> {
```
piotr-kalashuk: The result of the `updateInventory(CompositePoLine)` method is `CompletableFuture<Void>`. I believe `thenRun` can be replaced by `thenAccept` in case you would like info log message once all pieces created successfully. In case you prefer to use `thenRun`, `completedFuture(null)` is not required there...

javalkar-varun: I have removed `completedFuture(null)`since `thenRun` does not return anything at the end of the chain

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -1177,7 +1204,15 @@ private void verifyInventoryInteraction(CompositePurchaseOrder reqData, int crea
     // All existing and created items
     List<JsonObject> items = joinExistingAndNewItems();
 
-    assertEquals(createdPieces.size(), items.size());
+    int eQuantity0 = reqData.getCompositePoLines().get(0).getCost().getQuantityElectronic()!=null ? reqData.getCompositePoLines().get(0).getCost().getQuantityElectronic() : 0;
+    int physicalQuantity0 = reqData.getCompositePoLines().get(0).getCost().getQuantityPhysical()!=null ? reqData.getCompositePoLines().get(0).getCost().getQuantityPhysical() : 0;
+    int eQuantity1 = reqData.getCompositePoLines().get(1).getCost().getQuantityElectronic()!=null ? reqData.getCompositePoLines().get(1).getCost().getQuantityElectronic() : 0;
+    int physicalQuantity1 = reqData.getCompositePoLines().get(1).getCost().getQuantityPhysical()!=null ? reqData.getCompositePoLines().get(1).getCost().getQuantityPhysical() : 0;
+
+    int totalQuantity0 = eQuantity0 + physicalQuantity0 + eQuantity1 + physicalQuantity1;
```
KVupp: we could extract the total quantity calculation part to a separate method, as we are using in multiple places

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -213,10 +214,19 @@ public boolean isUpdateSuccessful() {
...
-               .filter(id -> !existingItemIds.contains(id))
-               .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
-
+        if(itemIds!=null) {
+	        itemIds.stream()
+	               .filter(id -> !existingItemIds.contains(id))
+	               .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
+        }
+        
+        // Calculate total quantity and create pieces when item record does not exists
```
KVupp: I see 2 problems here:
1) For a case where we are creating pieces corresponding to itemIds, the below block of code will be executed after already creating the pieces in line 220, so essentially creating twice the pieces. should this be in else block?
2) If we are creating pieces irrespective of inventory interaction, Why do we need to subtract  calculateInventoryItemsQuantity(compPOL) from the total quantity? The total quantity should just be physical+electronic

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -213,10 +214,19 @@ public boolean isUpdateSuccessful() {
             .collect(Collectors.toList());
 
         // Create piece records for item id's which do not have piece records yet
-        itemIds.stream()
-               .filter(id -> !existingItemIds.contains(id))
-               .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
-
+        if(itemIds!=null) {
```
KVupp: Here we are handling a case where partial pieces were created with those associated with item ids. 
We should also handle the case where there were partial pieces created for pieces not associated with item ids.

javalkar-varun: From what I understand, in that case, pieces already exist so we don't need to create again.

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
javalkar-varun: Thanks for catching these @KVupp. I have updated code to handle them.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
javalkar-varun: Done

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
KVupp: That is correct.But, Consider the scenario: We are creating 3 pieces without item ids. After 1 piece is created the calls fails. When we retry we are supposed to create the remaining 2 pieces. But that logic is missing here. (The same logic here needs to be rewritten for pieces without itemids)

javalkar-varun: Got it. I have updated the logic. Code coverage went up a bit after this push even without adding a new test which proves we have quite a lot scenarios already covered through unit test :)

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -205,18 +207,25 @@ public boolean isUpdateSuccessful() {
...
+          List<String> existingItemIds = pieces.getPieces()
+																	             .stream()
+																	             .map(Piece::getItemId)
+																	             .filter(StringUtils::isNotEmpty)
+																	             .collect(Collectors.toList());
+
+          // Create piece records for item id's which do not have piece records yet     
+	        itemIds.stream()
+	               .filter(id -> !existingItemIds.contains(id))
+	               .forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));
```
piotr-kalashuk: With the latest updates one case is missed. 
Let's consider PO Line of "P/E Mix" format with 3 Physical resources and 2 Electronic resources with `create_inventory` set to `false`.
`itemIds` will not be `null` containing 3 elements. 2 piece records for Electronic resources will not be created.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -521,6 +523,37 @@ public void testPostOpenOrderInventoryUpdateOnlyForFirstPOL() throws Exception {
...
+
+    List<JsonObject> createdPieces = MockServer.serverRqRs.get(PIECES, HttpMethod.POST);
+    int piecesSize = createdPieces!=null ? createdPieces.size() : 0;
+    // Verify total number of pieces created should be equal to total quantity
+    assertEquals( piecesSize, calculateTotalQuantity(reqData));
+  }
+   
+  // Calculate total quantity of resources based of the cost
+  private int calculateTotalQuantity(CompositePurchaseOrder reqData) {
+		int eQuantity0 = reqData.getCompositePoLines().get(0).getCost().getQuantityElectronic() != null
```
piotr-kalashuk: Please use `HelperUtils.calculateTotalQuantity`

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -521,6 +523,37 @@ public void testPostOpenOrderInventoryUpdateOnlyForFirstPOL() throws Exception {
...
+  public void testPutOrdersByIdTotalPiecesEqualsTotalQuantityWhenCreateInventoryIsFalse() throws Exception {
+    logger.info("=== Test Put Order By Id create Pieces when Item record does not exist ===");
+
+    CompositePurchaseOrder reqData = getMockDraftOrder().mapTo(CompositePurchaseOrder.class);
+    reqData.setId(ID_FOR_PRINT_MONOGRAPH_ORDER);
+    // Make sure that mock PO has 2 po lines
+    assertEquals(2, reqData.getCompositePoLines().size());
+
+    reqData.setWorkflowStatus(CompositePurchaseOrder.WorkflowStatus.OPEN);
+    verifyPut(String.format(COMPOSITE_ORDERS_BY_ID_PATH, reqData.getId()), JsonObject.mapFrom(reqData).toString(), "", 204);
```
piotr-kalashuk: Please set `create_inventory` to false for one or both PO Lines so the test corresponds to its name

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
javalkar-varun: When `create_inventory` is set to `false`, `updateInventory` will not be called so `itemIds` will be `null`. Is that correct?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
javalkar-varun: Done

src/test/resources/po_listed_print_monograph_pe_mix.json
```diff
@@ -0,0 +1,148 @@
...
+          "quantity": 2,
+          "quantity_electronic": 2,
+          "quantity_physical": 0
+        },
+        {
+          "po_line_id": "c0d08448-347b-418a-8c2f-5fb50248d67e",
+          "location_id": "758258bc-ecc1-41b8-abca-f7b610822ffd",
+          "quantity": 0,
+          "quantity_electronic": 0,
+          "quantity_physical": 0
```
piotr-kalashuk: Is it intended to be 0? According to `cost` content, 3 physical and 2 electronic resources are expected. The first location covers electronic resources. I believe the second one is intended for physical but quantity is 0 which means create nothing in Inventory for this location.

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
piotr-kalashuk: Looks like `po_listed_print_monograph_pe_mix.json` created to cover this case but the physical quantity is zero in both locations. Once I modified the second location specifying `"quantity": 3`, the test has failed

image

src/test/resources/po_listed_print_monograph_pe_mix.json
javalkar-varun: Got it. I have corrected it and made corresponding changes to the implementation.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
javalkar-varun: Done

src/main/java/org/folio/rest/impl/PutOrderLineByIdHelper.java
```diff
@@ -193,30 +199,44 @@ public boolean isUpdateSuccessful() {
...
+          List<String> piecesForPhysicalResources = itemIds.stream()
+          .filter(id -> !existingItemIds.contains(id)).collect(Collectors.toList());
+
+          // 3. Create piece records     
+          piecesForPhysicalResources.stream().forEach(itemId -> futuresList.add(createPiece(poLineId, itemId)));  
+	        
+	        // 4. Calculate count for remaining pieces to create when Physical and Electronic resources exists
+	        // Scenario: PO Line of "P/E Mix" format with 3 Physical resources and 2 Electronic resources with `create_inventory` set to `false`.
+	        // Create 2 piece records for Electronic resources
+	        remainingPiecesToCreate = Math.abs(calculateTotalQuantity(compPOL.getCost()) - piecesForPhysicalResources.size());
```
KVupp: The scenario of failed electronic resources in case of P/E is missing here. Say there are 2 physical and 4 electronic to be created. In the first run 2 physical and 2 electronic are created and then call fails. In the next try, remainingPiecesToCreate would still be 4, whereas we should be creating pieces for just the remaining 2 electronic pieces.

javalkar-varun: I have updated code so I think this scenario will be handled as well. Thanks

PR #106, opened 2019-02-18
MODORDERS-159 PO Line: receipt status

grabsefx:
https://issues.folio.org/browse/MODORDERS-159
## Purpose
Handle described status transitions
## Approach
* Updating composite_po_line.json and po_line.json schemas with default receipt status values.
* Once the order transitions to "Open", changing receipt status transition to "Awaiting Receipt" if still set to "Pending"
* If receipt status is "Receipt not Required" no any record created in the Inventory nor any piece record in the Orders storage.

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -123,26 +130,40 @@ private boolean isPoNumberChanged(JsonObject poFromStorage, CompositePurchaseOrd
...
-      return completedFuture(null);
-    }
+    return getPoLines(poFromStorage.getString(ID), lang, httpClient, ctx, okapiHeaders, logger)
+      .thenCompose(polinesFromStorage -> {
+        JsonArray existedPoLinesArray = polinesFromStorage.getJsonArray(PO_LINES);
+        if (isNotEmpty(compPO.getCompositePoLines())) {
+          return handlePoLines(compPO, poFromStorage, existedPoLinesArray);
+        } else if (isPoNumberChanged(poFromStorage, compPO)) {
+          return updatePoLinesNumber(compPO, existedPoLinesArray);
+        } else if (isEmpty(compPO.getCompositePoLines()) && (existedPoLinesArray != null)) {
```
piotr-kalashuk: This check is not required because of internal agreement. Please refer to [order.raml](https://github.com/folio-org/mod-orders/blob/master/ramls/order.raml#L69):
>          Update a purchase order with given {id}
>          - if request does not include po_lines or includes "po_lines": [] or "po_lines": null - update just purchase order summary
>          - if request includes array of "po_lines" - update PO lines as per request

grabsefx: implementation fixed

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -62,13 +54,40 @@
...
+  }
+
   /**
    * Handles update of the order. First retrieve the PO from storage and depending on its content handle passed PO.
    */
   public void updateOrder(String orderId, CompositePurchaseOrder compPO) {
     compPO.setId(orderId);
     getPurchaseOrderById(orderId, lang, httpClient, ctx, okapiHeaders, logger)
       .thenCompose(poFromStorage -> validatePoNumber(compPO, poFromStorage))
+      .thenCompose(poFromStorage -> changePoLinesReceiptStatus(compPO, poFromStorage).thenApply(v -> poFromStorage))
```
pautau-s: If an error occurs when the order is transferred to the "Open" status, it may happen that the receipt status will be changed, but order status will remain"Pending". It may be better to change the receipt status in `PutOrderLineByIdHelper.updateInventory()` method? In this case, do not need additional call to get poLines by order id.

grabsefx: please, take a look at updated implementation

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -70,22 +84,20 @@ public void updateOrder(String orderId, CompositePurchaseOrder compPO) {
     getPurchaseOrderById(orderId, lang, httpClient, ctx, okapiHeaders, logger)
       .thenCompose(poFromStorage -> validatePoNumber(compPO, poFromStorage))
       .thenCompose(poFromStorage -> updatePoLines(poFromStorage, compPO).thenApply(v -> poFromStorage))
-      .thenAccept(poFromStorage -> {
-        CompletableFuture<Void> updatedPoFuture;
+      .thenCompose(poFromStorage -> {
         if (isTransitionToOpen(compPO, poFromStorage)) {
-          updatedPoFuture = openOrder(compPO);
+          return openOrder(compPO)
+            .thenCompose(v -> changePoLinesReceiptStatus(compPO))
```
pautau-s: Since the order can be created with the status "Open", it is necessary to provide for the possibility of updating the receipt status in this case. So I think it makes sense to call this in the "openOrder" method.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -99,10 +97,21 @@ public void updateOrder(String orderId, CompositePurchaseOrder compPO) {
   public CompletableFuture<Void> openOrder(CompositePurchaseOrder compPO) {
     compPO.setWorkflowStatus(OPEN);
     compPO.setDateOrdered(new Date());
-    return updateInventory(compPO)
-      .thenCompose(this::updateOrderSummary);
+    return fetchCompositePoLines(compPO)
+      .thenCompose(v -> changePoLineReceiptStatuses(compPO))
```
piotr-kalashuk: Instead of `thenCompose` you might use `thenAccept`. In this case `changePoLineReceiptStatuses` will not return `CompletionStage` but just `void`.

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
```diff
@@ -70,22 +72,18 @@ public void updateOrder(String orderId, CompositePurchaseOrder compPO) {
     getPurchaseOrderById(orderId, lang, httpClient, ctx, okapiHeaders, logger)
       .thenCompose(poFromStorage -> validatePoNumber(compPO, poFromStorage))
       .thenCompose(poFromStorage -> updatePoLines(poFromStorage, compPO).thenApply(v -> poFromStorage))
-      .thenAccept(poFromStorage -> {
-        CompletableFuture<Void> updatedPoFuture;
+      .thenCompose(poFromStorage -> {
         if (isTransitionToOpen(compPO, poFromStorage)) {
-          updatedPoFuture = openOrder(compPO);
+          return openOrder(compPO).thenCompose(v -> updateCompositePoLines(compPO));
```
piotr-kalashuk: You might move `updateCompositePoLines` call to `openOrder` method

grabsefx: done

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
grabsefx: fixed

src/main/java/org/folio/rest/impl/PutOrdersByIdHelper.java
grabsefx: done

PR #111, opened 2019-03-05
MODORDERS-170 - Prevent ordering from Inactive vendors/access provider

khandramai:
[MODORDERS-170](https://issues.folio.org/browse/MODORDERS-170) - Prevent ordering from Inactive vendors/access provider

## Purpose
To keep users from ordering from inappropriate Vendors/Access Providers

## Approach
In the case of an inactive or non-existent vendor/access provider during purchase order creation/updating client will receive a response containing the corresponding errors with vendor's id as an additional parameter.

## Pre-Merge Checklist:

1. **folio-api-tests** (as a result of vendor's id generation each time during sample data processing - vendors should have permanent id for passing validation + probably, inactive vendor can be added to sample data)

### DIFF
--- a/src/main/java/org/folio/rest/impl/OrdersImpl.java
+++ b/src/main/java/org/folio/rest/impl/OrdersImpl.java
@@ -44,14 +44,6 @@ public class OrdersImpl implements Orders {
   private static final String ORDERS_LOCATION_PREFIX = "/orders/composite-orders/%s";
   private static final String ORDER_LINE_LOCATION_PREFIX = "/orders/order-lines/%s";
 
-  private static int getPoLineLimit(JsonObject config) {
-    try {
-      return Integer.parseInt(config.getString(PO_LINES_LIMIT_PROPERTY, DEFAULT_POLINE_LIMIT));
-    } catch (NumberFormatException e) {
-      throw new NumberFormatException("Invalid limit value in configuration.");
-    }
-  }
-
   @Override
   @Validate
   public void deleteOrdersCompositeOrdersById(String id, String lang, Map<String, String> okapiHeaders,
@@ -337,6 +329,14 @@ public class OrdersImpl implements Orders {
                                 Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     // Return 501 (Not Implemented) for now
     asyncResultHandler.handle(succeededFuture(Response.status(501).build()));
+  }
+
+  private static int getPoLineLimit(JsonObject config) {
+    try {
+      return Integer.parseInt(config.getString(PO_LINES_LIMIT_PROPERTY, DEFAULT_POLINE_LIMIT));
+    } catch (NumberFormatException e) {
+      throw new NumberFormatException("Invalid limit value in configuration.");
+    }
   }
 
   private void validatePoLinesQuantity(CompositePurchaseOrder compPO, JsonObject config) {
--- a/src/main/java/org/folio/rest/impl/OrdersImpl.java
+++ b/src/main/java/org/folio/rest/impl/OrdersImpl.java
@@ -14,9 +14,11 @@ import static org.folio.orders.utils.ResourcePathResolver.PO_NUMBER;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import javax.ws.rs.core.Response;
 
+import me.escoffier.vertx.completablefuture.VertxCompletableFuture;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.orders.rest.exceptions.HttpException;
@@ -41,6 +43,14 @@ public class OrdersImpl implements Orders {
 
   private static final String ORDERS_LOCATION_PREFIX = "/orders/composite-orders/%s";
   private static final String ORDER_LINE_LOCATION_PREFIX = "/orders/order-lines/%s";
+
+  private static int getPoLineLimit(JsonObject config) {
+    try {
+      return Integer.parseInt(config.getString(PO_LINES_LIMIT_PROPERTY, DEFAULT_POLINE_LIMIT));
+    } catch (NumberFormatException e) {
+      throw new NumberFormatException("Invalid limit value in configuration.");
+    }
+  }
 
   @Override
   @Validate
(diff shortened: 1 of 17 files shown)


src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -367,5 +383,17 @@ public void getOrdersReceivingHistory(int offset, int limit, String query, Strin
 
   }
 
+    private CompletableFuture<Errors> validate(String lang, Map<String, String> okapiHeaders, Context vertxContext, HttpClientInterface httpClient, CompositePurchaseOrder compPO) {
+      CompletableFuture<Errors> future = new VertxCompletableFuture<>(vertxContext);
+      VendorHelper vendorHelper = new VendorHelper(lang, httpClient, okapiHeaders, vertxContext);
+      vendorHelper.validateVendor(compPO)
+        .thenAccept(errors -> vendorHelper.validateAccessProviders(compPO)
+        .thenApply(accessProvidersErrors -> {
```
craigmcnally: The indentation here is misleading

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -367,5 +383,17 @@ public void getOrdersReceivingHistory(int offset, int limit, String query, Strin
 
   }
 
+    private CompletableFuture<Errors> validate(String lang, Map<String, String> okapiHeaders, Context vertxContext, HttpClientInterface httpClient, CompositePurchaseOrder compPO) {
```
craigmcnally: This method can be simplified a little via the use of `thenCompose(...)`

src/test/resources/mockdata/vendors/active_inactive_access_providers.json
```diff
@@ -0,0 +1,347 @@
+{
```
craigmcnally: Note that there's another PR open for updating these data files to have camelCase fields... Please coordinate accordingly.  https://github.com/folio-org/mod-orders/pull/112

craigmcnally: whoops, these are vendor records.  disregard my earlier comment

src/main/java/org/folio/rest/impl/OrdersImpl.java
khandramai: Thanks, Craig! Good point. Since the validation of vendor and access providers was conceived by me as independent actions, I can suggest using thenCombine() method instead. This will allow us to perform validations in parallel mode, and I think this will have a positive impact on performance. And I removed misleading indentation.

craigmcnally: Yes, this is better... you can simplify even futher:

```
  private CompletableFuture<Errors> validate(String lang, Map<String, String> okapiHeaders, Context vertxContext,
      HttpClientInterface httpClient, CompositePurchaseOrder compPO) {
    VendorHelper vendorHelper = new VendorHelper(lang, httpClient, okapiHeaders, vertxContext);
    return vendorHelper.validateVendor(compPO)
      .thenCombine(vendorHelper.validateAccessProviders(compPO), (errors, accessProvidersErrors) -> {
        errors.getErrors().addAll(accessProvidersErrors.getErrors());
        errors.setTotalRecords(errors.getErrors().size());
        return errors;
      });
  }
```

khandramai: Done.

src/main/java/org/folio/rest/impl/OrdersImpl.java
khandramai: Done.

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -105,28 +106,34 @@ public void postOrdersCompositeOrders(String lang, CompositePurchaseOrder compPO
...
-            .thenAccept(withFunds -> {
-
-              logger.info("Successfully Placed Order: " + JsonObject.mapFrom(withFunds).encodePrettily());
-              httpClient.closeClient();
-              javax.ws.rs.core.Response response = PostOrdersCompositeOrdersResponse.respond201WithApplicationJson(withFunds,
-                PostOrdersCompositeOrdersResponse.headersFor201().withLocation(String.format(ORDERS_LOCATION_PREFIX, withFunds.getId())));
-              AsyncResult<javax.ws.rs.core.Response> result = Future.succeededFuture(response);
-              asyncResultHandler.handle(result);
+
+    validate(lang, okapiHeaders, vertxContext, httpClient, compPO).thenAccept(errs -> {
```
piotr-kalashuk: This check was added basically for cases when an order is being created by external modules (like GOBI) because UI will work only with active vendors.
@craigmcnally @KVupp - is it okay to reject order creation in this case or you think order still can be created but in `Pending` state? At later stage we might add alert for such order specifying the reason...

craigmcnally: Checking that the vendor actually exists is probably not needed here... mod-gobi performs a lookup of the vendor UUID given a vendor code, e.g. GOBI

piotr-kalashuk: Okay, this extra validation then can be removed if modules like `mod-gobi` are going to validate the vendor as well (i.e. vendor exists and active).

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -0,0 +1,194 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(VendorHelper.class);
+
+  private final HttpClientInterface httpClient;
+  private final Map<String, String> okapiHeaders;
+  private final Context ctx;
+  private final String lang;
+
+  private static final String ID = "id";
+  private static final String VENDORS = "vendors";
+  private static final String VENDOR_STATUS = "vendor_status";
```
piotr-kalashuk: My personal preference is to use [mod-vendors schemas](https://github.com/folio-org/acq-models/tree/master/mod-vendors/schemas) instead of `JsonObject`. This requires extra conversion but it allows to catch potential issues at compile time rather than run-time.

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Returns list of access providers ids
+   *
+   * @param compPO composite purchase order
+   * @return list of access providers id
+   */
+  private List<String> verifyAndGetAccessProvidersIdsList(CompositePurchaseOrder compPO) {
+    return compPO.getCompositePoLines().stream()
+      .filter(p -> (p.getEresource() != null && p.getEresource().getAccessProvider() != null))
+      .map(p -> p.getEresource().getAccessProvider()).collect(Collectors.toList());
```
piotr-kalashuk: Please consider either adding `distinct()` or using `toSet()` collector just to not include duplicates to query

src/main/java/org/folio/rest/impl/OrdersImpl.java
craigmcnally: validating that all UUID's provided are valid is probably something we can hold off on for now

PR #120, opened 2019-03-16
MODORDERS-170 - Prevent ordering from Inactive vendors/access provider (Improvements)

khandramai:
Refactoring of an existing VendorHelper from PR [#111](https://github.com/folio-org/mod-orders/pull/111) in accordance with the general helper structure implemented in scope of [MODORDERS-141](https://issues.folio.org/browse/MODORDERS-141).

### DIFF
--- a/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
@@ -238,7 +238,7 @@ public class PurchaseOrderHelper extends AbstractHelper {
     }
 
     return validatePoLineLimit(compPO)
-      .thenCompose(isStillValid -> isStillValid ? validateVendor(compPO) : completedFuture(false));
+      .thenCompose(isStillValid -> isStillValid ? validateVendorsAndAccessProviders(compPO) : completedFuture(false));
   }
 
   /**
@@ -290,14 +290,13 @@ public class PurchaseOrderHelper extends AbstractHelper {
     return completedFuture(null);
   }
 
-  private CompletableFuture<Boolean> validateVendor(CompositePurchaseOrder compPO) {
+  private CompletableFuture<Boolean> validateVendorsAndAccessProviders(CompositePurchaseOrder compPO) {
     if (compPO.getWorkflowStatus() == WorkflowStatus.OPEN) {
-      VendorHelper vendorHelper = new VendorHelper(lang, httpClient, okapiHeaders, ctx);
+      VendorHelper vendorHelper = new VendorHelper(httpClient, okapiHeaders, ctx, lang);
       return vendorHelper
         .validateVendor(compPO)
-        .thenCombine(vendorHelper.validateAccessProviders(compPO), (vendorErrors, accessProvidersErrors) -> {
-          addProcessingErrors(vendorErrors.getErrors());
-          addProcessingErrors(accessProvidersErrors.getErrors());
+        .thenCombine(vendorHelper.validateAccessProviders(compPO), (rVoid, errors) -> {
+          addProcessingErrors(errors.getErrors());
           return getErrors().isEmpty();
         });
     }
--- a/src/main/java/org/folio/rest/impl/VendorHelper.java
+++ b/src/main/java/org/folio/rest/impl/VendorHelper.java
@@ -2,8 +2,6 @@ package org.folio.rest.impl;
 
 import io.vertx.core.Context;
 import io.vertx.core.json.JsonObject;
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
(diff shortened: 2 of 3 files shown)


khandramai:
Hi @craigmcnally / @piotr-kalashuk,
to match the refactoring performed in [MODORDERS-141](https://issues.folio.org/browse/MODORDERS-141) I would like to suggest the following changes for VendorHelper. I believe this will be useful given the unification of the structure of the helpers.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -290,14 +290,13 @@ protected Errors getProcessingErrors() {
...
+  private CompletableFuture<Boolean> validateVendorsAndAccessProviders(CompositePurchaseOrder compPO) {
     if (compPO.getWorkflowStatus() == WorkflowStatus.OPEN) {
-      VendorHelper vendorHelper = new VendorHelper(lang, httpClient, okapiHeaders, ctx);
+      VendorHelper vendorHelper = new VendorHelper(httpClient, okapiHeaders, ctx, lang);
       return vendorHelper
         .validateVendor(compPO)
-        .thenCombine(vendorHelper.validateAccessProviders(compPO), (vendorErrors, accessProvidersErrors) -> {
-          addProcessingErrors(vendorErrors.getErrors());
-          addProcessingErrors(accessProvidersErrors.getErrors());
+        .thenCombine(vendorHelper.validateAccessProviders(compPO), (rVoid, errors) -> {
```
piotr-kalashuk: Taking into account that results of the `validateVendor` method are not used here explicitly, no sense to use `thenCombine`...

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -63,29 +54,28 @@ public VendorHelper(String lang, HttpClientInterface httpClient, Map<String, Str
...
     if(id != null) {
       getVendorById(id)
         .thenApply(vendor -> {
           VendorStatus status = VendorStatus.valueOf(vendor.getVendorStatus().toUpperCase());
           if(status != VendorStatus.ACTIVE) {
-            errors.add(createErrorWithId(ORDER_VENDOR_IS_INACTIVE, id));
+            addProcessingError(createErrorWithId(ORDER_VENDOR_IS_INACTIVE, id));
           }
-          return buildErrorsObject(errors);
+          return getProcessingErrors();
```
piotr-kalashuk: The helper is stateful and shares all the errors already registered prior to this method call. In case the validation of the PO vendor results to success (i.e. no errors found) but the helper already registered errors, for example, for `validateAccessProviders`, the `CompletableFuture` will have `Errors` object with non-empty `Error` list. At least forr me this is confusing...

craigmcnally: I agree

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -2764,8 +2764,10 @@ public void testCreatePoWithDifferentVendorStatus() throws Exception {
     assertThat(activeVendorActiveAccessProviderErrors.getErrors(), empty());
 
     // Internal mod-vendor error
-    Errors internalServerError = verifyPostResponseErrors(1, MOD_VENDOR_INTERNAL_ERROR_ID, ACTIVE_ACCESS_PROVIDER_A, ACTIVE_ACCESS_PROVIDER_B);
-    assertEquals(GENERIC_ERROR_CODE.getCode(), internalServerError.getErrors().get(0).getCode());
+    Errors errors = verifyPostResponse(COMPOSITE_ORDERS_PATH, getPoWithVendorId(MOD_VENDOR_INTERNAL_ERROR_ID, ACTIVE_ACCESS_PROVIDER_A, ACTIVE_ACCESS_PROVIDER_B),
+      EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10, APPLICATION_JSON, 500).as(Errors.class);
+    assertEquals(errors.getTotalRecords(), new Integer(1));
```
craigmcnally: I think the `assertEquals(...)` args are in order (expected, actual), which is annoyingly the opposite of `assertThat(...)`...

Given this, and that `new Integer(1)` seems a little clumsy to me, I would suggest using the following instead:

`assertThat(errors.getTotalRecords(), equalTo(1));`

craigmcnally:
So I've reviewed and closed MODORDERS-170, though maybe I shouldn't have.  It's also listed as done in the mod-orders-4.0.0 release.  However, this has not been approved/merged, nor have the related API test changes in MODORDERS-192.  

So what's the plan for these PRs then?  Do we need to create a follow-on story/bug for these improvements?  Is the plan to get this work into this sprint?  Are we planning on create a patch release of mod-orders?

PR #123, opened 2019-03-19
Updates to PO line's cost and removing adjustments

piotr-kalashuk:
## Purpose
[MODORDERS-180](https://issues.folio.org/browse/MODORDERS-180) Updates to poLine Cost
[MODORDERS-181](https://issues.folio.org/browse/MODORDERS-181) Updates to PO total estimated cost

## Approach
- MODORDERS-180
  - PO Line's estimated price is calculated based on cost details before sending request to storage (the [HALF_UP](https://docs.oracle.com/javase/8/docs/api/java/math/RoundingMode.html#HALF_UP) rounding mode is used and scale depends on [Currency](https://docs.oracle.com/javase/8/docs/api/java/util/Currency.html) code)
  - Validation of the new Cost's properties is based on rules from [Acquisitions Interface Fields](https://docs.google.com/spreadsheets/d/1F88PoSnEKD7oFnza1ZuOgzAgKnK95XOyTQzxyww3tZA/edit#gid=1408113181&range=A37). The following new 4 error codes might be returned:
    - `costUnitPriceInvalid`
    - `costUnitPriceElectronicInvalid`
    - `costAdditionalCostInvalid`
    - `costDiscountInvalid`
- MODORDERS-181
  - Removed adjustment related logic
  - Total Estimated Price calculated as sum of PO Line's total prices

#### TODOS and Open Questions
- [x] Point to merge commit of the acq-models sub-module  once [acq-models PR#103](https://github.com/folio-org/acq-models/pull/103) is merged

## Pre-Merge Checklist:

  - [x] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [x] Do they have all they appropriate links to blocked/related issues?

### DIFF
--- a/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
@@ -15,8 +15,6 @@ import static org.folio.orders.utils.HelperUtils.getPurchaseOrderById;
 import static org.folio.orders.utils.HelperUtils.handleGetRequest;
 import static org.folio.orders.utils.HelperUtils.loadConfiguration;
 import static org.folio.orders.utils.HelperUtils.operateOnObject;
-import static org.folio.orders.utils.ResourcePathResolver.ADJUSTMENT;
-import static org.folio.orders.utils.ResourcePathResolver.PO_LINES;
 import static org.folio.orders.utils.ResourcePathResolver.PO_LINE_NUMBER;
 import static org.folio.orders.utils.ResourcePathResolver.PURCHASE_ORDER;
 import static org.folio.orders.utils.ResourcePathResolver.resourceByIdPath;
@@ -29,6 +27,8 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.stream.Collectors;
@@ -106,7 +106,8 @@ public class PurchaseOrderHelper extends AbstractHelper {
   public CompletableFuture<CompositePurchaseOrder> createPurchaseOrder(CompositePurchaseOrder compPO) {
     return setPoNumberIfMissing(compPO)
       .thenCompose(v -> poNumberHelper.checkPONumberUnique(compPO.getPoNumber()))
-      .thenCompose(v -> createPOandPOLines(compPO));
+      .thenCompose(v -> createPOandPOLines(compPO))
+      .thenApply(this::populateOrderSummary);
   }
 
   /**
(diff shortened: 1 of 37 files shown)


piotr-kalashuk:
The first commit contains business logic changes (+ unit tests). The second one mock data file changes. This should simplify code review a bit.

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
```diff
@@ -333,6 +341,49 @@ private String buildPoLineNumber(String poNumber, String sequence) {
...
+      BigDecimal ePrice = BigDecimal.valueOf(cost.getListUnitPriceElectronic())
+                                    .multiply(BigDecimal.valueOf(cost.getQuantityElectronic()));
+      total = total.add(ePrice);
+    }
+    if (cost.getDiscount() != null) {
+      BigDecimal discount;
+      if (Cost.DiscountType.AMOUNT == cost.getDiscountType()) {
+        discount = BigDecimal.valueOf(cost.getDiscount());
+      } else {
+        discount = total.multiply(BigDecimal.valueOf(cost.getDiscount()/100d));
```
KVupp: The calculation in the story for discount amount is given as 
>discountAmount = discountType == PERCENTAGE ? baseListPrice * discount : discount. 
but here we are calculating per 100. Both are not wrong, but it comes down if cost.discount = 10 vs cost.discount= 0.1
Either way, this must be corrected or the story description(because we are referencing the story# in the method description we must make sure story descriptions are always correct)

craigmcnally: I had assumed discount would be a floating point value < 1 if discountType == PERCENT

craigmcnally: I think we might want to actually put some validation in place for this... something along the lines of if it's a percent and > 1 throw an error.

craigmcnally: Honestly thought I didn't catch this when reviewing the acq-models PR which IIRC has examples with discount=10, discountType=Percent, so good catch @KVupp.

craigmcnally: I think keeping this as-is is probably fine, but we need to make sure it's documented in the schema

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -211,10 +210,34 @@
...
+   * See MODORDERS-181 for more details.
+   * @param poLines list of composite PO Lines
+   * @return estimated purchase order's total price
+   */
+  private Double calculateTotalEstimatedPrice(List<CompositePoLine> poLines) {
+    return poLines.stream()
+      .map(CompositePoLine::getCost)
+      .filter(Objects::nonNull)
+      .mapToDouble(cost -> Optional
+        .ofNullable(cost.getPoLineEstimatedPrice())
```
KVupp: when do we anticipate the totalEstimatedPrice to be null? At this point in code shouldn't it either be already calculated ?

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
```diff
@@ -333,6 +341,49 @@ private String buildPoLineNumber(String poNumber, String sequence) {
...
+    }
+  }
+
+  /**
+   * Calculates total estimated price. See MODORDERS-180 for more details.
+   * @param cost PO Line's cost
+   */
+  double calculateEstimatedPrice(Cost cost) {
+    BigDecimal total = BigDecimal.ZERO;
+    if (cost.getListUnitPrice() != null) {
```
KVupp: I think it would be nice to add these validations in the PurchaseOrderHelper.validateOrder method too, so that we do not end put calculating electronic costs for physical orders

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
piotr-kalashuk: This is just to cover the case when some third-party added the data directly to storage instead of working via `mod-orders` but probably you are right and this is not required

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
piotr-kalashuk: The description of the discount property in the schema file has been updated:
`Percentage (0 to 100) or amount (positive number) that is subtracted from the list price time quantities calculation before additional cost`

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
piotr-kalashuk: Validation has been updated

PR #127, opened 2019-03-27
MODORDERS-201 correct PO total estimated cost calculation

piotr-kalashuk:
## Purpose
[MODORDERS-201](https://issues.folio.org/browse/MODORDERS-201) Incorrect PO total estimated cost in some cases

## Approach
Replacing sum of regular `double` values by operation with `BigDecimal`

## Learning
[Why You Should Never Use Float and Double for Monetary Calculations](https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio)

## Pre-Merge Checklist:

### DIFF
--- a/src/test/java/org/folio/rest/impl/OrdersImplTest.java
+++ b/src/test/java/org/folio/rest/impl/OrdersImplTest.java
@@ -33,6 +33,7 @@ import static org.folio.orders.utils.ErrorCodes.ZERO_COST_QTY;
 import static org.folio.orders.utils.HelperUtils.COMPOSITE_PO_LINES;
 import static org.folio.orders.utils.HelperUtils.DEFAULT_POLINE_LIMIT;
 import static org.folio.orders.utils.HelperUtils.calculateEstimatedPrice;
+import static org.folio.orders.utils.HelperUtils.calculateTotalEstimatedPrice;
 import static org.folio.orders.utils.HelperUtils.calculateTotalQuantity;
 import static org.folio.orders.utils.HelperUtils.calculateInventoryItemsQuantity;
 import static org.folio.orders.utils.HelperUtils.convertIdsToCqlQuery;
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -924,8 +930,10 @@ private void verifyCalculatedData(CompositePurchaseOrder resp) {
       .getCompositePoLines()
       .stream()
       .map(CompositePoLine::getCost)
-      .mapToDouble(Cost::getPoLineEstimatedPrice)
-      .sum();
+      .map(Cost::getPoLineEstimatedPrice)
```
KVupp: Just curious, any reason behind using the same code in unit tests too?
if we are using the exact same calculation that is used in the code for verification too,  will it be fair testing? the value will always end up the same , and isn't it partly the reason why we didn't know about the loss of precision?

craigmcnally: I agree that having a static expected value to compare against is probably better here.

craigmcnally:
I know we ran into this same issue when first developing the mapper functionality, and I think I found the very same article.  I wonder if there's a better (move visible) place to document these sorts of pitfalls so that others on the team don't make the same mistake... especially since we're going to be doing a lot more monetary calculations once we start processing invoices and get into the finance side of things (payments/credits/transfers/etc.)  Any ideas?

src/test/java/org/folio/rest/impl/OrdersImplTest.java
piotr-kalashuk: Moved to utility helper.

KVupp: The idea here is to use a assign a value vs the value being calculated even from a utility helper
for ex: double expectedPrice = 49.32 etc..

PR #129, opened 2019-03-28
MODORDERS-202 Dereference contacts from vendor model

piotr-kalashuk:
## Purpose
[MODORDERS-202](https://issues.folio.org/browse/MODORDERS-202) Dereference contacts from vendor model

In scope of [MODORDERS-170](https://issues.folio.org/browse/MODORDERS-170) mod-orders started validating vendor on order transition to Open. The logic uses Vendor model generated from [vendor.json](https://github.com/folio-org/acq-models/blob/master/mod-vendors/schemas/vendor.json) which means once incompatible changes done there and schema reference is not updated, the logic is not able to work with Vendor and fails.

## Approach
- Updating to latest ACQ models
- Updating VendorHelper's logic to return vendor generic error code if some unknown issue happened (like content of the vendor cannot be recognized)
- The unit test added to verify that VendorHelper returns `vendorIssue` error is vendor content cannot be handled
- Existing vendor mock json files are updated leaving only minimal info just to simplify future updates

#### TODOS and Open Questions
- [ ] Merge this PR together with [mod-vendors PR#54](https://github.com/folio-org/mod-vendors/pull/54)

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -62,11 +62,12 @@ public VendorHelper(Map<String, String> okapiHeaders, Context ctx, String lang)
         })
         .thenAccept(future::complete)
         .exceptionally(t -> {
-          if(HttpStatus.HTTP_NOT_FOUND.toInt() == (((HttpException) t.getCause()).getCode())) {
+          Throwable cause = t.getCause();
+          if(cause instanceof HttpException && HttpStatus.HTTP_NOT_FOUND.toInt() == (((HttpException) cause).getCode())) {
```
craigmcnally: whitespace

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -100,12 +101,14 @@ public VendorHelper(Map<String, String> okapiHeaders, Context ctx, String lang)
...
-          .forEach(id -> errors.add(POL_ACCESS_PROVIDER_NOT_FOUND.toError().withAdditionalProperty(ID, id)));
+          .forEach(id -> errors.add(createErrorWithId(POL_ACCESS_PROVIDER_NOT_FOUND, id)));
         return handleAndReturnErrors(errors);
       }).thenAccept(future::complete)
         .exceptionally(t -> {
-          logger.error("Failed to validate access provider's status", t);
-          future.completeExceptionally(t);
+          Throwable cause = t.getCause();
+          logger.error("Failed to validate access provider's status", cause);
+          errors.add(VENDOR_ISSUE.toError().withMessage(cause.getMessage()));
```
craigmcnally: I wonder if it's worth being more specific about vendor issues... e.g. having separate codes for the various issues... Or, at least passing the same message you log above instead of the one from the cause?  

I say this not knowing what the cause message looks like in this case... So feel free to push back

piotr-kalashuk: There are really 2 ways to handle such unexpected errors:
1. Hide all the details and always use static/generic message.
In this case the only source to understand the cause of the problem is back-end application logs.
2. Use static/generic error code but provide details of the issue as error message. 
In this case any client of the API can contact back-end developers (of course by jira ticket) with useful details describing the problem

The UI application is not going to use error message but just display to end user localized message which is based on error code. So most of the consumers of the UI application will never know about real cause...
My personal preference as of back-end developer to be able to see details of the issue just using Browser's console and then use logs to have more details if required.

For example, the error for the case this PR is created for looks like following:
```json
{
  "message" : "Cannot deserialize instance of `java.lang.String` out of START_OBJECT token\n at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.folio.rest.acq.model.Vendor[\"contacts\"]->java.util.ArrayList[0])",
  "code" : "vendorIssue",
  "parameters" : [ ]
}
```

craigmcnally: I like the idea of having the additional "cause" field.  I didn't realize that was an option.

> The UI application is not going to use error message but just display to end user localized message which is based on error code. So most of the consumers of the UI application will never know about real cause...

This is exactly why I made the suggestion of using discrete codes instead of a general vendorIssue code.  It might be helpful if the UI could provide the user with a specific localized message which indicates what the problem was instead of just that there was a problem.

piotr-kalashuk: To be honest I do not know what error code we might return because this is exceptionally section which means something unexpected happened...
By the way I changed the handling a bit and here is the output of the sample response (from the unit test) for the unexpected error verifying order vendor and PO Line's acceess provider:
```json
{
  "errors": [
    {
      "message": "Order cannot be updated due to issues related to associated vendor/access provider",
      "code": "vendorIssue",
      "parameters": [
        {
          "key": "id",
          "value": "5a34ae0e-5a11-4337-be95-1a20cfdc3161"
        }
      ],
      "cause": "Unrecognized field \"contactus\" (class org.folio.rest.acq.model.Vendor), not marked as ignorable (36 known properties: \"vendor_types\", \"vendor_currencies\", \"licensor\", \"vendor_status\", \"name\", \"contacts\", \"emails\", \"language\", \"interfaces\", \"id\", \"governmental\", \"changelogs\", \"addresses\", \"urls\", \"accounts\", \"subscription_interval\", \"agreements\", \"edi\", \"access_provider\", \"code\", \"tax_percentage\", \"tax_id\", \"phone_numbers\", \"payment_method\", \"description\", \"aliases\", \"discount_percent\", \"erp_code\", \"liable_for_vat\", \"expected_receipt_interval\", \"material_supplier\", \"san_code\", \"claiming_interval\", \"expected_activation_interval\", \"expected_invoice_interval\", \"renewal_activation_interval\"])\n at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.folio.rest.acq.model.Vendor[\"contactus\"])"
    },
    {
      "message": "Order cannot be updated due to issues related to associated vendor/access provider",
      "code": "vendorIssue",
      "parameters": [],
      "cause": "Unrecognized field \"contactus\" (class org.folio.rest.acq.model.Vendor), not marked as ignorable (36 known properties: \"vendor_types\", \"vendor_currencies\", \"licensor\", \"vendor_status\", \"name\", \"contacts\", \"emails\", \"language\", \"interfaces\", \"id\", \"governmental\", \"changelogs\", \"addresses\", \"urls\", \"accounts\", \"subscription_interval\", \"agreements\", \"edi\", \"access_provider\", \"code\", \"tax_percentage\", \"tax_id\", \"phone_numbers\", \"payment_method\", \"description\", \"aliases\", \"discount_percent\", \"erp_code\", \"liable_for_vat\", \"expected_receipt_interval\", \"material_supplier\", \"san_code\", \"claiming_interval\", \"expected_activation_interval\", \"expected_invoice_interval\", \"renewal_activation_interval\"])\n at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.folio.rest.acq.model.Vendor[\"contactus\"])"
    }
  ],
  "total_records": 2
}
```

src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -32,6 +32,7 @@
   ORDER_VENDOR_IS_INACTIVE("vendorIsInactive", "Order cannot be open as the associated vendor is inactive"),
   POL_ACCESS_PROVIDER_IS_INACTIVE("accessProviderIsInactive", "Order cannot be open as the associated access provider is inactive"),
   ORDER_VENDOR_NOT_FOUND("vendorNotFound", "Order cannot be open as the associated vendor not found"),
+  VENDOR_ISSUE("vendorIssue", "Order cannot be updated due to issues related to associated vendor/access provider"),
```
KVupp: All we are using this code is when the we are unable to verify status, I think we can be more specific with the message.

craigmcnally: see above comments.  If needed we can address in a follow-up story

PR #110, opened 2019-03-05
MODORDERS-175 Add ability to move a received piece back to "Expected"

piotr-kalashuk:
## Purpose
[MODORDERS-175](https://issues.folio.org/browse/MODORDERS-175) Add ability to move a received piece back to "Expected"
## Approach
The logic extends functionality implemented in scope of PR #107:
* client sends request like following:
  ```json
  {
    "toBeReceived": [
      {
        "poLineId": "f217a5c2-2c56-4d05-9412-a96cfc8e52de",
        "received": 2,
        "receivedItems": [
          {
            "itemStatus": "On order",
            "pieceId": "56fbfde4-6335-4dd7-9a03-d100821f1d18"
          },
          {
            "itemStatus": "On order",
            "pieceId": "a6ff14ad-8a65-57f9-bb89-f4654a3923fb"
          }
        ]
      }
    ],
    "totalRecords": 2
  }
  ```
* retrieve piece records
* for those piece records which have an item id reference, retrieve items from inventory storage
  * ~~remove item's barcode~~ the item's barcode remains as is
  * update item's status to `On order`
  * update item record in inventory
* update each piece record removing received date and setting receiving status to `Expected`
* retrieve PO Line and calculate receipt status based on quantity of received and expected piece records. If the receipt status was `Fully Received` but the calculated one is either `Awaiting Receipt` or `Partially Received`, the receipt date is removed

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards - [sonarcloud PR#110](https://sonarcloud.io/project/issues?id=org.folio%3Amod-orders&pullRequest=110&resolved=false)?
  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [x] No

PR #100, opened 2019-02-06
MODORDERS-154 PO Line: make location repeatable

piotr-kalashuk:
## Purpose
[MODORDERS-154](https://issues.folio.org/browse/MODORDERS-154) PO Line: make location details repeatable

## Approach
- Switching to schemas where location is repeatable
- Update logic to work with list of locations
- Inventory interaction
  - The holding is created for each unique location of the particular PO Line
  - The items are created taking into account corresponding holdings
- Update all mock files
- Validation
  - physical/electronic quantity should depend on PO Line's `order_format`
  - sum of physical quantity of all locations must not exceed `cost.quantity_physical`
  - sum of electronic quantity of all locations must not exceed `cost.quantity_electronic`

### TODO
- [x] Validation of the PO Line
- [ ] Validation of the PO Line's locations presence cannot be done until UIOR-121 implemented

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards ([sonar mod-orders PR-100](https://sonarcloud.io/project/issues?id=org.folio%3Amod-orders&pullRequest=100&resolved=false))?
  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [x] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [ ] There are no breaking changes in this PR.
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

PR #130, opened 2019-03-29
MODORDERS-195- Specify material Type while creating item

KVupp:
https://issues.folio.org/browse/MODORDERS-195

## Purpose
https://issues.folio.org/browse/MODORDERS-195 - specify the material type while creating an item

## Approach

- calculate required item quantities by Material type
- added logic to verify material type to be provided if createInventory required "Item" to be created.

ADDITIONAL CHANGES
- Refactored logic to check and assign default create inventory first and then validate the order
- Fixed a missing logic where the default createInventory was not being assigned when a physical or electronic block is not given at all in a request

## TODO
- coordinate with [mod-orders-#194](https://github.com/folio-org/mod-orders/pull/131) before merging
- Merge it along with [mod-order-storage-#100](https://github.com/folio-org/mod-orders-storage/pull/100) and [acqmodels-#110](https://github.com/folio-org/acq-models/pull/110)

## Pre-Merge Checklist:


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -342,6 +343,16 @@ public static String getEndpointWithQuery(String query, Logger logger) {
       errors.add(ErrorCodes.ELECTRONIC_LOC_QTY_EXCEEDS_COST);
     }
 
+    if (compPOL.getPhysical().getCreateInventory() == Physical.CreateInventory.INSTANCE_HOLDING_ITEM
```
craigmcnally: Instead of repeating these blocks of code, can we pull this into a shared/common `validateMaterialTypes(...)` method, a la `validateCostPrices(...)`?

KVupp:
Changes required from (mod-orders-194)[https://issues.folio.org/browse/MODORDERS-194] are now merged and also the raml-cop issues are fixed. This PR is ready for review

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -244,12 +237,19 @@ private int calculateTotalItemsQuantity(List<CompositePoLine> poLines) {
...
 
-    // If static validation has failed, no need to call other services
-    if (!getErrors().isEmpty()) {
-      return completedFuture(false);
-    }
+    compPO.getCompositePoLines()
+      .forEach(poLine ->
+        orderLineHelper.setTenantDefaultCreateInventoryValues(poLine)
+          .thenApply(v -> {
+          addProcessingErrors(HelperUtils.validatePoLine(poLine));
```
piotr-kalashuk: This `thenApply` block might be executed after `validatePoLineLimit` completes and in some case errors might be ignored...

A bit of details. 
1. The `orderLineHelper.setTenantDefaultCreateInventoryValues` will send request to `mod-configuration` and `thenApply` will be executed only after response is received.
2. `validatePoLineLimit` method below will send independent request to `mod-configuration`

In case for some reason the response for the first call will come later than for the second one, any error found in the step 1 might be missed because both completable futures are fully independent.

I simulated this by following test (which fails `Expected status code <422> but was <201>`) modifying behavior of `MockServer#handleConfigurationModuleResponse` method:

```java
...
  @Test
  public void testPostOrdersWithEmptyCreateInventory2() throws Exception {
    JsonObject order = new JsonObject(getMockData(MONOGRAPH_FOR_CREATE_INVENTORY_TEST));
    // Get Open Order
    CompositePurchaseOrder reqData = order.mapTo(CompositePurchaseOrder.class);

    // Clear CreateInventory for setting default values
    reqData.getCompositePoLines().get(0).getPhysical().setCreateInventory(null);
    reqData.getCompositePoLines().get(0).getCost().setQuantityPhysical(-1);

    verifyPostResponse(COMPOSITE_ORDERS_PATH, JsonObject.mapFrom(reqData).toString(),
      prepareHeaders(EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10), APPLICATION_JSON, 422);
  }
...
  public static class MockServer {
...
    static int delay = 1001;
    private void handleConfigurationModuleResponse(RoutingContext ctx) {
      delay -= 500;
      vertx.setTimer(delay, id -> {
        try {
          String tenant = ctx.request().getHeader(OKAPI_HEADER_TENANT) ;
          if (PO_NUMBER_ERROR_X_OKAPI_TENANT.getValue().equals(tenant)) {
            tenant = EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10.getValue();
          }
          serverResponse(ctx, 200, APPLICATION_JSON, getMockData(String.format(CONFIG_MOCK_PATH, tenant)));
        } catch (IOException e) {
          serverResponse(ctx, 500, APPLICATION_JSON, INTERNAL_SERVER_ERROR);
        }
      });
    }
```

The proposal is to move `setTenantDefaultCreateInventoryValues` block to separate method like `CompletableFuture<CompositePurchaseOrder> setTenantDefaultValues(CompositePurchaseOrder compPO)` and start any validation only on completion

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
```diff
@@ -168,30 +166,84 @@ private CompositePurchaseOrder validateOrderState(CompositePurchaseOrder po) {
...
+        || compPOL.getOrderFormat().equals(OrderFormat.P_E_MIX)) {
       String tenantDefault = jsonConfig.getString(ERESOURCE);
-      Eresource.CreateInventory eresourceDefaultValue = StringUtils.isEmpty(tenantDefault) ?
-        Eresource.CreateInventory.INSTANCE_HOLDING : Eresource.CreateInventory.fromValue(tenantDefault);
-      compPOL.getEresource().setCreateInventory(eresourceDefaultValue);
+      Eresource.CreateInventory eresourceDefaultValue = getEresourceInventoryDefault(tenantDefault);
+      if (compPOL.getEresource() == null) {
+        compPOL.setEresource(new Eresource());
+      }
+      if(isEresourceInventoryNotPresent(compPOL)) {
```
piotr-kalashuk: Optional. This condition might be part of OrderFormat check above as it may turn out that no need to set value as it is already set...

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -142,7 +147,7 @@
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
-
+import javax.swing.plaf.synth.SynthSeparatorUI;
```
piotr-kalashuk: Unused import...

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -1797,15 +1809,42 @@ private void verifyHoldingsCreated(List<JsonObject> holdings, CompositePoLine po
   }
 
   private void verifyItemsCreated(List<JsonObject> inventoryItems, CompositePoLine pol, int expectedQuantity) {
-    int actualQuantity = 0;
+    Map<Piece.Format, Integer> expectedItemsPerResourceType = HelperUtils.calculatePiecesQuantity(pol,
+        pol.getLocations(), true);
+
+    Map<String, List<JsonObject>> itemsByMaterial = inventoryItems.stream()
+      .filter(item -> pol.getId().equals(item.getString(ITEM_PURCHASE_ORDER_LINE_IDENTIFIER)))
+      .collect(toMap(item -> item.getString(ITEM_MATERIAL_TYPE_ID), item -> {
```
piotr-kalashuk: Might be replaced by
```java
.collect(Collectors.groupingBy(item -> item.getString(ITEM_MATERIAL_TYPE_ID)));
```

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -3774,20 +3814,10 @@ private void handleGetHoldingRecord(RoutingContext ctx) {
     private void handleGetItemRecordsFromStorage(RoutingContext ctx) {
       logger.info("handleGetItemRecordsFromStorage got: " + ctx.request().path());
 
-      try {
-        JsonObject items;
-        if (ctx.request().getParam("query").contains(PO_LINE_ID_FOR_SUCCESS_CASE)) {
```
piotr-kalashuk: I think this is used to emulate case when items are partially created in inventory so the search of items by PO Line id returns 1 item all the time...

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -865,4 +877,15 @@ public static boolean isHoldingsUpdateRequired(CompositePoLine compPOL) {
   public static boolean isItemsUpdateRequired(CompositePoLine compPOL) {
     return isItemsUpdateRequiredForPhysical(compPOL) || isItemsUpdateRequiredForEresource(compPOL);
   }
+
+  private static boolean isMaterialMissing(CompositePoLine compPOL) {
+    if (compPOL.getOrderFormat().equals(OrderFormat.ELECTRONIC_RESOURCE)
+        || compPOL.getOrderFormat().equals(OrderFormat.P_E_MIX))
+      return compPOL.getEresource().getCreateInventory() == (Eresource.CreateInventory.INSTANCE_HOLDING_ITEM)
+          && isEmpty(compPOL.getEresource().getMaterialType());
```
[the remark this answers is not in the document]
KVupp: The below is not an else block it is another if block, the Physical case will be covered there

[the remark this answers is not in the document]
piotr-kalashuk: Looking at implementation I see that below if statement will not be invoked because `return` is used at 884 line

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
KVupp: This is to cover the case of a P/E mix order Format, and an eresource block is given with a createInventory block and a physical is blank, we are supposed to set the default only for the Physical block
for ex:
{
"orderFormat":"P/E Mix"
"eresource":{
"createInventory":"Instance"
}

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
KVupp: Ahh, good point, I missed this scenario

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
piotr-kalashuk: As I mentioned this is optional change...
By the way I think for the sample above `isEresourceInventoryNotPresent` will return `false` as the method checks only `eresource` section...

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
```diff
@@ -168,30 +166,84 @@ private CompositePurchaseOrder validateOrderState(CompositePurchaseOrder po) {
     }
   }
 
+  public static boolean isCreateInventoryNull(CompositePoLine compPOL) {
+    if (null == compPOL.getOrderFormat()) {
```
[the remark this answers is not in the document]
KVupp: I'd be in favor of that

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -585,15 +585,15 @@ private JsonObject buildInstanceRecordJsonObject(CompositePoLine compPOL, Map<St
   }
 
   private String getPhysicalMaterialTypeId(CompositePoLine compPOL) {
-    // the logic will be updated in scope of MODORDERS-195
-    return getElectronicMaterialTypeId(compPOL);
+    return Optional.ofNullable(compPOL.getPhysical())
```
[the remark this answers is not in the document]
KVupp: Done

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -231,13 +231,17 @@ public void putOrdersOrderLinesById(String lineId, String lang, CompositePoLine
...
       poLine.setId(lineId);
     }
 
     // First validate content of the PO Line and proceed only if all is okay
-    List<Error> errors = new ArrayList<>(validatePoLine(poLine));
+    List<Error> errors = new ArrayList<>();
+    helper.setTenantDefaultCreateInventoryValues(poLine).thenAccept(empty -> {
+      errors.addAll(validatePoLine(poLine));
+
     if (!lineId.equals(poLine.getId())) {
```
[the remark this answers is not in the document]
KVupp: Done

src/test/java/org/folio/rest/impl/OrdersImplTest.java
KVupp: Done

src/test/java/org/folio/rest/impl/OrdersImplTest.java
KVupp: Done

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
KVupp: correct, that is the expected behavior because we don't want to set the createInventory for eresource in this case, and set it only for physical

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
KVupp: Added a new method to handle the scenario

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -2381,8 +2381,6 @@ public void testPostOrdersLinePhysicalFormatDiscount() {
     // Check that no any calls made by the business logic to other services
     assertTrue(MockServer.serverRqRs.isEmpty());
   }
-
-  @Test
```
piotr-kalashuk: Looks like this is deleted by accident. Please add it back.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
KVupp: Even after deleting this block, the tests seem to pass. Looks like we have to revisit the unit test case

src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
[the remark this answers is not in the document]
piotr-kalashuk: Probably worth to include this change to https://github.com/folio-org/acq-models/pull/110 just to not come back to it later. But this is up to you...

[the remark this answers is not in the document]
KVupp: Made orderFormat required and also modified the required test data

PR #133, opened 2019-04-02
Modorders 189 No correlation between location and cost quantity

pautau-s:
## Purpose
[MODORDERS-189](https://issues.folio.org/browse/MODORDERS-189) Remove the ability to create or update poLine with:
- locations and cost quantities mismatch
- invalid value of location quantities fields eg. specify field `locations.quantityPhysical` for electronic order etc.

## Approach
- added validation for matching locations and cost quantities 
- added validation for the absence of invalid values for physical and electronic quantities of locations
-  Ignore `location.quantity` from POST and PUT requests and calculate it from `location.quantityElectronic` and `location.quantityPhysical` depending on `orderFormat`

The locations quantities checks are activated in the following cases: 
- if `createInventory `of any resource is "Instance, Holding" or "Instance, Holding, Item"
- if locations property not empty

## Pre-Merge Checklist:


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -57,9 +57,9 @@
   public static final String DEFAULT_POLINE_LIMIT = "1";
   private static final String MAX_POLINE_LIMIT = "500";
   public static final String OKAPI_URL = "X-Okapi-Url";
-  public static final String PO_LINES_LIMIT_PROPERTY = "poLines-limit";
+  private static final String PO_LINES_LIMIT_PROPERTY = "poLines-limit";
```
craigmcnally: why does this need to be private?

pautau-s: The only reason is that this constant is not used anywhere except in this class.

src/test/java/org/folio/rest/impl/OrdersImplTest.java
```diff
@@ -3582,6 +3626,24 @@ private Response verifyPut(String url, String body, String expectedContentType,
             .response();
   }
 
+
+  private void verifyLocationQuantity(Location location, CompositePoLine.OrderFormat orderFormat) {
+    switch (orderFormat) {
+      case P_E_MIX:
+        assertEquals(location.getQuantityPhysical() + location.getQuantityElectronic(), location.getQuantity().intValue());
```
craigmcnally: It would be nice if we were consistent wrt using `assertThat(..., equalT(...))` and `assertEquals(...)`

src/main/java/org/folio/orders/utils/HelperUtils.java
craigmcnally: I guess it's a personal preference, but I find it silly to limit scope of static constants as there's no benefit IMO.  Additionally, if you ever want to refer to them later on (e.g. from a unit test or other class) it would require changing the scope again.

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -320,37 +320,68 @@ public static String getEndpointWithQuery(String query, Logger logger) {
...
+  public static int getPhysicalCostQuantity(CompositePoLine compPOL) {
+    return defaultIfNull(compPOL.getCost().getQuantityPhysical(), 0);
+  }
+
+  public static int getElectronicCostQuantity(CompositePoLine compPOL) {
+    return defaultIfNull(compPOL.getCost().getQuantityElectronic(), 0);
+  }
+
+  private static List<ErrorCodes> validateLocations(CompositePoLine compPOL) {
+    if (isLocationsValidationRequired(compPOL)) {
```
piotr-kalashuk: Please check `createInventory` separately for Physical and Electronic resources.
In case the PO Line is `P/E Mix` and physical.createInventory is "Instance, Holding, Item" (or "Instance, Holding") but `eresource.createInventory` is "Instance" (or "None"), an user might want to fill in locations info only for physical quantities but validation will force to specify electronic quantities there as well...

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -375,57 +406,52 @@ private static void validateCostPrices(Cost cost, CompositePoLine.OrderFormat or
...
-    if (discount < 0d || (cost.getDiscountType() == Cost.DiscountType.PERCENTAGE && discount > 100d)
-      // validate that discount does not exceed total price
-      || (discount > 0d && cost.getDiscountType() == Cost.DiscountType.AMOUNT && calculateEstimatedPrice(cost) < 0d)) {
+    if (isDiscountExceedTotalPrice(cost)) {
       errors.add(ErrorCodes.COST_DISCOUNT_INVALID);
     }
+    return errors;
+  }
+
+  private static boolean isDiscountExceedTotalPrice(Cost cost) {
```
piotr-kalashuk: (nitpick) The method checks also if discount is not negative which does not correspond to method name

src/main/java/org/folio/orders/utils/HelperUtils.java
pautau-s: Done

src/main/java/org/folio/orders/utils/HelperUtils.java
pautau-s: Renamed

PR #134, opened 2019-04-03
MODORDERS-204 - When creating a brief order instance record in Inventory, data on Contributors are not populated

khandramai:
[MODORDERS-204](https://issues.folio.org/browse/MODORDERS-204) - When creating a brief order instance record in Inventory, data on Contributors are not populated

## Purpose
For fixing issue when changes the order status from 'Pending' to 'Open', and a brief order instance record in Inventory is created, then data on Contributors are not being populated.

## Approach
Enable support of contributors populating from CompositePoLine to Instance record JsonObject in the InventoryHelper.buildInstanceRecordJsonObject(...)

#### TODOS and Open Questions
Additionally, one need to implement support of populating of "contributorNameTypeId" in the PO Line on UI side.

## Pre-Merge Checklist:


khandramai:
Verified on local folio/testing box:
Step-1
Step-2
Step-3

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -865,4 +865,21 @@ public static boolean isHoldingsUpdateRequired(CompositePoLine compPOL) {
   public static boolean isItemsUpdateRequired(CompositePoLine compPOL) {
     return isItemsUpdateRequiredForPhysical(compPOL) || isItemsUpdateRequiredForEresource(compPOL);
   }
+
+  public enum ContributorNameTypeName {
```
piotr-kalashuk: Please move this to InventoryHelper as this is inventory specific info

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -865,4 +865,21 @@ public static boolean isHoldingsUpdateRequired(CompositePoLine compPOL) {
...
+    MEETING_NAME("Meeting name"),
+    PERSONAL_NAME("Personal name");
+
+    ContributorNameTypeName(String name) {
+      this.name = name;
+    }
+
+    private String name;
+
+    public String getContributorNameTypeName() {
```
piotr-kalashuk: Better to rename just to `getName`

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -626,4 +644,28 @@ private JsonObject getFirstObjectFromResponse(JsonObject response, String proper
...
+  /**
+   * Returns JsonObject with id of ContributorNameType retrieved from inventory-storage.
+   *
+   * @param contributorNameTypeName ContributorNameType
+   * @return JsonObject ContributorNameTypeId
+   */
+  private CompletableFuture<JsonObject> getContributorNameTypeId(ContributorNameTypeName contributorNameTypeName) {
+    String name = contributorNameTypeName.getContributorNameTypeName();
+    String endpoint = String.format("/contributor-name-types?query=name==%s", encodeQuery(name, logger));
+    if(ctx.get(name) == null) {
```
piotr-kalashuk: Please add the same logic to `getInstanceType` and `getStatus` just to be consistent

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -626,4 +644,28 @@ private JsonObject getFirstObjectFromResponse(JsonObject response, String proper
...
+
+  /**
+   * Returns JsonObject with id of ContributorNameType retrieved from inventory-storage.
+   *
+   * @param contributorNameTypeName ContributorNameType
+   * @return JsonObject ContributorNameTypeId
+   */
+  private CompletableFuture<JsonObject> getContributorNameTypeId(ContributorNameTypeName contributorNameTypeName) {
+    String name = contributorNameTypeName.getContributorNameTypeName();
+    String endpoint = String.format("/contributor-name-types?query=name==%s", encodeQuery(name, logger));
```
piotr-kalashuk: Move this line into `if` block

src/main/java/org/folio/rest/impl/InventoryHelper.java
khandramai: Hi @piotr-kalashuk,
this case I can suggest generalizing of retrieving and caching logic for `getInstanceType`, `getStatus` and `getContributorNameTypeId`.

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -626,4 +643,61 @@ private JsonObject getFirstObjectFromResponse(JsonObject response, String proper
...
   }
+
+  /**
+   * Returns JsonObject with id of ContributorNameType retrieved from inventory-storage.
+   *
+   * @param contributorNameTypeName ContributorNameType
+   * @return JsonObject ContributorNameTypeId
+   */
+  private CompletableFuture<JsonObject> getContributorNameTypeId(ContributorNameTypeName contributorNameTypeName) {
+    return cacheAndGet(contributorNameTypeName.getName(),
```
piotr-kalashuk: As you are using function as handler you might pass function to construct endpoint whe needed like
```java
  private CompletableFuture<JsonObject> getContributorNameTypeId(ContributorNameTypeName contributorNameTypeName) {
    UnaryOperator<String> endpoint = query -> String.format("/contributor-name-types?query=name==%s", encodeQuery(query, logger));
    return cacheAndGet(contributorNameTypeName.getName(), endpoint,
      entries -> {
        JsonObject contributorPersonalNameType = new JsonObject();
        contributorPersonalNameType.put(CONTRIBUTOR_NAME_TYPE_ID, getFirstObjectFromResponse(entries, CONTRIBUTOR_NAME_TYPES).getString(ID));
        return contributorPersonalNameType;
      });
  }
...
  private CompletableFuture<JsonObject> cacheAndGet(String key, UnaryOperator<String> endpoint, Function<JsonObject, JsonObject> handler) {
    JsonObject response = ctx.get(key);
    if(response == null) {
      return handleGetRequest(endpoint.apply(key), httpClient, ctx, okapiHeaders, logger).thenApply(json -> {
        JsonObject result = handler.apply(json);
        ctx.put(key, result);
        return result;
      });
    } else {
      CompletableFuture<JsonObject> future = new CompletableFuture<>();
      future.complete(ctx.get(key));
      return future;
    }
  }
```

src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -626,4 +646,60 @@ private JsonObject getFirstObjectFromResponse(JsonObject response, String proper
...
+  /**
+   * Caches value in Vert.X Context and returns it by key.
+   *
+   * @param key key for retrieving value from cache
+   * @param endpointConstructor endpoint for GET request to retrieve value
+   * @param fn Function<JsonObject, JsonObject> for transformation JsonObject to new one if needed
+   *
+   * @return value from cache
+   */
+  private CompletableFuture<JsonObject> cacheAndGet(String key, UnaryOperator<String> endpointConstructor, Function<JsonObject, JsonObject> fn) {
```
craigmcnally: I would think "getAndCache" might be a better name, but it's certainly not a big deal, especially given simplicity of the function.

PR #137, opened 2019-04-09
MODORDERS-209 fix for the hanging orders bug

grabsefx:
https://issues.folio.org/browse/MODORDERS-209
## Purpose
Bugfix needed for the case of order creation with `poLine.physical.createInventory/poLine.eresource.createInventory` not specified in the order and the tenant-specific settings are empty.
## Approach
* fixing future completion, adding unit test to cover this piece of code

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
+++ b/src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
@@ -156,7 +156,7 @@ class PurchaseOrderLineHelper extends AbstractHelper {
           if (!config.isEmpty() && !config.getString(CREATE_INVENTORY).isEmpty()) {
             return future.complete(new JsonObject(config.getString(CREATE_INVENTORY)));
           } else {
-            return completedFuture(new JsonObject());
+            return future.complete(new JsonObject());
           }
         })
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/PurchaseOrderLineHelper.java
```diff
@@ -156,7 +156,7 @@ private CompositePurchaseOrder validateOrderState(CompositePurchaseOrder po) {
           if (!config.isEmpty() && !config.getString(CREATE_INVENTORY).isEmpty()) {
```
piotr-kalashuk: As far as I understand this issue happened because when `config.isEmpty()` is `true`, the logic went to `else` section and brand new completed `CompletableFuture` was created...
However when `config` has some values, the second condition 
```java
config.getString(CREATE_INVENTORY).isEmpty()
```
fails with NPE because `config.getString(CREATE_INVENTORY)` returns `null` so the `exceptionally` block completes `future` defined above. Probably better to replace second (or both) condition by
```java
StringUtils.isNotEmpty(config.getString(CREATE_INVENTORY))
```

grabsefx: condition fixed

PR #140, opened 2019-04-11
MODORDERS-210 - Make locationId as not required in Check-in flow

khandramai:
[MODORDERS-210](https://issues.folio.org/browse/MODORDERS-210) - Make locationId as not required in Check-in flow

## Purpose
After removing locationId from required fields [PR#114](https://github.com/folio-org/acq-models/pull/114) one need to validate that locationId is provided when createInventory is set to "Instance, Holding" or "Instance, Holding, Item". If not, an appropriate error message & code are returned.

## Approach
1. Logic for checking locationId presence implemented as part of checkin flow. If the conditions are not met, the corresponding error is added without finishing flow execution.
2. POLines added as  property of Receiving/Checkin helper because we need to retrieve this information several times during checkin flow execution.
3. Unit test implemented.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/CheckinHelper.java
+++ b/src/main/java/org/folio/rest/impl/CheckinHelper.java
@@ -13,7 +13,6 @@ import java.util.Map;
 import java.util.Date;
 import java.util.concurrent.CompletableFuture;
 
-import me.escoffier.vertx.completablefuture.VertxCompletableFuture;
 import one.util.streamex.StreamEx;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.rest.acq.model.Piece;
@@ -49,9 +48,8 @@ public class CheckinHelper extends CheckinReceivePiecesHelper<CheckInPiece> {
 
     // 1. Get piece records from storage
     return retrievePieceRecords(checkinPieces)
-      // 2. Check locationId presence for pieces related to
-      // POLine with createInventory "Instance, Holding" or "Instance, Holding, Item"
-      .thenCompose(this::checkLocationId)
+      // 2. Filter locationId
+      .thenCompose(this::filterMissingLocations)
       // 3. Update items in the Inventory if required
       .thenCompose(this::updateInventoryItems)
       // 4. Update piece records with checkIn details which do not have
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/rest/impl/CheckinHelper.java
```diff
@@ -47,19 +49,48 @@
...
+  private CompletableFuture<Map<String, List<Piece>>> checkLocationId(Map<String, List<Piece>> piecesRecords) {
+    getPoLines(piecesRecords)
+      .thenAccept(poLinesReply -> {
+        for (PoLine poLine : poLinesReply) {
+          Eresource.CreateInventory eci = poLine.getEresource() != null ? poLine.getEresource().getCreateInventory() : null;
+          Physical.CreateInventory pci = poLine.getPhysical() != null ? poLine.getPhysical().getCreateInventory() : null;
+          if (Eresource.CreateInventory.INSTANCE_HOLDING.equals(eci) || Eresource.CreateInventory.INSTANCE_HOLDING_ITEM.equals(eci)
+            || Physical.CreateInventory.INSTANCE_HOLDING.equals(pci) || Physical.CreateInventory.INSTANCE_HOLDING_ITEM.equals(pci)) {
+            for (CheckInPiece cip : checkinPieces.get(poLine.getId()).values()) {
+              if (cip.getLocationId() == null) {
```
piotr-kalashuk: 1. Seems "P/E Mix" is not considered here. For such PO Lines E- and Physical resources are fully independent. To validate properly records for such PO Line piece format needs to be taken into account as well.
2. When looking at `updatePieceWithCheckinInfo` method here if location is not provided in the request for particular piece, it remains unchanged:
```java
    if (StringUtils.isNotEmpty(checkinPiece.getLocationId())) {
      piece.setLocationId(checkinPiece.getLocationId());
    }
```
So the entry in the request like following
```json
{
  "id": "71d9322b-5cdd-45d8-ad45-c7f3044802e7",
  "itemStatus":"Received"
}
```
means leave location unchanged but just check-in piece. So validation needs to be a bit enhanced.
@folio-org/acquisitions - please have a look if you agree with such approach.

craigmcnally: +1 wrt handling of P/E Mix.

> if location is not provided in the request for particular piece, it remains unchanged
I don't think this is the way it should work.  In cases where location is required (based on createInventory), location must be specified.  If not, an error should be returned.

src/main/java/org/folio/rest/impl/CheckinHelper.java
```diff
@@ -47,19 +49,59 @@
...
+          for (Piece piece : pieces) {
+            // Check if locationId doesn't presented in piece from request and retrieved from storage
+            // Corresponding piece from collection
+            CheckInPiece pieceFromRq = checkinPieces.get(poLine.getId()).get(piece.getId());
+            if (pieceFromRq.getLocationId() == null && piece.getLocationId() == null) {
+              if (piece.getFormat() == Piece.Format.ELECTRONIC) {
+                // Check Eresource
+                if (poLine.getEresource() != null && poLine.getEresource().getCreateInventory() != Eresource.CreateInventory.NONE
+                  && poLine.getEresource().getCreateInventory() != Eresource.CreateInventory.INSTANCE) {
+                  addError(poLine.getId(), piece.getId(), LOC_NOT_PROVIDED.toError());
```
piotr-kalashuk: As far as I can see even though the error is added, the piece remains in the `piecesRecords` map and will be processed by further logic like any other valid piece. In this case piece (and linked item) will be updated but the logic which prepares response will mark that piece as failed...
Probably better to rename the method to something like `filterMissingLocations` and instead of
```java
for (Piece piece : pieces) {
```
use
`pieces.removeIf(Predicate)` where Predicate contains this logic and returns boolean as a result

src/main/java/org/folio/rest/impl/CheckinHelper.java
piotr-kalashuk: Current approach for check-in/receiving flows contains 2 steps:

1. Create piece (and item).
   - For check-in an user can specify location along with other data (caption, comment, itemId etc.).
   - For receiving flow the location will be always specified on Order transition to Open.
2. Check-in/receive that piece. At this stage location might be already specified in piece record. That is why location might be omitted in the request

UI works with the piece data to preselect location in the DDL and an user can change it when checking-in/receiving a piece. So it is expected that the location actually will be forced to be provided by UI anyway.
Forcing `locationId` in the check-in/receiving request (if `createInventory` of the PO Line requires is) by business logic is a bit inconsistent with other properties like `comment` or `caption`. But taking into account that at the moment the only consumer of the receiving/check-in API is UI app, the business logic mostly is based on what UI sends...

src/main/java/org/folio/rest/impl/CheckinHelper.java
```diff
@@ -47,19 +49,59 @@
...
+  private CompletableFuture<Map<String, List<Piece>>> checkLocationId(Map<String, List<Piece>> piecesRecords) {
+    getPoLines(piecesRecords)
+      .thenAccept(poLines -> {
+        for(PoLine poLine : poLines) {
+          List<Piece> pieces = piecesRecords.get(poLine.getId());
+          for (Piece piece : pieces) {
+            // Check if locationId doesn't presented in piece from request and retrieved from storage
+            // Corresponding piece from collection
+            CheckInPiece pieceFromRq = checkinPieces.get(poLine.getId()).get(piece.getId());
+            if (pieceFromRq.getLocationId() == null && piece.getLocationId() == null) {
```
piotr-kalashuk: The API can be used to check-in piece or revert already checked-in piece. I believe we do not need to force location id to be provided if piece needs to become `Expected` (and related item "On order" if available)

src/main/java/org/folio/rest/impl/CheckinHelper.java
khandramai: > I don't think this is the way it should work. In cases where location is required (based on createInventory), location must be specified. If not, an error should be returned.

Current implementation works without piece locationId validation, i.e. error will be returned if only CheckIn request doesn't contain info about locationId.

PR #144, opened 2019-04-18
MODORDERS-210 - Make locationId as not required in Check-in flow

khandramai:
Fixing issue with failed piece removing from piece records map for [PR#140](https://github.com/folio-org/mod-orders/pull/140)


khandramai:
Fix verified on localhost: [MODORDERS-210.docx](https://github.com/folio-org/mod-orders/files/3094005/MODORDERS-210.docx)

piotr-kalashuk:
please update "unit" test to verify also number of requests made to other modules to guarantee that such issue is fixed

PR #147, opened 2019-04-24
MODORDERS-222 - Order Line has inactive Access Provider

khandramai:
[MODORDERS-222](https://issues.folio.org/browse/MODORDERS-222) - Order Line has inactive Access Provider

## Purpose
PO Line numbers should be displayed in the error modal if associated access provider is inactive/not found.

## Approach
Included poLineNumber population to error response during Vendor/AccessProvider validation.

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -59,15 +62,15 @@ public VendorHelper(Map<String, String> okapiHeaders, Context ctx, String lang)
         .thenApply(vendor -> {
           VendorStatus status = VendorStatus.valueOf(vendor.getVendorStatus().toUpperCase());
           if(status != VendorStatus.ACTIVE) {
-            errors.add(createErrorWithId(ORDER_VENDOR_IS_INACTIVE, id));
+            errors.add(createErrorWithId(ORDER_VENDOR_IS_INACTIVE, id, poLines));
```
craigmcnally: vendor is at the PO level, I'm not sure it makes sense to add an error for every single POL

khandramai: @craigmcnally, done. Thanks!

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -150,6 +153,24 @@ private Error createErrorWithId(ErrorCodes errorCodes, String id) {
...
+   *
+   * @param id vendor's/access provider's identifier
+   * @param errorCodes error code
+   * @param poLines list of composite PoLines
+   * @return {@link Error} with id of failed vendor/access provider
+   */
+  private Error createErrorWithId(ErrorCodes errorCodes, String id, List<CompositePoLine> poLines) {
+    Error error = createErrorWithId(errorCodes, id);
+    for (CompositePoLine poLine : poLines) {
+      if (poLine.getPoLineNumber() != null) {
```
piotr-kalashuk: This logic does not take into account if the PO Line's access provider id equals to vendor `id` but just adds all PO Line numbers as parameters.
I verified this by `PurchaseOrdersApiTest#testPutOrdersByIdToChangeStatusToOpenInactiveVendor` unit test adding in the beginning of the test
```java
    for (int i = 0; i < reqData.getCompositePoLines().size(); i++) {
      reqData.getCompositePoLines().get(i).setPoLineNumber("number-" + i);
    }
```
After updates above the test fails when there is one active and one inactive access provider specified

src/main/java/org/folio/rest/impl/VendorHelper.java
```diff
@@ -91,7 +92,12 @@ public VendorHelper(Map<String, String> okapiHeaders, Context ctx, String lang)
   public CompletableFuture<Errors> validateAccessProviders(CompositePurchaseOrder compPO) {
     CompletableFuture<Errors> future = new VertxCompletableFuture<>(ctx);
     Set<String> ids = verifyAndGetAccessProvidersIdsList(compPO);
```
piotr-kalashuk: Looks like this set is no longer needed because this is the same as `poLinesMap.keySet()`

PR #148, opened 2019-04-25
MODORDERS-218 adjust order status based on poLine's statuses

piotr-kalashuk:
## Purpose
[MODORDERS-218](https://issues.folio.org/browse/MODORDERS-218) Automatically adjust order status based on poLine's paymentStatus and receivingStatus

## Approach
- The order status handler is registered for event bus consumer. The handler receives array of order ids and performs following logic:
  - get order by id.
  - if the order workflow status is not pending, get all the PO Lines of this order
  - in case conditions described in MODORDERS-218 are matched, the PO is updated
- The event bus message is sent only in following cases
  - particular PO Line is successfully updated
  - receiving/check-in successfully resulted to PO Line receiving status update
- Unit tests updates:
  - the mock version of the handler is used which just registers message received
  - verify that event message is registered only if it is expected (see conditions above). If the message is registered, the expected content is verified
  - the handler is covered by separate unit tests
- POM dependencies added: `folio-di-support` (spring dependency injection) and `awaitility`

## Learning
- [folio-di-support](https://github.com/folio-org/folio-di-support) - Dependency Injection support for FOLIO back-end modules
- [Awaitlity Usage](https://github.com/awaitility/awaitility/wiki/Usage) -  a library which provides a simple domain-specific language (DSL) for asynchronous systems testing.

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/InitEventBus.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> resultHandler) {
+    vertx.executeBlocking(
+      handler -> {
+        EventBus eb = vertx.eventBus();
+        eb.consumer(MessageAddress.ORDER_STATUS.address, orderStatusHandler);
+
+        handler.complete();
```
KVupp: Isn't it better to complete the handler with the result from the eventbus registration?

piotr-kalashuk: This is just to register consumer and a handler which is going to be used when any message is sent to the address. I.e. the handler will be called only and as many times when the message is sent. This is why `handler.complete()` is called as standalone operation just to let vertx know that all consumers created (for now it is just one but might be more defined later).

piotr-kalashuk:
> 
> 
> I know this is not mentioned in the story and the API listing doc could be out of date, but looks like PUT **/orders/composite-orders/{id}** will be called by UI orders. so If I update the payment status to **"Fully Paid"** and receipt status to "**fully Received**" using this endpoint the order status does not change automatically. Is this an expected behavior?

At the moment UI sends `PUT /orders/composite-orders/{id}` requests just to update order content but PO Lines might not be sent at all or their content is the same as in storage. Currently UI sends `PUT /orders/order-lines/{id}` requests when user updates PO Line's content. This is why I decided to stick to the cases from jira description to send events.

src/main/java/org/folio/rest/impl/InitEventBus.java
KVupp: Correct, I understand that, but I was just wondering of a scenario where the consumer registration could be delayed but we are going ahead and completing the operation. It usually happens in a clustered event bus but probably the scenario is rare or doesn't even occur in our case, but just a thought.

piotr-kalashuk: Now I see what you meant. I applied change so now `localConsumer` is used just to not propagate the consumer to cluster (if deployed in such mode). Also I added logic to complete event bus init logic only when consumer is registered. In case in future this requires changes, it can be easily done here.

PR #153, opened 2019-05-02
MODORDERS-229 - Sort PO Lines array by POLNumber in composite PO GET by id response

khandramai:
[MODORDERS-229](https://issues.folio.org/browse/MODORDERS-229) - Sort PO Lines array by POLNumber in composite PO GET by id response

## Purpose
To support default sorting in PO Lines list with minimal efforts.

## Approach

Approach | Advantages | Disadvantages
-- | -- | --
Custom "comparator" in the PostgreSQL | "Native approach" | Hard implementation related to work on DB side
PoLineNumber format changing - adding zeroes at the begin of poLineNumber suffix |Simple implementation | This approach relates with additional difficulties in  the implementation of this approach, since it is necessary to provide  additional comparison logic and to process the string value for correct  displaying.
Comparator for suffix in mod-orders code | Simple implementation | Negative performance impact

Comparator for suffix in mod-orders code for sorting PoLines seems more simple. Sorting looks like, for example: 841240-1, 841240-02, 841240-2, 841240-21.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
@@ -52,7 +52,6 @@ public class PurchaseOrderHelper extends AbstractHelper {
 
   private static final String SEARCH_PURCHASE_ORDERS = resourcesPath(SEARCH_ORDERS) + "?limit=%s&offset=%s%s&lang=%s";
   private static final String GET_PURCHASE_ORDERS = resourcesPath(PURCHASE_ORDER) + "?limit=%s&offset=%s&lang=%s";
-  private static final String DASH = "-";
 
   private final PoNumberHelper poNumberHelper;
   private final PurchaseOrderLineHelper orderLineHelper;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -561,4 +566,27 @@ private boolean hasNewPoLines(CompositePurchaseOrder compPO, JsonArray poLinesFr
       .stream()
       .filter(poLine -> !lineIdsInStorage.contains(poLine.getId()));
   }
+
+  private static int comparePoLinesByPoLineNumber(CompositePoLine poLine1, CompositePoLine poLine2) {
```
piotr-kalashuk: Please move these 2 methods to `PurchaseOrderLineHelper`

khandramai:
Improved on the assumption that the prefixes should be the same, and the comparison should be carried out by suffixes that are integer values with a maximum numeric digits equal to three.

PR #158, opened 2019-05-15
Make required inventory types configurable

piotr-kalashuk:
## Purpose
- [MODORDERS-239](https://issues.folio.org/browse/MODORDERS-239) Make `instance-type` configurable
- [MODORDERS-240](https://issues.folio.org/browse/MODORDERS-240) Make `instance-status` configurable
- [MODORDERS-241](https://issues.folio.org/browse/MODORDERS-241) Make `loan-type` configurable

## Approach
The expected inventory types values are now checked in tenant specific configs and default values are used only if nothing found:

Type| Config Code | System Default Value
------------ | ------------- | -------------
 |  |
Instance type|`inventory-instanceTypeCode`|`zzz`
Instance status|`inventory-instanceStatusCode`|`temp`
Loan type|`inventory-loanTypeName`|`Can circulate`

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/InventoryHelper.java
```diff
@@ -68,42 +68,49 @@
   static final String CONTRIBUTOR_NAME_TYPES = "contributorNameTypes";
   static final String INSTANCE_STATUSES = "instanceStatuses";
   static final String INSTANCE_TYPES = "instanceTypes";
-
   static final String ITEMS = "items";
+  static final String LOAN_TYPES = "loantypes";
```
javalkar-varun: Following camelCase naming convention, should this be `loanTypes` ?

piotr-kalashuk: This is in accordance with [inventory loantypes.json](https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/loantypes.json#L6) so we cannot change the value

javalkar-varun: Also, check API docs, its not camelCase https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/loan-type.html#loan_types_get

PR #176, opened 2019-06-20
MODORDERS-173 - ensure receiptStatus consistency between piece and poLine

javalkar-varun:
## Purpose

https://issues.folio.org/browse/MODORDERS-173

## Approach

receiptStatus inconsistency between Pieces and PoLine can occur when updating a Piece. Following is a flow to handle it: 
 1. Before updating the piece, get a piece by id from storage and store its receiving status
 2. Update piece with new content
 3. Create a message and check if receivingStatus is not consistent with storage; if yes - send an event to `org.folio.orders.po-line.update.receipt-status`
4. Get all pieces for the poLineId
5. Get the PO Line using poLine Id
6. Recalculate receipt status using pieces and PoLine
7. Update poLine in storage
8. Check to make sure if purchase order's order status has changed by sending an event to `org.folio.orders.order.update.status`; if yes - then update the order status in storage
9. Cleanup redundant code
10. Add/update unit tests

## Pre-Merge Checklist:


src/main/java/org/folio/orders/events/handlers/MessageAddress.java
```diff
@@ -2,7 +2,8 @@
 
 public enum MessageAddress {
 
-  ORDER_STATUS("org.folio.orders.order.update.status");
+  ORDER_STATUS("org.folio.orders.order.update.status"),
+  RECEIPT_STATUS("org.folio.orders.receipt.status");
```
craigmcnally: I'd follow the same general naming convention... maybe something like:

`org.folio.orders.order-line.update.receiptStatus` or 
`org.folio.orders.order-line.update.receipt-status` or
`org.folio.orders.po-line.update.receiptStatus` or 
`org.folio.orders.po-line.update.receipt-status`

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,195 @@
...
+
+      // 2. if receivingStatus is different - before writing to storage vs after writing to storage
+      if (!receivingStatus.toString()
+        .equalsIgnoreCase(receivingStatusBeforeUpdate)) {
+        String poLineId = piece.getPoLineId();
+
+        String endpoint = String.format(PIECES_ENDPOINT, poLineId, LANG);
+
+        // 3. Get all pieces for poLineId above
+        getPieces("", endpoint, httpClient, ctx, okapiHeaders, logger).thenAccept(piecesCollection -> {
```
craigmcnally: Why you bake the query into the endpoint above and pass in an empty string for query here?

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,195 @@
...
+
+    // 1. Get updated piece by id from storage
+    getPieceById(pieceIdUpdate, lang, httpClient, ctx, okapiHeaders, logger).thenAccept(jsonPiece -> {
+      Piece piece = jsonPiece.mapTo(Piece.class);
+      logger.info("piece for pieceRecord.json" + piece + " pieceById: " + pieceIdUpdate);
+      ReceivingStatus receivingStatus = piece.getReceivingStatus();
+      String receivingStatusBeforeUpdate = body.getString(RECEIVING_STATUS_BEFORE_UPDATE);
+
+      // 2. if receivingStatus is different - before writing to storage vs after writing to storage
+      if (!receivingStatus.toString()
```
craigmcnally: Why compare as String instead of the enum?

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,195 @@
...
+    final Integer expectedQty = count;
+    return CompletableFuture.supplyAsync(() -> expectedQty);
+  }
+
+  CompletableFuture<PieceCollection> getPieces(String query, String path, HttpClientInterface httpClient, Context ctx,
+      Map<String, String> okapiHeaders, Logger logger) {
+    CompletableFuture<PieceCollection> future = new VertxCompletableFuture<>(ctx);
+    try {
+      String queryParam = isEmpty(query) ? EMPTY : "&query=" + encodeQuery(query, logger);
+      String endpoint = String.format(path, LIMIT, OFFSET, queryParam, LANG);
```
craigmcnally: I think this means you'll only get up to the hardcoded `LIMIT` number of records.  TBH it's probably not going to be a common occurrence for there to be more than 500 Pieces for a given poLine, but I think it's possible and should be handled appropriately.

javalkar-varun: To avoid hardcoding, the only way I know is to get total piece records is to query `/orders-storage/pieces` which is another call to the orders-storage module. Is there a better way to get a total piece count without hitting orders-storage module?

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: Done, using enum to compare

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: Done, the refactored code to avoid passing an empty string for the query.

src/main/java/org/folio/orders/events/handlers/MessageAddress.java
javalkar-varun: Using `org.folio.orders.po-line.update.receipt-status`

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
craigmcnally: Instead of asking for 500, ask for a smaller, more reasonable number, say 10 - 20.  In the response you'll get the total number of results.  Make additional calls while increasing the offset by the appropriate amount until you have all of the results.

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,189 @@
...
+    // Now wait for all operations to be completed and send reply
+    completeAllFutures(ctx, futures, message);
+  }
+
+
+  private CompletableFuture<PoLine.ReceiptStatus> calculatePoLineReceiptStatus(PoLine poLine,
+      List<org.folio.rest.acq.model.Piece> pieces) {
+
+    if (pieces.isEmpty())
+      return completedFuture(poLine.getReceiptStatus());
```
craigmcnally: inconsistent to not use `{`. `}`

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,189 @@
...
+      .anyMatch(piece -> ReceivingStatus.RECEIVED == piece.getReceivingStatus())) {
+      return CompletableFuture.completedFuture(PARTIALLY_RECEIVED);
+    }
+    // Pieces were rolled-back to Expected. In this case we have to check if
+    // there is any Received piece in the storage
+    return getPiecesQuantityByPoLineAndStatus(ReceivingStatus.RECEIVED, pieces)
+      .thenApply(receivedQty -> receivedQty == 0 ? AWAITING_RECEIPT : PARTIALLY_RECEIVED);
+  }
+
+  private CompletableFuture<Integer> getPiecesQuantityByPoLineAndStatus(ReceivingStatus receivingStatus, List<Piece> pieces) {
```
craigmcnally: Again, it's an enum, no need to compare as String...  Can be simplified:
```
    return CompletableFuture.supplyAsync(() -> (int) pieces.stream()
      .filter(piece -> piece.getReceivingStatus() == receivingStatus)
      .count());
```

src/test/java/org/folio/orders/events/handlers/ReceiptStatusConsistencyTest.java
```diff
@@ -0,0 +1,147 @@
...
+    // explicitly setting RECEIVED to create inconsistency
+    sendEvent(createBody(RECEIVED, VALID_PIECE_ID), context.asyncAssertSuccess(result -> {
+      logger.info("getPoLineSearches()--->" + getPoLineSearches());
+      logger.info("getPoLineUpdates()--->" + getPoLineUpdates());
+      logger.info("getPieceSearches()--->" + getPieceSearches());
+      assertEquals(2, getPieceSearches().get(0).getJsonArray("pieces").size());
+      
+      Piece piece0 = getPieceSearches().get(0).getJsonArray("pieces").getJsonObject(0).mapTo(Piece.class);
+      Piece piece1 = getPieceSearches().get(0).getJsonArray("pieces").getJsonObject(1).mapTo(Piece.class);
+      assertEquals(piece0.getReceivingStatus().toString().toLowerCase(),ReceivingStatus.EXPECTED.value().toLowerCase());
```
craigmcnally: no need for the toString/toLowerCase stuff... just compare the enum values.

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: Done

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: Done

src/test/java/org/folio/orders/events/handlers/ReceiptStatusConsistencyTest.java
javalkar-varun: Done

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: I have updated logic to get all pieces by using Integer.MAX_VALUE as the limit. This should take care even if we have more than 500 pieces. Offset/pageable will introduce unnecessary complexity by making recurrent calls to GET all pieces if `getTotalRecords` exceeds a smaller reasonable set LIMITS.

craigmcnally: @folio-org/acquisitions does anyone know off-hand if we use paging in other places in acquisitions?  Using a giant limit param doesn't seem ideal to me, if there are enough records and the storage module is slow enough you run the risk of the request timing out.  It also places additional memory strain on the storage module as it would need to load all of those records into memory before returning them.

piotr-kalashuk: As far as I know we do not use this pattern yet. Other option is to use 0 as limit parameter which basically means return no records at all but only count of such records i.e. `totalRecords`. For example, [CheckinReceivePiecesHelper#getPiecesQuantityByPoLineAndStatus(poLineId, receivingStatus)](https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/rest/impl/CheckinReceivePiecesHelper.java#L563) is designed such way.

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,178 @@
...
+      }
+    })
+      .exceptionally(e -> {
+        logger.error("The error happened getting a piece by id {}", pieceIdUpdate, e);
+        future.completeExceptionally(e);
+        return null;
+      });
+
+    // Now wait for all operations to be completed and send reply
+    completeAllFutures(ctx, httpClient, futures, message);
```
piotr-kalashuk: In case the logic of `ReceiptStatusConsistency` updated POL's `receiptStatus` to `Fully Received` (or wise versa - from `Fully Received` to `Partially Received` or `Awaiting Receipt`), potentially entire order status can be changed as well. I believe these cases should be covered as well and an event to `org.folio.orders.order.update.status` should be sent...

javalkar-varun: Should this event be sent from `AbstractHelper#updatePoLineReceiptStatus` where status changes to `FULLY_RECEIVED` or `AWAITING_RECEIPT` or `PARTIALLY_RECEIVED` ?

piotr-kalashuk: As far as I can see current implementation of receiving/check-in flow is unchanged and POL's status is being calculated as part of that flow (initially I thought that this jira was going move that logic to new handler).
As far as I understand in most of the cases order is going to have mainly one line. But our implementation assumes that order might have up to 500 PO lines (the POL number sequence is up to 999). Potentially there might be some libraries which are going to work with such big orders. 
When the pieces are being received might belong to different lines and in case there are a lot lines involved e.g. user receives all the resources for 100 POL's, the logic might trigger up to 100 events and `OrderStatus` handler will start fetching a lot of data from storage but only the last one might end up updating order's status...

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
craigmcnally: OK thanks @piotr-kalashuk.  Something to keep in mind going forward I guess

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
craigmcnally: > As far as I can see current implementation of receiving/check-in flow is unchanged and POL's status is being calculated as part of that flow (initially I thought that this jira was going move that logic to new handler).
I agree we should consolidate the calculation 

> When the pieces are being received might belong to different lines and in case there are a lot lines involved e.g. user receives all the resources for 100 POL's, the logic might trigger up to 100 events and OrderStatus handler will start fetching a lot of data from storage but only the last one might end up updating order's status...
Good point.  We need to be careful about where/when we fire the events from.

src/main/java/org/folio/rest/impl/PiecesHelper.java
```diff
@@ -26,12 +27,27 @@ public PiecesHelper(Map<String, String> okapiHeaders, Context ctx, String lang)
...
     return createRecordInStorage(JsonObject.mapFrom(entity), resourcesPath(PIECES)).thenApply(entity::withId);
   }
-  
+
   public CompletableFuture<Void> updatePieceRecord(Piece piece) {
-    return handlePutRequest(resourceByIdPath(PIECES, piece.getId()), JsonObject.mapFrom(piece), httpClient, ctx, okapiHeaders,
-        logger);
+    JsonObject messageToEventBus = new JsonObject();
+
+    messageToEventBus.put("receivingStatusBeforeUpdate", piece.getReceivingStatus());
```
piotr-kalashuk: Quick question. Let's assume that any client uses `PUT /orders/peices/<id>` endpoint to update content of the piece record along with status. Let's assume that status in the storage is `Expected` and passed `piece` record has `Received` status. The logic here gets its status (i.e. `Received`) and stores it in the JsonObject under `receivingStatusBeforeUpdate` key. Then we send PUT request to storage and once completed (i.e. the piece's status in the storage is already `Received`) trigger an event. As far as I can see `ReceiptStatusConsistency` as the first step gets piece record from the storage and compares its status with the one specified under `receivingStatusBeforeUpdate` key. It looks to me that the statuses will be always the same and the rest of the `ReceiptStatusConsistency` logic will never be done...

javalkar-varun: I think you are correct. If we check for inconsistency before sending a message then it should work for eg. this will be the flow:
  - get a piece by id
  - check if receivingStatus is not consistent with storage before sending message to event bus. if yes only do below
  - send message
  - rest of the steps remain the same

Let me know your thoughts?

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -88,6 +92,14 @@ private HelperUtils() {
 
   }
 
+  public static Map<String, String> getOkapiHeaders(Message<JsonObject> message) {
```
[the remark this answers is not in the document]
javalkar-varun: Done

src/main/java/org/folio/rest/impl/AbstractHelper.java
```diff
@@ -95,6 +106,54 @@ public void closeHttpClient() {
     return processingErrors.getErrors();
   }
 
+  protected boolean isCheckin(PoLine poLine) {
+    return defaultIfNull(poLine.getCheckinItems(), false);
```
piotr-kalashuk: The [schema defines default value](https://github.com/folio-org/acq-models/blob/master/mod-orders-storage/schemas/po_line.json#L18) so I think this method is not needed.

src/main/java/org/folio/rest/impl/AbstractHelper.java
```diff
@@ -95,6 +106,54 @@ public void closeHttpClient() {
...
+  }
+
+  public CompletableFuture<String> updatePoLineReceiptStatus(PoLine poLine, ReceiptStatus status, HttpClientInterface httpClient,
```
piotr-kalashuk: This method can be moved to helper utilities class

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
piotr-kalashuk: I believe this part should be covered as well

src/main/java/org/folio/rest/impl/AbstractHelper.java
javalkar-varun: Done

src/main/java/org/folio/rest/impl/AbstractHelper.java
javalkar-varun: Done, few tests had to be refactored.

src/main/java/org/folio/rest/impl/CheckinReceivePiecesHelper.java
```diff
@@ -542,7 +540,7 @@ private void updateOrderStatus(List<PoLine> poLines) {
   private CompletableFuture<ReceiptStatus> calculatePoLineReceiptStatus(int expectedPiecesQuantity, PoLine poLine,
       List<Piece> pieces) {
     // Fully Received:If receiving and there is no expected piece remaining
-    if (!isCheckin(poLine) && expectedPiecesQuantity == 0) {
+    if (expectedPiecesQuantity == 0) {
```
piotr-kalashuk: Instead of just removing method call, it should be replaced by `!poLine.getCheckinItems()`. Current version introduces regression...

javalkar-varun: Done

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
```diff
@@ -0,0 +1,165 @@
...
+          logger.error("The expected receipt status for PO Line '{}' cannot be calculated", e, poLine.getId());
+          return null;
+        });
+    }
+  }
+
+  private CompletableFuture<ReceiptStatus> calculatePoLineReceiptStatus(int expectedPiecesQuantity,
+      List<org.folio.rest.acq.model.Piece> pieces) {
+    
+    if (expectedPiecesQuantity == 0) {
```
javalkar-varun: @piotr-kalashuk - I think the code will never reach here because if `pieces` are empty then future will complete and returned on line 111(`pieces.isEmpty()`) above. Is there any scenario when `expectedPiecesQuantity` be 0 and pieces are not empty?

src/test/java/org/folio/rest/impl/MockServer.java
```diff
@@ -953,6 +956,42 @@ private void handleGetGenericSubObj(RoutingContext ctx, String subObj) {
...
   }
 
+  private void handleGetPieceById(RoutingContext ctx) {
+
+    logger.info("handleGetPiecesById got: " + ctx.request()
+      .path());
+    String pieceId = ctx.request().getParam(ID);
+    JsonObject body;
+    try {
+      if (NON_EXIST_VENDOR_ID.equals(pieceId)) {
```
piotr-kalashuk: I believe first 2 `if` branches should be removed because they are copied from vendors and looks like are not relevant to pieces...

src/test/java/org/folio/rest/impl/MockServer.java
```diff
@@ -953,6 +956,42 @@ private void handleGetGenericSubObj(RoutingContext ctx, String subObj) {
...
+      } else if (PIECE_POLINE_CONSISTENCY_404_POLINE_NOT_FOUND_ID.equals(pieceId)) {
+        body = new JsonObject(ApiTestBase
+          .getMockData(PIECE_RECORDS_MOCK_DATA_PATH + "pieceRecord-poline-not-exists-5b454292-6aaa-474f-9510-b59a564e0c8d.json"));
+        if (body != null) {
+          serverResponse(ctx, HttpStatus.HTTP_OK.toInt(), APPLICATION_JSON, body.encodePrettily());
+        }
+      } else if (PIECE_POLINE_CONSISTENT_RECEIPT_STATUS_ID.equals(pieceId)) {
+        body = new JsonObject(ApiTestBase.getMockData(
+            PIECE_RECORDS_MOCK_DATA_PATH + "pieceRecord-received-consistent-receipt-status-5b454292-6aaa-474f-9510-b59a564e0c8d2.json"));
+        if (body != null) {
```
piotr-kalashuk: The `body` can never be `null` because of `new JsonObject(<file_content>)`

src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
```diff
@@ -228,7 +223,7 @@ public void testPostCheckinRevertPhysicalResource() {
       PoLine poLine = pol.mapTo(PoLine.class);
       assertThat(poLine.getCheckinItems(), is(true));
       assertThat(poLine.getReceiptStatus(), is(PoLine.ReceiptStatus.AWAITING_RECEIPT));
-      assertThat(poLine.getReceiptDate(), is(nullValue()));
```
piotr-kalashuk: This check should remain as is once comment for `updatePoLineReceiptStatus` method from `HelperUtils.java` is addressed

src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
```diff
@@ -384,7 +379,7 @@ public void testPostReceivingElectronicPartially() {
     polUpdates.forEach(pol -> {
       PoLine poLine = pol.mapTo(PoLine.class);
       assertThat(poLine.getReceiptStatus(), is(PoLine.ReceiptStatus.PARTIALLY_RECEIVED));
-      assertThat(poLine.getReceiptDate(), is(nullValue()));
```
piotr-kalashuk: This check should remain as is once comment for `updatePoLineReceiptStatus` method from `HelperUtils.java` is addressed

src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
```diff
@@ -432,7 +427,7 @@ public void testPostReceivingPhysicalWithErrors() throws IOException {
...
```
piotr-kalashuk: This check should remain as is once comment for `updatePoLineReceiptStatus` method from `HelperUtils.java` is addressed

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
pautau-s: I see that the poLines with receivingStatus are still updated during "Receiving/Check-in" instead of sending event. Are these changes going to be added to this PR?

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -1053,6 +1068,33 @@ public static PoLine convertToPoLine(CompositePoLine compPoLine) {
...
+      Context ctx, Map<String, String> okapiHeaders, Logger logger) {
+    
+    if (status == null || poLine.getReceiptStatus() == status) {
+      return completedFuture(null);
+    }
+
+    // Update receipt date and receipt status
+    if (status == FULLY_RECEIVED) {
+      poLine.setReceiptDate(new Date());
+    } else if (poLine.getReceiptStatus()
```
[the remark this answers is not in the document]
piotr-kalashuk: Sorry, I made a mistake above.

Based on [original condition](https://github.com/folio-org/mod-orders/pull/176/files#diff-44c71f622dac45bfb345fce9654c41c6L582) the check should be 
```java
} else if (poLine.getCheckinItems() && ...
```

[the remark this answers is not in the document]
javalkar-varun: Done

src/test/java/org/folio/rest/impl/MockServer.java
javalkar-varun: Sure, removed it.

src/test/java/org/folio/rest/impl/MockServer.java
javalkar-varun: Removed null check

src/main/java/org/folio/orders/events/handlers/ReceiptStatusConsistency.java
javalkar-varun: No, there will a separate PR for that.

PR #180, opened 2019-06-28
MODORDERS-266 CRUD for acquisitions unit assignments

RamanAsadchy:
[MODORDERS-266](https://issues.folio.org/browse/MODORDERS-266)

## Purpose

Implement CRUD for acquisitions unit assignments
## Pre-Merge Checklist:

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -295,7 +295,7 @@
       ]
     },
     {
-      "id": "acquisitions-unit-assignments",
+      "id": "orders.acquisitions-unit-assignments",
       "version": "1.0",
       "handlers": [
         {
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -295,6 +295,42 @@
       ]
     },
     {
+      "id": "acquisitions-unit-assignments",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/acquisitions-unit-assignments",
+          "permissionsRequired": ["orders.acquisitions-units-assignments.assignments.collection.get"],
+          "modulePermissions": ["orders-storage.acquisitions-unit-assignments.collection.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/orders/acquisitions-unit-assignments",
+          "permissionsRequired": ["orders.acquisitions-units-assignments.assignments.item.post"],
+          "modulePermissions": ["orders-storage.acquisitions-unit-assignments.item.post"]
+        },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/acquisitions-unit-assignments/{id}",
+          "permissionsRequired": ["orders.acquisitions-units-assignments.assignments.item.get"],
+          "modulePermissions": ["orders-storage.acquisitions-unit-assignments.item.get"]
+        },
+        {
+          "methods": ["PUT"],
+          "pathPattern": "/orders/acquisitions-unit-assignments/{id}",
+          "permissionsRequired": ["orders.acquisitions-units-assignments.assignments.item.put"],
+          "modulePermissions": ["orders-storage.acquisitions-unit-assignments.item.put"]
+        },
+        {
+          "methods": ["DELETE"],
+          "pathPattern": "/orders/acquisitions-unit-assignments/{id}",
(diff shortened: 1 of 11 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -294,6 +294,42 @@
         }
       ]
     },
+    {
+      "id": "acquisitions-unit-assignments",
```
piotr-kalashuk: The same interface name is going to be for invoicing and finance modules. According to [Okapi Reference - Multiple interfaces](https://github.com/folio-org/okapi/blob/master/doc/guide.md#multiple-interfaces):
> Normally, Okapi proxy allows exactly one module at once to provide a given interface. By using `interfaceType` `multiple` in the `provides` section, Okapi allows any number of modules to implement the same interface. The consequence, however, is that the user of the interface must choose which module to call by specifying HTTP header `X-Okapi-Module-Id`

If we would like to implement the same interface for acq modules, `interfaceType": "multiple"` should be specified and UI (or any other client) has to send `X-Okapi-Module-Id` (value should include module name with version e.g. `mod-orders-6.0.0`). Also in this case I believe the endpoint can be renamed to `/acquisitions-unit-assignments` and all the modules will provide the same endpoints.
Alternative option is to rename interface name to something like `orders.acquisitions-unit-assignments`

craigmcnally: `orders.acquisitions-unit-assignments` please.... likewise, we'll have `invoices.acquisitions-unit-assignemnts`, etc.

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -386,4 +389,85 @@ public void deleteOrdersPiecesById(String pieceId, String lang, Map<String, Stri
...
+    AcquisitionsUnitAssignmentsHelper helper = new AcquisitionsUnitAssignmentsHelper(okapiHeaders, vertxContext, lang);
+
+    helper.createAcquisitionsUnitAssignment(entity)
+      .thenAccept(unit -> {
+        if (logger.isInfoEnabled()) {
+          logger.info("Successfully created new acquisitions unit: " + JsonObject.mapFrom(unit).encodePrettily());
+        }
+
+        asyncResultHandler.handle(succeededFuture(helper.buildResponseWithLocation(String.format(ACQUISITIONS_UNIT_ASSIGNMENTS_LOCATION_PREFIX, unit.getId()), unit)));
+        helper.closeHttpClient();
```
khandramai: public Response buildResponseWithLocation(String endpoint, Object body) already execute http-client closing. No need to explicitly execute closeHttpClient().

src/main/java/org/folio/rest/impl/AcquisitionsUnitAssignmentsHelper.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final String GET_UNIT_ASSIGNMENTS_BY_QUERY = resourcesPath(ACQUISITIONS_UNIT_ASSIGNMENTS) + "?offset=%s&limit=%s%s&lang=%s";
+
+  AcquisitionsUnitAssignmentsHelper(Map<String, String> okapiHeaders, Context ctx, String lang) {
+    super(okapiHeaders, ctx, lang);
+  }
+
+  CompletableFuture<AcquisitionsUnitAssignmentCollection> getAcquisitionsUnitAssignments(String query, int offset, int limit) {
+    CompletableFuture<AcquisitionsUnitAssignmentCollection> future = new VertxCompletableFuture<>(ctx);
+
+    try {
```
khandramai: [Optional]:
From my point of view this try-catch block is redundant because the only one source of exception here is HelperUtils.encodeQuery(String query, Logger logger) method. And this exception may be thrown only as a result of UnsupportedEncodingException due to unsupported encoding charset in the URLEncoder.encode(String s, String enc) method. But we hardcoded absolutely supported standard encoding UTF-8 so I don’t see the possibility of such exception.

piotr-kalashuk:
In general changes look good but please add more unit tests to increase [code coverage](https://sonarcloud.io/dashboard?id=org.folio%3Amod-orders&branch=MODORDERS-266_CRUD_for_acquisitions-unit-assignments&resolved=false)

src/main/java/org/folio/rest/impl/OrdersImpl.java
RamanAsadchy: Good point. Thank you. I already fixed this.

PR #181, opened 2019-07-02
MODORDERS-265 - Restrict fields updated once order is open

KVupp:
## Purpose
https://issues.folio.org/browse/MODORDERS-265- Restrict fields from changing once the Order status has changed from PENDING

## Approach
1. Throw error if the restricted fields are changed
2. Handling, sub object cases if added/deleted/modified

## Pre-Merge Checklist:


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -1097,4 +1097,88 @@ public static PoLine convertToPoLine(CompositePoLine compPoLine) {
...
+
+    return completedFuture(compPOFromStorage);
+
+  }
+
+  public static Set<String> findChangedPoLineProtectedFields(CompositePoLine compPOLine, CompositePoLine compPOLineFromStorage) {
+    Set<String> fields = new HashSet<>();
+    for (String field : POLineProtectedFields.getFieldNames()) {
+      try {
+        if (field.contains(".")) {
```
pautau-s: I'd like to suggest some changes, but it's up to you.
Instead of listing all the options with subobjects and then looking for and splitting them by a dot, I would add a list of "subobjects" field to POLineProtectedFields. This would allow you to get all the subobjects at once, like getSubobjects(), and if the list is not empty then check them in the loop.
I think it would simplify the code a bit.

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -1097,4 +1097,88 @@ public static PoLine convertToPoLine(CompositePoLine compPoLine) {
...
+      throws IllegalAccessException {
+    return !isFieldNull(compPOLine, splits) && isFieldNull(compPOLineFromStorage, splits);
+  }
+
+  private static boolean isFieldNull(CompositePoLine compPoLine, String splits) throws IllegalAccessException {
+    return null == readDeclaredField(compPoLine, splits, true);
+  }
+
+  private static boolean isFieldChanged(Object newObject, Object existedObject, String field) throws IllegalAccessException {
+    return !EqualsBuilder.reflectionEquals(FieldUtils.readDeclaredField(newObject, field, true),
```
pautau-s: `IllegalArgumentException` can be thrown if `field` is not present in the `newObject` or `existedObject`.

pautau-s: I remember clearly that I caught such an exception when I was working on API tests for mod-invoice. I'm not sure how this is possible, because when RMB deserialize json into a POJO it initializes missing fields with null or empty list. 
You can ignore the previous comment because I couldn't reproduce it.

src/main/java/org/folio/orders/utils/HelperUtils.java
piotr-kalashuk: As an alternative option instead of reflection, `JsonObject` can be used to get value from by path. RMB provides [JsonPathParser](https://github.com/folio-org/raml-module-builder/blob/v26.0.0/domain-models-runtime/src/main/java/org/folio/rest/tools/parser/JsonPathParser.java) to simplify work with nested properties.

src/main/java/org/folio/orders/utils/POProtectedFields.java
```diff
@@ -0,0 +1,33 @@
+package org.folio.orders.utils;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public enum POProtectedFields {
+  PO_NUMBER("poNumber"),
+  CLOSE_REASON("closeReason"),
+  MANUAL_PO("manualPo"),
```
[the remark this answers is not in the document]
KVupp: date Ordered is already a read-only field. 
Added template

src/main/java/org/folio/orders/utils/POLineProtectedFields.java
```diff
@@ -0,0 +1,57 @@
...
+  ERESOURCE_MATERIAL_TYPE("eresource.materialType"),
+  ERESOURCE_USER_LIMIT("eresource.userLimit"),
+  FUND_DISTRIBUTION("fundDistribution"),
+  LOCATION("locations") ,
+  ORER_FORMAT("orderFormat"),
+  PHYSICAL_VOLUMES("physical.volumes"),
+  PHYSICAL_MATERIAL_TYPE("physical.materialType"),
+  PUBLICATION_DATE("publicationDate"),
+  PUBLISHER("publisher"),
+  PO_LINE_NUMBER("poLineNumber"),
```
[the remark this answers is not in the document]
KVupp: removed it

src/main/java/org/folio/orders/utils/HelperUtils.java
KVupp: Good find, Using JsonPathParser for parsing the json

src/main/java/org/folio/orders/utils/POLineProtectedFields.java
```diff
@@ -0,0 +1,57 @@
+package org.folio.orders.utils;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public enum POLineProtectedFields {
+
+  CHECKIN_ITEMS("checkinItems"),
+  COST_LIST_UNIT_PRICE("cost.listUnitPrice"),
```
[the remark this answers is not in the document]
KVupp: correct, that is the reason Cost was not added directly but each field had to be specified 
added additional Cost too

src/main/java/org/folio/orders/utils/POLineProtectedFields.java
```diff
@@ -0,0 +1,57 @@
...
+  PUBLICATION_DATE("publicationDate"),
+  PUBLISHER("publisher"),
+  PO_LINE_NUMBER("poLineNumber"),
+  RUSH("rush"),
+  SOURCE("source"),
+  SELECTOR("selector"),
+  TITLE("title"),
+  VENDORDETAIL_INSTRUCTION_TO_VENDOR("vendorDetail.instructions"),
+  VENDORDETAIL_VENDOR_ACCOUNT("vendorDetail.vendorAccount");
+
```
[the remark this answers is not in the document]
KVupp: added

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -150,6 +131,27 @@
       );
   }
 
+
+  public CompletableFuture<JsonObject> validateIfPOProtectedFieldsChanged(CompositePurchaseOrder compPO,
+      JsonObject compPOFromStorage) {
+    if (!compPOFromStorage.getString(WORKFLOW_STATUS)
```
pautau-s: Just a reason to think...
If converting an order into a composite order before `if`, the conditional expression becomes more readable, but I have doubts whether it is correct to convert and validate the order in the one method.

src/test/java/org/folio/rest/impl/ApiTestBase.java
```diff
@@ -362,4 +365,40 @@ void verifyReceiptStatusUpdateEvent(int msgQty) {
       assertThat(message.body().getString(HelperUtils.LANG), not(isEmptyOrNullString()));
     }
   }
+
+  void checkPreventProtectedFieldsModificationRule(String path, JsonObject compPO, Map<String, Object> updatedFields)
+      throws IllegalAccessException {
```
pautau-s: `IllegalAccessException` is never thrown in this method

src/test/java/org/folio/rest/impl/ApiTestBase.java
```diff
@@ -362,4 +365,40 @@ void verifyReceiptStatusUpdateEvent(int msgQty) {
...
+
+    Error error = errors.getErrors()
+      .get(0);
+    assertThat(error.getCode(), equalTo(PROHIBITED_FIELD_CHANGING.getCode()));
+
+    Object[] failedFieldNames = getModifiedProtectedFields(error);
+    Object[] expected = updatedFields.keySet()
+      .stream()
+      .map(s -> {
+        return s.replace(COMPOSITE_PO_LINES_PREFIX, "");
```
pautau-s: Can be changed to an expression lambda

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -1107,4 +1105,24 @@ public static PoLine convertToPoLine(CompositePoLine compPoLine) {
             return null;
           });
   }
+
+  public static JsonObject verifyProtectedFieldsChanged(List<String> protectedFields, JsonObject objectFromStorage,
+      JsonObject requestObject) {
+    Set<String> fields = new HashSet<>();
+    JsonPathParser oldObject = new JsonPathParser(JsonObject.mapFrom(objectFromStorage));
```
pautau-s: Is it really necessary to map from jsonObject to jsonObject?

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -130,13 +131,24 @@
...
+      .equals(CompositePurchaseOrder.WorkflowStatus.PENDING.value())) {
+      verifyProtectedFieldsChanged(POProtectedFields.getFieldNames(), compPOFromStorage, JsonObject.mapFrom(compPO));
+    }
+    return completedFuture(compPOFromStorage);
+  }
+
+  private JsonObject findCorrespondingCompositePoLine(CompositePoLine poLine, JsonArray poLinesFromStorage) {
+    return poLinesFromStorage.copy()
```
piotr-kalashuk: The `copy` operation might not be needed. In case there is a reason why we need to work with copy of original object, I would advise to add as the last step (right before `.orElse`) the following: `.map(JsonObject::copy)`

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -150,6 +131,27 @@
...
+      .stream()
+      .filter(line -> ((JsonObject) line).getString(ID)
+        .equals(poLine.getId()))
+      .findFirst()
+      .map(line -> (JsonObject) line)
+      .orElse(JsonObject.mapFrom(poLine));
```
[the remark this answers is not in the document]
KVupp: correct, the check is done prior, removing the return

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
KVupp: I didn't want to convert it to an object because JsonObject is needed all the way for verification, it will just be an over head to convert back and forth, so leaving it as is

KVupp:
Made necessary changes except where I detailed a reason

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
pautau-s: You still can use jsonObject from parameter for verification

PR #183, opened 2019-07-05
MODORDERS-251 - Restrict creation of PO, POL, Piece records based upon acquisitions unit

khandramai:
[MODORDERS-251](https://issues.folio.org/browse/MODORDERS-251) Restrict creation of PO, POL, Piece records based upon acquisitions unit 

## Purpose
Restrict Creation of orders, poLines and pieces based upon the acquisitions unit memberships of the user and the acquisitions unit being assigned to the record.

## Approach
[ProtectionHelper](https://github.com/folio-org/mod-orders/pull/183/files#diff-60aec540e07e6b2deb4e6e8123c8d5ff) introduced. This helper retrieves units (header X-OKAPI-USER-ID -> unit ids from memberships API -> units -> merging protection status for CRUD operations)

[Option #3](https://wiki.folio.org/display/FOLIJET/Acquisition+Units#AcquisitionUnits-OrderofOperations/Chicken&EggProblem) is selected for managing `acquisitionUnitsAssignments`

#### TODOS and Open Questions

1. Restrictions on deletion and creation of poLines during the order update will be handled in the scope of [MODORDERS-254](https://issues.folio.org/browse/MODORDERS-254)/[MODORDERS-255](https://issues.folio.org/browse/MODORDERS-255) PR.

## Pre-Merge Checklist:


src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * This method determines status of operation restriction based on unit IDs from {@link CompositePurchaseOrder}.
+   * @param unitIds list of unit IDs.
+   *
+   * @return true if operation is restricted, otherwise - false.
+   */
+  public CompletableFuture<Boolean> isOrderCreationRestricted(List<String> unitIds) {
+    if (unitIds != null && !unitIds.isEmpty()) {
```
pautau-s: @craigmcnally should we allow to create an order without unit-assignments?

src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * This method determines status of operation restriction based on ID of {@link CompositePurchaseOrder}.
+   * @param recordId corresponding record ID.
+   *
+   * @return true if operation is restricted, otherwise - false.
+   */
+  public CompletableFuture<Boolean> isOperationRestricted(String recordId) {
+    String userId;
+    if(okapiHeaders != null && (userId = okapiHeaders.get(OKAPI_USERID_HEADER)) != null) {
```
pautau-s: In my opinion, it would be reasonable to do this check before calling any API. For example, in `OrdersImpl#postOrdersCompositeOrders`

src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * This method determines status of operation restriction based on units analysis and user ID.
+   * @param unitIds list of unit's IDs.
+   * @param userId user's ID.
+   *
+   * @return true if operation is restricted, otherwise - false.
+   */
+  private CompletableFuture<Boolean> isOperationRestricted(String userId, List<String> unitIds) {
```
pautau-s: The `userId` parameter is unnecessary because we can take the user ID from the `okapiHeaders` at any time.

src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,142 @@
...
+    if(!unitIds.isEmpty()) {
+      CompletableFuture<Boolean> future = new CompletableFuture<>();
```
pautau-s: Instead of completing this future inside the `getUnitsByIds()...` pipeline, you can return the result of `getUnitsByIds()...`.  That will make the method more compact.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -90,10 +98,18 @@
...
    */
   public CompletableFuture<CompositePurchaseOrder> createPurchaseOrder(CompositePurchaseOrder compPO) {
-    return setPoNumberIfMissing(compPO)
-      .thenCompose(v -> poNumberHelper.checkPONumberUnique(compPO.getPoNumber()))
-      .thenCompose(v -> createPOandPOLines(compPO))
-      .thenApply(this::populateOrderSummary);
+    return protectionHelper.isOperationRestricted(compPO.getAcqUnitIds())
+      .thenCompose(isProtected -> {
+        if(isProtected) {
+          throw new HttpException(HttpStatus.HTTP_FORBIDDEN.toInt(), USER_HAS_NOT_PERMISSIONS);
```
pautau-s: The result of the `isOperationRestricted` method is just used to throw in an exception in case the operation is restricted. It may be better to throw an exception inside the method when the operation is restricted otherwise just to return `CompleatableFuture<Void>`

src/main/java/org/folio/rest/impl/ProtectionHelper.java
khandramai: Moreover, X-Okapi-User-Id validation seems unnecessary due to https://github.com/folio-org/okapi/blob/master/doc/guide.md#x-okapi-headers.

README.md
```diff
@@ -11,6 +11,9 @@ This is the Orders business logic module.
 
 ## Additional information
 
+IMPORTANT NOTE! It is required that user should have permissions `orders-storage.acquisitions-unit-assignments.item.post` to create an purchase order
```
KVupp: The user needs to have the extra permission, **only** if an **acquisition unit** has to be assigned to the Order, it would be nice to specifically mention that here

src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -47,7 +47,10 @@
   MISSING_CONTRIBUTOR_NAME_TYPE("missingContributorNameType", "The specified contributor-name-types were not found in inventory-storage"),
   MISSING_LOAN_TYPE("missingLoanType", "Loan-type is a required field for creating an Item in Inventory"),
   ENCUMBRANCE_CREATION_FAILURE("encumbranceCreationFailure", "One or more encumbrance record(s) failed to be created"),
-  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified");
+  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified"),
+  UNKNOWN_USER("restrictedDueToUnknownUser", "User Id is not presented"),
+  ORDER_UNITS_NOT_FOUND("orderUnitsNotFound", "Order doesn't contain unit ids or units not found"),
```
KVupp: This error message is used only in https://github.com/folio-org/mod-orders/blob/d4eb9bb800a3885d80316013e2726a88ce44e00c/src/main/java/org/folio/rest/impl/ProtectionHelper.java#L86
and is thrown only if Acq Units are not found, so "Order doesn't contain unit ids" can be removed

craigmcnally:
Good feedback @pautau-s, @KVupp

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -90,10 +98,18 @@
...
+        } else {
+          return setPoNumberIfMissing(compPO)
+            .thenCompose(v -> poNumberHelper.checkPONumberUnique(compPO.getPoNumber()))
+            .thenCompose(v -> createPOandPOLines(compPO))
+            .thenCompose(v -> createUnitAssignments(compPO))
```
KVupp: The assignments are created after the Order,so if the User doesn't have the permission "orders-storage.acquisitions-unit-assignments.item.post", the order is still created, but without assignments. We should either 
1. clean up the order
2. Check the assignment permission before creating the order

craigmcnally: it should be 2) per https://wiki.folio.org/display/FOLIJET/Acquisitions+Units#AcquisitionsUnits-RestrictingCreation/ModificationofRecords

ramls/order.raml
```diff
@@ -57,7 +57,7 @@ resourceTypes:
     get:
       is: [pageable, searchable: {description: "using CQL (indexes for purchase orders)", example: "workflow_status==\"Pending\""} ]
     post:
-      description: Post a purchase order (PO) and a number of PO lines; record fund transactions corresponding to the order
+      description: Post a purchase order (PO) and a number of PO lines; record fund transactions corresponding to the order. IMPORTANT! It is required that user should have permissions orders-storage.acquisitions-unit-assignments.item.post to create an order during the call to the API. Otherwise, error 403 will be returned.
```
KVupp: please update that it's needed only if acquisition units are being assigned

src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * This method returns list of units ids associated with User.
+   * @param userId id of User.
+   *
+   * @return list of unit ids associated with user.
+   */
+  private CompletableFuture<Integer> getUnitIdsAssignedToUserAndOrder(String userId, List<String> unitIdsAssignedToOrder) {
+    String query = String.format("userId==%s AND %s", userId, convertIdsToCqlQuery(unitIdsAssignedToOrder, "acquisitionsUnitId"));
```
KVupp: This method is not checking if the user is part of all the acquisition units which have protectCreate=true:For example:
If MAIN and LAW both have protectCreate = true
User Bob is part of Only MAIN.
Order has acquistions Units : MAIN, LAW

This method returns the id of just MAIN, and an Order will be created with unit LAW even though Bob is not part of it and it has protect create

src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -47,7 +47,9 @@
   MISSING_CONTRIBUTOR_NAME_TYPE("missingContributorNameType", "The specified contributor-name-types were not found in inventory-storage"),
   MISSING_LOAN_TYPE("missingLoanType", "Loan-type is a required field for creating an Item in Inventory"),
   ENCUMBRANCE_CREATION_FAILURE("encumbranceCreationFailure", "One or more encumbrance record(s) failed to be created"),
-  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified");
+  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified"),
+  ORDER_UNITS_NOT_FOUND("orderUnitsNotFound", "Order doesn't contain unit ids or units not found"),
+  USER_HAS_NOT_PERMISSIONS("userHasNoPermission", "User hasn't permissions - operation is restricted");
```
KVupp: Looks like a typo: Must be USER_HAS_**NO**_PERMISSIONS
Also the description should be, User has no permissions

src/main/java/org/folio/rest/impl/PiecesHelper.java
```diff
@@ -1,5 +1,6 @@
 package org.folio.rest.impl;
 
+import static org.folio.orders.utils.ErrorCodes.USER_HAS_NOT_PERMISSIONS;
```
KVupp: This import is not being used anywhere

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -321,6 +331,18 @@ private int calculateTotalItemsQuantity(List<CompositePoLine> poLines) {
     return validateOrder(compPO);
   }
 
+  private CompletableFuture<CompositePurchaseOrder> createUnitAssignments(CompositePurchaseOrder comPO) {
+    CompletableFuture<CompositePurchaseOrder> future = new CompletableFuture<>();
+    CompletableFuture.allOf(comPO.getAcqUnitIds().stream().map(id -> acquisitionsUnitAssignmentsHelper
+      .createAcquisitionsUnitAssignment(new AcquisitionsUnitAssignment())).toArray(CompletableFuture[]::new))
```
KVupp: The acquisition Unit assignment will not be created for the record here, just a new Object is being sent with no values

src/test/java/org/folio/rest/impl/ProtectionHelperTest.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final Header[] ALLOWED_CREATION_HEADERS = {X_OKAPI_USER_WITH_UNITS_ASSIGNED_TO_ORDER};
+
+  private static final String ACQUISITIONS_UNITS = "acquisitionsUnits";
+  private static final String ACQUISITIONS_MEMBERSHIPS = "acquisitionsMemberships";
+  private static final String ACQUISITIONS_UNIT_ASSIGNMENTS = "acquisitionsUnitAssignments";
+
+
+  @Test
+  public void testAssignmentCreation() {
+    logger.info("=== Test request without user id header - expecting no calls to Units, Memberships, Assignments API ===");
```
KVupp: Please change the logger message to reflect the test

src/test/java/org/folio/rest/impl/ProtectedEntities.java
```diff
@@ -0,0 +1,125 @@
...
+        piece.setPoLineId(PO_LINE_FOR_ORDER_WITHOUT_UNITS_ID);
+        return JsonObject.mapFrom(piece).encode();
+      case ORDER_LINES:
+        CompositePoLine poLine = getMinimalContentCompositePoLine();
+        poLine.setId(PO_LINE_FOR_ORDER_WITHOUT_UNITS_ID);
+        poLine.setPurchaseOrderId(ORDER_WITHOUT_UNITS_ID);
+        return JsonObject.mapFrom(poLine).encode();
+      case ORDERS:
+        CompositePurchaseOrder compPo = getMinimalContentCompositePurchaseOrder();
+        compPo.setAcqUnitIds(Arrays.asList(NON_EXISTED_UNIT_ID, "0f2bb7a2-728f-4e07-9268-082577a7bedb"));
```
KVupp: Method name says "WithoutUnits" but units are being assigned here?

src/test/java/org/folio/rest/impl/ProtectionHelperTest.java
```diff
@@ -0,0 +1,155 @@
...
+  @Test
+  public void testOperationWithoutUnits() {
+    logger.info("=== Test corresponding order hasn't units - expecting of call only to Assignments API ===");
+
+    for(ProtectedOperations operation : ProtectedOperations.values()) {
+      for(ProtectedEntities holder : Arrays.asList(PIECES, ORDER_LINES)) {
+        operation.process(holder.getEndpoint(), holder.getSampleForFlowWithoutUnits(), prepareHeaders(EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10, X_OKAPI_USER_ID), APPLICATION_JSON, HttpStatus.HTTP_CREATED.toInt());
+        validateNumberOfRequests(0, 0, 1);
+      }
+      // Composite PO already contains acquisition unit IDs
```
KVupp: Method name/description doesnt match the test here, please  change

src/test/java/org/folio/rest/impl/ReceivingHistoryApiTest.java
```diff
@@ -46,7 +46,7 @@ public void testGetReceivingHistory() {
 
     List<String> queryParams = getQueryParams(RECEIVING_HISTORY);
     assertThat(queryParams, hasSize(1));
-    assertThat(queryParams.get(0), equalTo(NO_ACQ_UNIT_ASSIGNED_CQL));
+//    assertThat(queryParams.get(0), equalTo(NO_ACQ_UNIT_ASSIGNED_CQL));
```
KVupp: is this code commented purposefully?

src/test/java/org/folio/rest/impl/PurchaseOrderLinesApiTest.java
```diff
@@ -660,7 +666,7 @@ public void testGetOrderPoLinesByEmptyQuery() {
 
     List<String> queryParams = getQueryParams(PO_LINES);
     assertThat(queryParams, hasSize(1));
-    assertThat(queryParams.get(0), equalTo(NO_ACQ_UNIT_ASSIGNED_CQL));
+//    assertThat(queryParams.get(0), equalTo(NO_ACQ_UNIT_ASSIGNED_CQL));
```
KVupp: is this commented on purpose?

src/test/java/org/folio/rest/impl/ReceivingHistoryApiTest.java
khandramai: Definitely yes - this should be uncommented after assignment/membership/unit mocks number will be stable. I'm going to do this in scope of story for protection of GET orders/lines/pieces as this test introduced there.

src/test/java/org/folio/rest/impl/PurchaseOrderLinesApiTest.java
khandramai: The same as above.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
khandramai: Done.

src/main/java/org/folio/rest/impl/PiecesHelper.java
khandramai: Fixed.

src/main/java/org/folio/orders/utils/ErrorCodes.java
khandramai: Done

README.md
khandramai: Done.

ramls/order.raml
khandramai: Done.

src/main/java/org/folio/orders/utils/ErrorCodes.java
khandramai: Fixed.

src/test/java/org/folio/rest/impl/ProtectedEntities.java
khandramai: Method name fixed.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
khandramai: Assignments creation moved before order creation. So if permission is not presented flow should break. I updated unit test to cover this flow - if POST assignment was made then POST order was made too, otherwise - both requests are absent.

src/main/java/org/folio/rest/impl/ProtectionHelper.java
khandramai: @KVupp, according to the algorithm: https://wiki.folio.org/display/FOLIJET/Acquisitions+Units?preview=/25724746/25725391/Orders_Create.jpg#AcquisitionsUnits-Users. If order creation is protected by all order’s units we just need to check if User is member of order’s units.

What does this means?

Initially I supposed that we just need to check if User belongs to even one Unit from Order’s units. This case we don’t need to validate User’s protection status based on its units (included belonging and not belonging order’s units).

As per your comments this is not correct. And we need to check User’s protection status based on all assigned to it Units (even if part of Units is not belonging Order’s units). After that we can make decision if current operation is restricted or no. Am I correct?

KVupp: My understanding is User must not be able to create an order in acq unit that he doesn't belong to. @craigmcnally please confirm!

craigmcnally: if protectCreate == true, the user needs to belong to at least one of the specified acquisitions units.  I don't think we need to restrict them to being a member of all units the order is being created with.

This is consistent with the behavior of edit/update

pautau-s: The [Acquisitions Units documentation](https://wiki.folio.org/display/FOLIJET/Acquisitions+Units) under "Conflicts and Priorities" states: "Bob, Ben, and Brenda all belong to at least one of the acquisitions units and can therefore perform any actions on this record."

KVupp: so essentially this comes down to, If I am the user, all I need is just to be part of a single Group with protect create = true, I have then the full power to create a record in any other acq unit that I wish(even though the creation is restricted to certain members).I still think it presents problems to the consortium colleges, however we can handle it when we get there.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
KVupp: This still needs a change because, id will not be present in the incoming POST request, so getId() method used below will not have any value  
https://github.com/folio-org/mod-orders/blob/551e598bbad5494188e911b1344187ee5f440e39/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java#L340

Also present in the unit test logs:
Screen Shot 2019-07-25 at 8 37 01 AM

khandramai: @KVupp,

pautau-s: To make it possible, we have to remove `recordId` foreign key

KVupp: I think it's better to check the user permission before, instead of removing the foreign key constraint.

KVupp: Another scenario to consider is, what if we create the Order, and then assignments API fails for some reason(even thought it has the required permissions), then an order will be created without any acq units, I think in this scenario we should clean up the created order and return an error

pautau-s: > I think it's better to check the user permission before, instead of removing the foreign key constraint.

@KVupp  Look at last comment https://wiki.folio.org/display/FOLIJET/Acquisitions+Units

KVupp: then other thing I can think of is to create a new /validate endpoint to see if the user has assignment permission.

craigmcnally: OK, I'm not sure there are any great options here... I gave it some thought tonight and wrote up the options on the wiki page for comparison (https://wiki.folio.org/display/FOLIJET/Acquisitions+Units#AcquisitionsUnits-OrderofOperationsDuringRecordCreation)

I've added a 4th option not discussed yet.  Please take a look; we can discuss tomorrow.  If I missed anything feel free to add it to the page or comments section.

craigmcnally: Updated:  https://wiki.folio.org/display/FOLIJET/Acquisition+Units?focusedCommentId=25727487#AcquisitionUnits-OrderofOperations/Chicken&EggProblem

README.md
```diff
@@ -11,6 +11,9 @@ This is the Orders business logic module.
 
 ## Additional information
 
+IMPORTANT NOTE! Only in case an acquisition unit has to be assigned to the Order it is required that user should have 
+extra permission `orders-storage.acquisitions-unit-assignments.item.post` to create an purchase order.
```
piotr-kalashuk: I believe this note has to be updated to mention `orders.assignments.item.post` instead of `orders-storage.acquisitions-unit-assignments.item.post`

descriptors/ModuleDescriptor-template.json
```diff
@@ -20,6 +20,7 @@
           "methods": ["POST"],
           "pathPattern": "/orders/composite-orders",
           "permissionsRequired": ["orders.item.post"],
+          "permissionsDesired": ["orders.assignments.item.post"],
```
piotr-kalashuk: The permission name might be confusing because it does not highlight that this is for acq units...

descriptors/ModuleDescriptor-template.json
```diff
@@ -243,7 +249,11 @@
           "pathPattern": "/orders/pieces",
           "permissionsRequired": ["orders.pieces.item.post"],
           "modulePermissions": [
-            "orders-storage.pieces.item.post"
+            "orders-storage.pieces.item.post",
+            "orders-storage.po-lines.collection.get",
+            "acquisitions-units-storage.units.collection.get",
+            "acquisitions-units-storage.memberships.collection.get",
+            "orders-storage.acquisitions-unit-assignments.collection.get"
```
piotr-kalashuk: `orders-storage.acquisitions-unit-assignments.*` are going to be removed in https://github.com/folio-org/mod-orders-storage/pull/146

ramls/order.raml
```diff
@@ -57,7 +54,7 @@ resourceTypes:
     get:
       is: [pageable, searchable: {description: "using CQL (indexes for purchase orders)", example: "workflow_status==\"Pending\""} ]
     post:
-      description: Post a purchase order (PO) and a number of PO lines; record fund transactions corresponding to the order
+      description: Post a purchase order (PO) and a number of PO lines; record fund transactions corresponding to the order. Only in case an acquisition unit has to be assigned to the Order it is required that user should have extra permission orders-storage.acquisitions-unit-assignments.item.post to create an purchase order.
```
piotr-kalashuk: `orders-storage.acquisitions-unit-assignments.item.post` should be replaced by new permission name

README.md
KVupp: please change the permission name to "orders.assignments.item.post"

ramls/order.raml
KVupp: please change the permission name to "orders.assignments.item.post"

descriptors/ModuleDescriptor-template.json
```diff
@@ -136,7 +139,10 @@
             "orders-storage.po-lines.item.post",
             "orders-storage.alerts.item.post",
             "orders-storage.reporting-codes.item.post",
-            "configuration.entries.collection.get"
+            "configuration.entries.collection.get",
+            "acquisitions-units-storage.units.collection.get",
+            "acquisitions-units-storage.memberships.collection.get",
+            "orders-storage.acquisitions-unit-assignments.collection.get"
```
KVupp: we are no longer going to use "orders-storage.acquisitions-unit-assignments.collection.get"

descriptors/ModuleDescriptor-template.json
KVupp: same here, please remove invalid permission

src/main/java/org/folio/rest/impl/OrdersImpl.java
```diff
@@ -80,34 +85,20 @@ public void postOrdersCompositeOrders(String lang, CompositePurchaseOrder compPO
...
+                        .withLocation(String.format(ORDERS_LOCATION_PREFIX, withIds.getId())));
                   asyncResultHandler.handle(succeededFuture(response));
-                })
-                .exceptionally(t -> {
-                  logger.error("Failure happened applying funds to created order");
-                  asyncResultHandler.handle(succeededFuture(helper.buildErrorResponse(t)));
-                  return null;
-                });
-            })
-            .exceptionally(t -> {
```
piotr-kalashuk: I think one `exceptionally` block should still remain to handle any issues which might happen in `createPurchaseOrder`

piotr-kalashuk: I see, it is going to be handled by outer `exceptionally`

descriptors/ModuleDescriptor-template.json
```diff
@@ -243,7 +249,11 @@
           "pathPattern": "/orders/pieces",
           "permissionsRequired": ["orders.pieces.item.post"],
           "modulePermissions": [
-            "orders-storage.pieces.item.post"
+            "orders-storage.pieces.item.post",
+            "orders-storage.po-lines.collection.get",
```
piotr-kalashuk: `orders-storage.po-lines.item.get` and `orders-storage.purchase-orders.item.get` instead of `orders-storage.po-lines.collection.get`

src/main/java/org/folio/rest/impl/ProtectionHelper.java
```diff
@@ -0,0 +1,92 @@
...
+   * This method determines status of operation restriction based on unit IDs from {@link CompositePurchaseOrder}.
+   * @param unitIds list of unit IDs.
+   *
+   * @throws HttpException if user hasn't permissions or units not found
+   */
+  public CompletableFuture<Void> isOperationRestricted(List<String> unitIds, ProtectedOperationType operation) {
+    if (CollectionUtils.isNotEmpty(unitIds)) {
+      return getUnitsByIds(unitIds)
+        .thenCompose(units -> {
+          if (unitIds.size() == units.size()) {
```
piotr-kalashuk: This check potentially might introduce problems. Let's imagine that order has 2 units assigned `A` and `B`. For some reason unit `B` has been deleted but order still references its id. User attempts to remove the unit from the order. This check won't pass and an error will be returned back to client i.e. there is no way to unassign such unit...

src/main/java/org/folio/orders/utils/ErrorCodes.java
```diff
@@ -47,7 +47,10 @@
   MISSING_CONTRIBUTOR_NAME_TYPE("missingContributorNameType", "The specified contributor-name-types were not found in inventory-storage"),
   MISSING_LOAN_TYPE("missingLoanType", "Loan-type is a required field for creating an Item in Inventory"),
   ENCUMBRANCE_CREATION_FAILURE("encumbranceCreationFailure", "One or more encumbrance record(s) failed to be created"),
-  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified");
+  PROHIBITED_FIELD_CHANGING("protectedFieldChanging","Protected fields can't be modified"),
+  ORDER_UNITS_NOT_FOUND("orderUnitsNotFound", "Units not found"),
+  USER_HAS_NO_PERMISSIONS("userHasNoPermission", "User does not have permissions - operation is restricted"),
+  USER_HAS_NO_DESIRED_PERMISSIONS("userHasNoDesiredPermission", "User does not have desired permissions - operation is restricted");
```
KVupp: I think we can make the error message more descriptive, that the permission is related to acquisition units.

README.md
KVupp: Just realized we both posted the same comments at the same time, Going to remove the duplicates

src/main/java/org/folio/rest/impl/ProtectionHelper.java
pautau-s: I don't see an easy way to solve this here, maybe it's worth to put some logic to Acquisitions units API that will check if there are any records referring to this unit before deleting it...

ramls/order.raml
pautau-s: done

README.md
pautau-s: done

descriptors/ModuleDescriptor-template.json
pautau-s: renamed

descriptors/ModuleDescriptor-template.json
pautau-s: removed

ramls/order.raml
pautau-s: updated

descriptors/ModuleDescriptor-template.json
pautau-s: deleted

descriptors/ModuleDescriptor-template.json
pautau-s: done

src/main/java/org/folio/orders/utils/ErrorCodes.java
pautau-s: updated

src/test/java/org/folio/rest/impl/protection/LinesProtectionTest.java
```diff
@@ -0,0 +1,86 @@
...
+public class LinesProtectionTest extends ProtectedEntityTestBase {
+
+  private static final Logger logger = LoggerFactory.getLogger(LinesProtectionTest.class);
+
+  @Test
+  @Parameters({
+    "CREATE"
+  })
+  public void testOperationWithNonExistedUnits(ProtectedOperations operation) {
+    logger.info("=== Test corresponding order doesn't have units - expecting of call only to Assignments API ===");
```
KVupp: please correct the logger messages and the comments in this method

pautau-s: done

src/main/java/org/folio/rest/impl/ProtectionHelper.java
piotr-kalashuk: Looks like this is the only way to resolve potential issue....

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -3,47 +3,90 @@
...
   private final PoNumberHelper poNumberHelper;
   private final PurchaseOrderLineHelper orderLineHelper;
+  private final ProtectionHelper  protectionHelper;
 
   PurchaseOrderHelper(Map<String, String> okapiHeaders, Context ctx, String lang) {
     super(getHttpClient(okapiHeaders), okapiHeaders, ctx, lang);
 
     poNumberHelper = new PoNumberHelper(httpClient, okapiHeaders, ctx, lang);
     orderLineHelper = new PurchaseOrderLineHelper(httpClient, okapiHeaders, ctx, lang);
+    protectionHelper = new ProtectionHelper(okapiHeaders, ctx, lang);
```
piotr-kalashuk: I believe better to add constructor to `ProtectionHelper` which accepts `httpClient` as well

pautau-s: done

descriptors/ModuleDescriptor-template.json
```diff
@@ -545,6 +554,25 @@
         "acquisitions-units.units.item.delete"
       ]
     },
+    {
+      "permissionName": "orders.acquisitions-units-assignments.item.post",
+      "displayName": "Acquisitions unit assignment - create unit assignment",
+      "description": "Create a new acquisitions unit assignments"
```
KVupp: Minor: Descriptions and display name for these permissions can be changed a bit to reflect that the assignment is being added to order

PR #229, opened 2019-10-15
MODORDERS-316 Prevent duplicate ISBNs

pautau-s:
## Purpose
[MODORDERS-316](https://issues.folio.org/browse/MODORDERS-316)  after conversion to 13 digit format ISBN duplicates should be identified and stripped out before being stored/returned to the client.

## Approach
- added [logic](https://issues.folio.org/browse/MODORDERS-316?focusedCommentId=62526&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-62526) to remove duplicate ISBN 
- enabled `jsonschema2pojo.config.includeHashcodeAndEquals` to generate models with `equal` `hashCode` methods

additional work
- handled error case when ISBN is not valid by adding `invalidISBN` error code in response

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/orders/utils/HelperUtils.java
+++ b/src/main/java/org/folio/orders/utils/HelperUtils.java
@@ -1159,4 +1159,8 @@ public class HelperUtils {
       .map(HelperUtils::convertToPoLine)
       .collect(toList());
   }
+
+  public static boolean isProductIdsExist(CompositePoLine compPOL) {
+    return compPOL.getDetails() != null && CollectionUtils.isNotEmpty(compPOL.getDetails().getProductIds());
+  }
 }
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "orders",
-      "version": "7.1",
+      "version": "8.0",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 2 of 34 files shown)


src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -1159,4 +1159,8 @@ public static JsonObject verifyProtectedFieldsChanged(List<String> protectedFiel
       .map(HelperUtils::convertToPoLine)
       .collect(toList());
   }
+
+  public static boolean isProductIdsExist(CompositePoLine compPOL) {
+    return compPOL.getDetails() != null && CollectionUtils.isNotEmpty(compPOL.getDetails().getProductIds());
```
khandramai: It may make sense to foresee the case when ProductIdType doesn't exist in request because this field isn't required in schema?

pautau-s: I think we'll have to make `productIdType` required,since this field is used to add identifiers to the instance and it is required. 
https://github.com/folio-org/mod-orders/blob/cfd04c3cfd4248350b3b6bd851a89b90bd2088ca/src/main/java/org/folio/rest/impl/InventoryHelper.java#L485
https://github.com/folio-org/mod-inventory-storage/blob/9405f40053e24ed79912ef0c583e05bce59fa017/ramls/instance.json#L81-L82

KVupp: I don't think we can make productIdType required because we have scenarios where inventory interaction is not required

pautau-s: Forgot to link a story [MODORDSTOR-115](https://issues.folio.org/browse/MODORDSTOR-115) that I created to handle this. There are some clarifications from the POs in the comments.

PR #267, opened 2020-03-26
MODORDERS-345 Fund distribution totals must add upto 100%

javalkar-varun:
## Purpose

https://issues.folio.org/browse/MODORDERS-345

## Approach

- Before opening an Order verify that Fund distribution totals do not go above 100% or total price
- Add unit tests

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
@@ -471,6 +471,7 @@ public class PurchaseOrderHelper extends AbstractHelper {
     compPO.setWorkflowStatus(OPEN);
     compPO.setDateOrdered(new Date());
     return getOrderWithLines(compPO)
+      .thenApply(this::validateFundDistributionTotal)
       .thenApply(this::validateMaterialTypes)
       .thenCompose(this::fetchNonPackageTitles)
       .thenCompose(lineIdTitles -> {
@@ -529,7 +530,6 @@ public class PurchaseOrderHelper extends AbstractHelper {
       .thenCompose(v -> validateIsbnValues(compPO))
       .thenCompose(v -> validatePoLineLimit(compPO))
       .thenCompose(v -> validateVendor(compPO))
-      .thenCompose(v -> validateFundDistributionTotal(compPO))
       .thenAccept(v -> validateRenewalInfo(compPO))
       .thenApply(v -> getErrors().isEmpty());
   }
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -661,6 +669,57 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
...
+
+            /**
+             * calculate remaining percentage to carry forward, required if there are more fund distributions with amount and percentage
+             * remainingPercentage = remainingAmount/totalEstimatedPrice * 100
+             */
+            MonetaryAmount divideResult = remainingAmount.divide(totalEstimatedPrice);
+            remainingPercent = divideResult.multiply(100).getNumber().doubleValue();
+
+            MonetaryAmount zeroMoney = Money.of(0, currency);
+            if (remainingAmount.isLessThan(zeroMoney)) {
```
pautau-s: I would recommend choosing one of the validation's variants either percentage or amount and convert all distributions to the same type, it will reduce the amount of duplicated code.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -661,6 +669,57 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
...
   }
 
+  private CompletableFuture<Void> validateFundDistributionTotal(CompositePurchaseOrder compPO) {
+    if (compPO.getWorkflowStatus() == WorkflowStatus.OPEN && compPO.getTotalEstimatedPrice() != null) {
+      Double totalEstimatedPrice = compPO.getTotalEstimatedPrice();
+      Double remainingPercent = 100.0;
+
+      for (CompositePoLine cPoLine : compPO.getCompositePoLines()) {
+        String currency = cPoLine.getCost().getCurrency();
+        MonetaryAmount remainingAmount = Money.of(totalEstimatedPrice, currency);
```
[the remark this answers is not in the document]
javalkar-varun: There are multiple fields associated with amount in lines. Should [`listUnitPrice`](https://github.com/folio-org/acq-models/blob/master/mod-orders-storage/schemas/cost.json#L6) field be used to calculate the amount?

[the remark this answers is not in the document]
pautau-s: [poLineEstimatedPrice](https://github.com/folio-org/acq-models/blob/c3573085dcb17408c7c3eba09ec7a65dd195a775/mod-orders-storage/schemas/cost.json#L43) should be used

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -661,6 +669,57 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
...
+            MonetaryAmount totalEstimatedPriceMoney = Money.of(totalEstimatedPrice, currency);
+            /**
+             * calculate remaining amount to carry forward, required if there are more fund distributions with percentage and amount
+             * currentAmount = totalEstimatedPrice * value/100;
+             * remainingAmount = totalEstimatedPrice - currentAmount;
+             */
+            MonetaryAmount currentAmount = totalEstimatedPriceMoney.with(MonetaryOperators.percent(percentValue));
+            remainingAmount = remainingAmount.subtract(currentAmount);
+
+            if (remainingPercent < 0) {
```
pautau-s: Description says fundDistributions must add upto 100% of total.  Doesn't that mean that their sum should be equal to the total value?

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
javalkar-varun: Done, converting all to the amount.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
javalkar-varun: I followed UI. In UI `Remaining amount to be distributed` in Fund distribution is set to total and subtracted from the total at runtime. Therefore, the total equals `0` when the sum equals the total.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -661,6 +669,45 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
       .thenApply(v -> compPO);
   }
 
+  private CompositePurchaseOrder validateFundDistributionTotal(CompositePurchaseOrder compPO) {
+    if (compPO.getWorkflowStatus() == WorkflowStatus.OPEN) {
```
pautau-s: now this method is part of the `openOrder`, it guarantees that the order status is open

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
pautau-s: @craigmcnally could you clarify which of these is true: how it's described in the ticket or how it's implemented in the UI

craigmcnally: It almost sounds like you're both saying the same thing...

If we have a total of $100

Valid:
* 50%, 50%
* $25, $75
* $50, 50%

Invalid:
* 50% (need to distribute the other 50% to one or more funds)
* 33%, 33%, 33% (only adds up to 99%)
* $50, 25%, only adds up to $75)

pautau-s: In that case logic has to be updated

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -661,6 +669,43 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
...
+            MonetaryAmount poLineEstimatedPriceMoney = Money.of(poLineEstimatedPrice, currency);
+            // convert percent to amount
+            percentToAmount = poLineEstimatedPriceMoney.with(MonetaryOperators.percent(value));
+            amountValueMoney = percentToAmount;
+          }
+
+          remainingAmount = remainingAmount.subtract(amountValueMoney);
+
+          MonetaryAmount zeroMoney = Money.of(0, currency);
+          if (remainingAmount.isLessThan(zeroMoney)) {
```
pautau-s: This condition rejects only those cases when the sum of distributions is greater than poLineEstimatedPrice

pautau-s: There is an example how to handle fundDistibutions validation https://github.com/folio-org/mod-invoice/blob/1ca896afca3b6c26c7687578ebec95f0bf96170a/src/main/java/org/folio/rest/impl/InvoiceHelper.java#L524

javalkar-varun: Got it but I _think_ we don't have to do all that. This one line change;(`if (!remainingAmount.isZero())`) will reject all cases (greater than, less than poLineEstimatePrice). Let me know if any case is missing.

src/main/java/org/folio/rest/impl/PurchaseOrderHelper.java
```diff
@@ -672,7 +672,7 @@ private void validateRenewalInfo(CompositePurchaseOrder compPO) {
   private CompositePurchaseOrder validateFundDistributionTotal(CompositePurchaseOrder compPO) {
     for (CompositePoLine cPoLine : compPO.getCompositePoLines()) {
 
-      if (cPoLine.getCost().getPoLineEstimatedPrice() != null) {
+      if (cPoLine.getCost().getPoLineEstimatedPrice() != null && cPoLine.getFundDistribution().size() >0 ) {
```
craigmcnally: nitpick syntax `>0` -> `> 0`

PR #407, opened 2021-01-20
buildMaven.yml

ankitasen-ubmainz:
FOLIO-2913: SPIKE: Evaluate GitHub actions


funkymalc (approved):
Hi @ankitasen-ubmainz - Approved with some comments/questions:

1.  I don't it's necessary to define 'okapiUrl' in this pipeline.  It's not used anywhere. 
2.  'folioRegistry' probably should be a globally (site-wide).   I'll add FOLIO_REGISTRY to the site-wide vars. 
3. Shouldn't 'on.push.branches' be more inclusive? (all branches)

PR #474, opened 2021-07-16
[MODFISTO-247] Updating a transaction summary if it already exists

damien-git:
## Purpose
[MODFISTO-247](https://issues.folio.org/browse/MODFISTO-247)

## Approach
Do not try to create a transaction summary if there is already one for the order id. Update it instead.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/service/finance/transaction/TransactionSummariesService.java
+++ b/src/main/java/org/folio/service/finance/transaction/TransactionSummariesService.java
@@ -17,6 +17,7 @@ public class TransactionSummariesService {
 
     private static final String ENDPOINT = "/finance/order-transaction-summaries";
     private static final String BY_ID_ENDPOINT = ENDPOINT + "/{id}";
+    private static final String BY_ID_GET_ENDPOINT = "/finance-storage/order-transaction-summaries/{id}";
 
     private final RestClient restClient;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/service/finance/transaction/TransactionSummariesService.java
```diff
@@ -61,16 +62,18 @@ public TransactionSummariesService(RestClient restClient) {
       String orderId = Stream.concat(orderIdFromCreate, orderIdFromStorage)
         .findFirst()
         .orElse(null);
-        if (CollectionUtils.isEmpty(holder.getEncumbrancesFromStorage())) {
+      return getOrderTransactionSummary(orderId, requestContext)
+        .exceptionally(t -> null)
```
grabsefx, line 66: I beleive it was expected that exception will be raised only on 404 http code. But it seems creation of new transaction summary will bee procceded on any error, despite the http code

damien-git: I added code to only catch a 404, is that what you meant ? Not catching the other issues could also potentially have side effects, since we were not getting the summary there previously.

damien-git: Actually, I see this creates issues in the mod-orders integration tests because of the missing permission.

grabsefx: ah, so it looks like permission issue. thanks for pointing on this

PR #625, opened 2022-07-18
[MODORDERS-708]. Provide endpoint to validate fund distributions for a po line

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODORDERS-708

## Approach
Provide new endpoint to validate fund distributions for a PO line that will be called by UI to have consistent validation with BE

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -137,7 +137,7 @@
     },
     {
       "id": "order-lines",
-      "version": "3.1",
+      "version": "3.2",
       "handlers": [
         {
           "methods": [
@@ -301,6 +301,15 @@
             "orders-storage.po-lines.item.patch",
             "orders-storage.po-lines.item.get",
             "orders-storage.po-lines.item.put"
+          ]
+        },
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/orders/order-lines/fund-distributions/validate",
+          "permissionsRequired": [
+            "orders.po-lines.fund-distributions.validate"
           ]
         }
       ]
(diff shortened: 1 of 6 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -302,6 +302,15 @@
             "orders-storage.po-lines.item.get",
             "orders-storage.po-lines.item.put"
           ]
+        },
+        {
+          "methods": [
+            "PUT"
+          ],
+          "pathPattern": "/orders/order-lines/fund-distributions/validate",
```
okolawole-ebsco, line 310: nitpick: REST endpoints should have nouns instead of verbs so "validate" would not be ideal here. Actions such as this would extra data points to work and I doubt FOLIO endpoints follow this as well. That is why this is just FYI for now. It could be something like the below:
`
curl --location --request PUT '/orders/order-lines/fund-distributions' \
--header 'Content-Type: application/folio.orders.fundvalidate+json'
`
I doubt our RAMLs handle content types...

src/main/java/org/folio/orders/utils/FundDistributionUtils.java
```diff
@@ -12,42 +15,45 @@
...
+  }
 
-      if (cPoLine.getCost().getPoLineEstimatedPrice() != null && !cPoLine.getFundDistribution().isEmpty()) {
-        Double poLineEstimatedPrice = cPoLine.getCost().getPoLineEstimatedPrice();
-        if (poLineEstimatedPrice == 0d) {
-          validateZeroPrice(cPoLine.getFundDistribution());
-          continue;
-        }
-        BigDecimal remainingPercent = BigDecimal.valueOf(100);
+  public static void validateFundDistributionForPoLine(Cost cost, List<FundDistribution> fundDistributions) {
```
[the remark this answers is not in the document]
okolawole-ebsco: Nevermind, it looks like it is called externally.

okolawole-ebsco (approved):

descriptors/ModuleDescriptor-template.json
SerhiiNosko: Yes, agree that http method should be as verb and endpoint name - noun, but I have not yet met handing based on content type in RMB in the same way as provided in comment, but definitely it's a good point

Abdulkhakimov (approved):

PR #638, opened 2022-08-11
Modorders 748-Adjusting Instance status, instance type and loan type defaults is ignored buy order creation

SinghAdes:
## Purpose

https://issues.folio.org/browse/MODORDERS-748
## Approach

Adjusting Instance status, instance type and loan type defaults .
## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/service/inventory/InventoryManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryManager.java
@@ -984,20 +984,12 @@ public class InventoryManager {
   }
 
   /**
-   * Caches id's in Vert.X Context and returns it by tenantId.entryType.key.
-   *
-   * @param entryType name of object whose id we want to get from cache
    * @return configuration value by entry type
    */
   public CompletableFuture<JsonObject> getEntryValue(String entryType, RequestContext requestContext) {
       CompletableFuture<String> entryTypeValueFuture;
-      String tenantId = TenantTool.tenantId(requestContext.getHeaders());
-      var entryTypeKey = String.format("%s.%s", tenantId, entryType);
-
       Context ctx = requestContext.getContext();
-
       Map<String, String> okapiHeaders = requestContext.getHeaders();
-
       entryTypeValueFuture = getEntryTypeValue(entryType, requestContext);
 
     return entryTypeValueFuture.thenCompose(key -> {
--- a/src/main/java/org/folio/service/inventory/InventoryManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryManager.java
@@ -748,7 +748,7 @@ public class InventoryManager {
         Lock lock = lockResult.result();
         try {
           vertx.setTimer(30000, timerId -> releaseLock(lock, lockName));
-          getAndCache(entryType, requestContext).whenComplete((ok, err) -> {
+          getEntryValue(entryType, requestContext).whenComplete((ok, err) -> {
             releaseLock(lock, lockName);
             if (err == null) {
               future.complete(ok);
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/service/inventory/InventoryManager.java
```diff
@@ -996,14 +996,7 @@ public CompletableFuture<JsonObject> getAndCache(String entryType, RequestContex
...
       Map<String, String> okapiHeaders = requestContext.getHeaders();
-
-      String entryTypeCachedValue = ctx.get(entryTypeKey);
-
-      if (entryTypeCachedValue != null) {
-        entryTypeValueFuture = CompletableFuture.completedFuture(entryTypeCachedValue);
-      } else {
-        entryTypeValueFuture = getEntryTypeValue(entryType, requestContext);
-      }
+      entryTypeValueFuture = getEntryTypeValue(entryType, requestContext);
```
SerhiiNosko, line 999: Basically you omitting cache by this code, can you please spend some time to investigate where this cache populated, maybe for case described in current ticket cache is not populated correctly.

Abdulkhakimov: I suggest expanding the number of arguments in entryTypeCacheKey and the problem will be solved.

Abdulkhakimov: Removing the basic logic of caching, not really a good approach

SerhiiNosko: If we remain cache - it is possible situation that for different instance ids can be different entity types, so if we want to support this cache - we need to introduce composite key: instanceId + entryType as cache key.
And even if we do this, lets imagine we have 2 instances of mod-orders, context object is not shared because vertx running not in the clustered mode, so it possible that in each instances will be stored different value of cache keys.

okolawole-ebsco: Please describe why the cache is being removed. It is not clear from looking at this change OR PR details

SinghAdes: > Please describe why the cache is being removed. It is not clear from looking at this change OR PR details
[MODORDERS-748](https://issues.folio.org/browse/MODORDERS-748)
When changing the default settings for Instance status, instance type and loan type defaults and opening a NEW order that is set to create instance holding item. The choose defaults are not used. System still uses previous default settings because of cache.

SerhiiNosko: @SinghAdes let's remove cache(the approach that you committed initially) for MG bug fix release, but also need to delete line of code to populate cache [here](https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/service/inventory/InventoryManager.java#L1010)  and rename 'getAndCache' method.
For Nolana I will create story to introduce full-fledged Caffeine cache instead for this place as @okolawole-ebsco did for SRM module [here in this PR](https://github.com/folio-org/mod-source-record-manager/pull/674/files).
I will post additional comment with link after story will be created, thanks

SerhiiNosko: Here is new story created: https://issues.folio.org/browse/MODORDERS-750

src/main/java/org/folio/service/inventory/InventoryManager.java
```diff
@@ -989,26 +989,18 @@ private CompletableFuture<String> getSourceId(RequestContext requestContext) {
    * @param entryType name of object whose id we want to get from cache
    * @return configuration value by entry type
    */
```
SerhiiNosko, line 991: Also need to correct this java doc

SerhiiNosko (approved):

PR #643, opened 2022-08-31
MODORDERS-753: Implement processing of DI Kafka events by mod-orders

Aliaksandr-Fedasiuk:
## Purpose
Implement processing of DI Kafka events by mod-orders

## Approach
- add dependency on [folio-kafka-wrapper](https://github.com/folio-org/mod-pubsub/blob/master/folio-kafka-wrapper) library
- add dependency on [data-import-processing-core](https://github.com/folio-org/data-import-processing-core) library
- implement JobProfileSnapshot cache 
- implement [AsyncRecordHandler](https://github.com/folio-org/folio-kafka-wrapper/blob/master/src/main/java/org/folio/kafka/AsyncRecordHandler.java) to receive event about Order creation
- add stub implementation of [EventHandler](https://github.com/folio-org/data-import-processing-core/blob/master/src/main/java/org/folio/processing/events/services/handler/EventHandler.java) to handle received event and create new Order from incoming MARC record (will be added in scope of [MODORDERS-754](https://issues.folio.org/browse/MODORDERS-754))
- create and subscribe Kafka consumer using [KafkaConsumerWrapper](https://github.com/folio-org/folio-kafka-wrapper/blob/master/src/main/java/org/folio/kafka/KafkaConsumerWrapper.java) on the corresponding event
- add tests

## Learning
[MODORDERS-753](https://issues.folio.org/browse/MODORDERS-753)


src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -35,15 +51,43 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
         DatabindCodec.mapper().setConfig(deserializationConfig);
         DatabindCodec.prettyMapper().setConfig(deserializationConfig);
         SpringContextUtil.init(vertx, context, ApplicationConfig.class);
-        handler.complete();
+        SpringContextUtil.autowireDependencies(this, context);
+        deployConsumersVerticles(vertx).onSuccess(car -> {
```
Abdulkhakimov, line 55: Hi. What does "car" mean?

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -35,15 +51,43 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
+    vertx.registerVerticleFactory(verticleFactory);
+
+    Promise<String> deployRawMarcChunkConsumer = Promise.promise();
+
+    vertx.deployVerticle(getVerticleName(verticleFactory, OrderConsumerVerticle.class),
+      new DeploymentOptions()
+        .setWorker(true)
+        .setInstances(orderConsumerInstancesNumber), deployRawMarcChunkConsumer);
+
+    return GenericCompositeFuture.all(Arrays.asList(
```
Abdulkhakimov, line 86: Could we return just simple future instead of composite?

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
```diff
@@ -0,0 +1,105 @@
...
+        .kafkaConfig(kafkaConfig)
+        .loadLimit(loadLimit)
+        .globalLoadSensor(globalLoadSensor)
+        .subscriptionDefinition(subscriptionDefinition)
+        .processRecordErrorHandler(getErrorHandler())
+        .build();
+      kafkaConsumersStorage.addConsumer(event, consumerWrapper);
+
+      futures.add(consumerWrapper.start(getHandler(),
+        constructModuleName() + "_" + getClass().getSimpleName()));
```
Abdulkhakimov, line 87: Maybe would be better replace by using String.format?

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -105,6 +109,36 @@
 @ComponentScan({ "org.folio" })
 public class ApplicationConfig {
 
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  @Value("${KAFKA_HOST:kafka}")
```
okolawole-ebsco, line 114: nitpick: I think this should default to localhost.

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -105,6 +109,36 @@
...
+  private String kafkaHost;
+  @Value("${KAFKA_PORT:9092}")
+  private String kafkaPort;
+  @Value("${OKAPI_URL:http://okapi:9130}")
```
okolawole-ebsco, line 118: nitpick: I think this should default to `http://localhost:9130`

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -105,6 +109,36 @@
...
+  private String okapiUrl;
+  @Value("${REPLICATION_FACTOR:1}")
```
okolawole-ebsco, line 120: is `REPLICATION_FACTOR` exposed in any other folio module? is it related to kafka?

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -1,27 +1,43 @@
...
 
 /**
  * The class initializes vertx context adding spring context
  */
 public class InitAPIs implements InitAPI {
-  private final Logger logger = LogManager.getLogger();
+
+  private final Logger LOGGER = LogManager.getLogger();
+
+  private static final String SPRING_CONTEXT_KEY = "springContext";
```
okolawole-ebsco, line 37: why is the spring context key added now, but was not present before?

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -35,15 +51,43 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
         }
       });
   }
+
+  private Future<?> deployConsumersVerticles(Vertx vertx) {
+    AbstractApplicationContext springContext = vertx.getOrCreateContext().get(SPRING_CONTEXT_KEY);
+    VerticleFactory verticleFactory = springContext.getBean(SpringVerticleFactory.class);
+    vertx.registerVerticleFactory(verticleFactory);
+
+    Promise<String> deployRawMarcChunkConsumer = Promise.promise();
```
okolawole-ebsco, line 79: add the word "promise" to the name of this variable.

src/main/java/org/folio/rest/impl/InitAPIs.java
```diff
@@ -35,15 +51,43 @@ public void init(Vertx vertx, Context context, Handler<AsyncResult<Boolean>> res
...
         } else {
-          logger.error("Failure to init API", result.cause());
+          LOGGER.error("Failure to init API", result.cause());
           resultHandler.handle(Future.failedFuture(result.cause()));
         }
       });
   }
+
+  private Future<?> deployConsumersVerticles(Vertx vertx) {
+    AbstractApplicationContext springContext = vertx.getOrCreateContext().get(SPRING_CONTEXT_KEY);
```
okolawole-ebsco, line 75: I believe i have seen TODO's in mod-source-record-manager code to update this approach with Spring. Do you know what the enhancement would be OR how to fix the TODOs?

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class JobProfileSnapshotCache {
+
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  @Value("${orders.profile-snapshot-cache.expiration.time.seconds:3600}")
```
okolawole-ebsco, line 29: Something rubs me the wrong way about this configuration name. I think it is the hypen and how long it is. Maybe `orders.cache.jobprofile.expiration.seconds`

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
```diff
@@ -0,0 +1,72 @@
...
+  private long cacheExpirationTime;
+  private AsyncCache<String, Optional<ProfileSnapshotWrapper>> cache;
```
okolawole-ebsco, line 31: I think it is fine to return optional in the api methods of this class. Having optional saved into the cache means that it is possible to put Optional.empty into the cache. Is that intentional?

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.github.benmanes.caffeine.cache.AsyncCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class JobProfileSnapshotCache {
```
okolawole-ebsco, line 25: I like the intent of this class!

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
```diff
@@ -0,0 +1,72 @@
...
+    } catch (Exception e) {
+      LOGGER.warn("Error loading ProfileSnapshotWrapper by id: '{}'", profileSnapshotId, e);
+      return Future.failedFuture(e);
+    }
+  }
+
+  private CompletableFuture<Optional<ProfileSnapshotWrapper>> loadJobProfileSnapshot(String profileSnapshotId, OkapiConnectionParams params) {
+    LOGGER.debug("Trying to load jobProfileSnapshot by id  '{}' for cache, okapi url: {}, tenantId: {}", profileSnapshotId, params.getOkapiUrl(), params.getTenantId());
+
+    return RestUtil.doRequest(params, "/data-import-profiles/jobProfileSnapshots/" + profileSnapshotId, HttpMethod.GET, null)
```
okolawole-ebsco, line 53: Is using RestUtil the standard pattern in mod-orders?

okolawole-ebsco, line 53: nitpick: the URI should probably be saved in a constant somewhere.

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private static final GlobalLoadSensor globalLoadSensor = new GlobalLoadSensor();
+
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  @Autowired
+  @Qualifier("newKafkaConfig")
+  private KafkaConfig kafkaConfig;
+
+  @Value("${orders.kafka.OrderConsumer.loadLimit:5}")
```
okolawole-ebsco, line 45: might want to change `OrderConsumer` here. It is too generic.

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
```diff
@@ -0,0 +1,105 @@
...
+  private int loadLimit;
+
+  @Autowired
+  private KafkaConsumersStorage kafkaConsumersStorage;
+
+  @Autowired
+  @Qualifier("OrdersKafkaHandler")
+  private AsyncRecordHandler<String, String> ordersKafkaHandler;
```
okolawole-ebsco, line 53: please, change the name of this variable. Imagine if there was more than one kafka handler in mod-orders, what would you name this?

src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorage.java
```diff
@@ -0,0 +1,33 @@
+package org.folio.verticle.consumers.consumerstorage;
+
+import org.folio.kafka.KafkaConsumerWrapper;
+
+import java.util.Collection;
+
+/**
+ * Storage for Kafka consumers. Consumers are not tenant specific.
+ */
+public interface KafkaConsumersStorage {
```
okolawole-ebsco, line 10: Consider adding this to folio-kafka-wrapper so that SRM and mod-orders use it.

src/test/java/org/folio/di/StubTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+import java.util.UUID;
+
+import static org.junit.Assert.assertTrue;
+
+@RunWith(VertxUnitRunner.class)
+public class StubTest extends  DiAbstractRestTest {
```
okolawole-ebsco, line 15: Will this class be removed later?

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.folio.DataImportEventTypes.DI_COMPLETED;
+import static org.folio.DataImportEventTypes.DI_ERROR;
+import static org.springframework.beans.factory.config.BeanDefinition.SCOPE_PROTOTYPE;
+
+@Component
+@Scope(SCOPE_PROTOTYPE)
+public class OrderConsumerVerticle extends AbstractVerticle {
```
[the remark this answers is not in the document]
SerhiiNosko: Just thinking, if we implementing generic contract with generic name - potentually we can use this Kafka contract not only for DI. For example provide to customers Kafka as alternative to replace loading huge amount of orders via REST API.

VRohach (approved):

src/test/java/org/folio/di/StubTest.java
Aliaksandr-Fedasiuk: It was stub test. Now it was changed to the original test.

src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorage.java
Aliaksandr-Fedasiuk: This feature has now been removed from mod-orders because it is used for flow control - this functionality is not needed in mod-orders now.

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: renamed to DataImportConsumerVerticle

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
Aliaksandr-Fedasiuk: agree. completed.

src/main/java/org/folio/rest/impl/InitAPIs.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/rest/impl/InitAPIs.java
Aliaksandr-Fedasiuk: private static final String SPRING_CONTEXT_KEY = "springContext" is private in org.folio.spring.SpringContextUtil (folio-di-support lib).
I need to get a SpringVerticleFactory bean from a Spring context. This operation was not required in the module before. Only in test I found: context.remove("springContext");

src/main/java/org/folio/config/ApplicationConfig.java
Aliaksandr-Fedasiuk: the default values used for the container in vagrant

src/main/java/org/folio/config/ApplicationConfig.java
Aliaksandr-Fedasiuk: the default values used for the container in vagrant

src/main/java/org/folio/rest/impl/InitAPIs.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/rest/impl/InitAPIs.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
Aliaksandr-Fedasiuk: What we will gain from this?

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
Aliaksandr-Fedasiuk: we will fix it during using this cache

Abdulkhakimov (approved):

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -105,6 +109,36 @@
...
+    KafkaConfig kafkaConfig = KafkaConfig.builder()
+      .envId(envId)
+      .kafkaHost(kafkaHost)
+      .kafkaPort(kafkaPort)
+      .okapiUrl(okapiUrl)
+      .replicationFactor(replicationFactor)
+      .maxRequestSize(maxRequestSize)
+      .build();
+
+    LOGGER.info("kafkaConfig: " + kafkaConfig);
```
Abdulkhakimov, line 138: Could you replace the line to built-in formatting like as:  LOGGER.info("kafkaConfig: {}", kafkaConfig); ?

src/main/java/org/folio/verticle/OrderConsumerVerticle.java
Aliaksandr-Fedasiuk: completed

src/main/java/org/folio/config/ApplicationConfig.java
okolawole-ebsco: sounds good.

src/main/java/org/folio/config/ApplicationConfig.java
okolawole-ebsco: sounds good

src/main/java/org/folio/rest/impl/InitAPIs.java
okolawole-ebsco: gotcha.

src/main/java/org/folio/rest/impl/InitAPIs.java
okolawole-ebsco: looks good.

src/main/java/org/folio/config/ApplicationConfig.java
Aliaksandr-Fedasiuk: yes. we used this parameter in all our modules related to Kafka

okolawole-ebsco: I think it would be good if it is called `KAFKA_REPLICATION_FACTOR`.

src/main/java/org/folio/rest/impl/InitAPIs.java
Aliaksandr-Fedasiuk: I need some investigation into this. I hope I can work with this in the scope of future tasks.

src/main/java/org/folio/config/ApplicationConfig.java
Aliaksandr-Fedasiuk: For this, we should rename it on our other modules.

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
okolawole-ebsco: from aliaks, `we will use this code in nearest future and we will be able to add your remark the moment we use this code. At this point, the task was to prepare the code so we could quickly start adding the necessary business logic.`

SerhiiNosko (approved):

src/main/java/org/folio/service/caches/JobProfileSnapshotCache.java
Aliaksandr-Fedasiuk: this is the standard pattern for Folio. I think we can use it.

okolawole-ebsco (approved):

src/main/java/org/folio/config/ApplicationConfig.java
Aliaksandr-Fedasiuk: done

PR #673, opened 2022-12-09
MODORDERS-775 - Add deduplication mechanism for Create Order event

RuslanLavrov:
## Purpose
to avoid events duplication during data import for order creation

## Approach
* add orderIdStorage implementation to handle possible events duplication
* add tests
* fix tests run on jenkins

## Learning
[MODORDERS-775](https://issues.folio.org/browse/MODORDERS-775)


KaterynaSenchenko (approved):

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -0,0 +1,10 @@
+{
+  "tables": [],
+  "scripts": [
+    {
+      "run": "after",
+      "snippetPath": "create_records_orders_table.sql",
+      "fromModuleVersion": "mod-source-record-manager-12.6.0"
```
okolawole-ebsco, line 7: should this reference mod-orders instead?

src/main/resources/templates/db_scripts/create_records_orders_table.sql
```diff
@@ -0,0 +1,5 @@
+-- Creates table to store association between record id and order id
+CREATE TABLE IF NOT EXISTS records_orders (
```
okolawole-ebsco, line 2: is record here a Kafka record?

does this table get cleaned out sometimes?

SerhiiNosko (approved):

src/main/resources/templates/db_scripts/schema.json
RuslanLavrov: sure, thanks!

src/main/resources/templates/db_scripts/create_records_orders_table.sql
RuslanLavrov: here it means an srs record. For now, this table is not cleaned.

Aliaksandr-Fedasiuk (approved):

VRohach (approved):
@RuslanLavrov Perfect work!

okolawole-ebsco (approved):

PR #693, opened 2023-01-16
MODORDERS-773 - Implement action profile handler for order creation

RuslanLavrov:
## Purpose
to create order and order line from incoming MARC-BIB record

## Approach
* implement [EventHandler](https://github.com/folio-org/data-import-processing-core/blob/master/src/main/java/org/folio/processing/events/services/handler/EventHandler.java) to handle received event and create a new order and related order line from incoming MARC record
* add logic for 'activationDue' calculation
* add logic for po lines limit overriding
* override 'createInventory' po line field depending on actions in a job profile
* add tests

## Learning
[MODORDERS-773](https://issues.folio.org/browse/MODORDERS-773)

### DIFF
--- a/src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
+++ b/src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
@@ -10,7 +10,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.ActionProfile;
 import org.folio.DataImportEventPayload;
-import org.folio.DataImportEventTypes;
 import org.folio.helper.PurchaseOrderHelper;
 import org.folio.helper.PurchaseOrderLineHelper;
 import org.folio.kafka.exception.DuplicateEventException;
@@ -24,6 +23,7 @@ import org.folio.rest.core.exceptions.HttpException;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.CompositePoLine;
 import org.folio.rest.jaxrs.model.CompositePurchaseOrder;
+import org.folio.rest.jaxrs.model.CompositePurchaseOrder.WorkflowStatus;
 import org.folio.rest.jaxrs.model.EntityType;
 import org.folio.service.configuration.ConfigurationEntriesService;
 import org.folio.service.dataimport.IdStorageService;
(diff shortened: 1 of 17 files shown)


src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
```diff
@@ -0,0 +1,325 @@
...
+                                 JobProfileSnapshotCache jobProfileSnapshotCache) {
+    this.purchaseOrderHelper = purchaseOrderHelper;
+    this.poLineHelper = poLineHelper;
+    this.configurationEntriesService = configurationEntriesService;
+    this.idStorageService = idStorageService;
+    this.jobProfileSnapshotCache = jobProfileSnapshotCache;
+  }
+
+  @Override
+  public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload dataImportEventPayload) {
```
SerhiiNosko, line 109: We should not use java.util.CompletableFuture, please use Verx Future

SerhiiNosko: I see now, it is interface from data-import-processing-core, and this library uses CompletableFuture

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
```diff
@@ -0,0 +1,325 @@
...
+            }
+            LOGGER.warn("saveOrder:: Error during creation order in the storage", e);
+            return Future.failedFuture(e);
+          });
+      });
+  }
+
+  private Future<CompositePoLine> saveOrderLines(String orderId, DataImportEventPayload dataImportEventPayload,
+                                                 JsonObject tenantConfig, RequestContext requestContext) {
+    CompositePoLine poLine = Json.decodeValue(dataImportEventPayload.getContext().get(ORDER_LINES_KEY), CompositePoLine.class);
```
SerhiiNosko, line 193: Now it is possible to save only one po line?

KaterynaSenchenko (approved):

SerhiiNosko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
RuslanLavrov: during record import only one po line should be created per record. So far in the current PR a new order is created per every incoming record, but in subsequent changes the logic will be changed to create one order per multiple incoming records based on po lines limit per one order

SerhiiNosko: got it, thanks

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
RuslanLavrov: yes, it is required by the interface contract, but all async results of the handler and di-processing-core execution are handled by vertx thread

src/main/java/org/folio/verticle/DataImportConsumerVerticle.java
```diff
@@ -43,6 +44,9 @@ public class DataImportConsumerVerticle extends AbstractVerticle {
   @Value("${orders.kafka.OrderConsumer.loadLimit:5}")
   private int loadLimit;
 
+  @Value("${mod.orders.dataimport.kafka.max.distribution.number:100}")
```
okolawole-ebsco, line 47: i think this should have started with just "orders" like loadLimit

PR #696, opened 2023-01-20
MODORDERS-817: Associate mapped POLines with single Order based on POLine limit value

Aliaksandr-Fedasiuk:
## Purpose
Each MARC Bib record is mapped in a single POLine. Add a mechanism for associating mapped POLines with a single Order based on the order of the MARC Bib record in the file and the POLine limit value.

## Approach
Thr service is created that will save the orderId if it has not yet been created before or return the orderId if it has already been created.

## Learning
[MODORDERS-817](https://issues.folio.org/browse/MODORDERS-817)


RomanChernetskyi (approved):

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
```diff
@@ -159,6 +167,24 @@ private Map<String, String> extractOkapiHeaders(DataImportEventPayload dataImpor
     return headers;
   }
 
+  private Future<String> generateSequentialOrderId(DataImportEventPayload dataImportEventPayload, JsonObject tenantConfig) {
+    LOGGER.debug("generateSequentialOrderId :: jobExecutionId: {} ", dataImportEventPayload.getJobExecutionId());
+
+    String tenantId = dataImportEventPayload.getTenant();
+    if (isEmpty(tenantConfig.getString(PO_LINES_LIMIT_PROPERTY))) {
+      return Future.failedFuture(new IllegalArgumentException("Parameter poLines-limit is missing."));
```
VRohach, line 175: @Aliaksandr-Fedasiuk Let's add in the log at least jobExecutionId

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
```diff
@@ -159,6 +167,24 @@ private Map<String, String> extractOkapiHeaders(DataImportEventPayload dataImpor
...
+    }
+    Integer poLinesLimit = Integer.valueOf(tenantConfig.getString(PO_LINES_LIMIT_PROPERTY));
+
+    Record parsedMarcBibRecord = new JsonObject(dataImportEventPayload.getContext().get(MARC_BIBLIOGRAPHIC.value())).mapTo(Record.class);
+    if (parsedMarcBibRecord.getOrder() == null) {
+      return Future.failedFuture(new IllegalArgumentException("Order parameter is missing."));
```
VRohach, line 181: The same about log.

VRohach (approved):

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
RuslanLavrov: propose not to fail event processing and instead to use 1 as the default poLines limit, if the parameter is absent (like logic in the mod-orders services)

Aliaksandr-Fedasiuk: done

src/main/java/org/folio/service/dataimport/handlers/CreateOrderEventHandler.java
Aliaksandr-Fedasiuk: done

SerhiiNosko (approved):

RuslanLavrov (approved):

PR #713, opened 2023-02-20
Add limit for retrieving organizations

RomanChernetskyi:
## Purpose
Prevent returning of 10 first organizations by default if the limit is not defined in the request params

## Approach
Set limit to 1000 for getting organization for MappingParametersCache

### DIFF
--- a/src/main/java/org/folio/service/caches/MappingParametersCache.java
+++ b/src/main/java/org/folio/service/caches/MappingParametersCache.java
@@ -28,9 +28,11 @@ import java.util.concurrent.TimeUnit;
 public class MappingParametersCache {
   private static final Logger LOGGER = LogManager.getLogger();
   public static final String ORGANIZATIONS = "/organizations/organizations?limit=%s";
-  public static final int ORGANIZATIONS_LIMIT = 1000;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/service/caches/MappingParametersCache.java
```diff
@@ -27,7 +27,8 @@
 @Component
 public class MappingParametersCache {
   private static final Logger LOGGER = LogManager.getLogger();
-  public static final String ORGANIZATIONS = "/organizations/organizations";
+  public static final String ORGANIZATIONS = "/organizations/organizations?limit=%s";
+  public static final int ORGANIZATIONS_LIMIT = 1000;
```
KaterynaSenchenko, line 31: let's make it configurable, something like in SRM 
https://github.com/folio-org/mod-source-record-manager/blob/master/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java#L96

RomanChernetskyi: Thanks, done

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #804, opened 2023-12-12
MODORDERS-974. Add validation for claimingActive, claimingInterval fields

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODORDERS-974


Dmitriy-Butramyou (approved):

src/main/java/org/folio/service/orders/BaseValidationService.java
```diff
@@ -0,0 +1,24 @@
...
+
+public abstract class BaseValidationService {
+  /**
+   * Checks if claiming interval is greater than 0 if claiming is active.
+   *
+   * @param claimingActive is claiming active
+   * @param claimingInterval claiming interval
+   * @return list of error codes
+   */
+  public List<ErrorCodes> checkClaimingConfig(Boolean claimingActive, Integer claimingInterval) {
```
Abdulkhakimov, line 17: Maybe better to use protected access modifier here. Because you are calling this method from subclasses

SerhiiNosko: Agree, fixed

Abdulkhakimov (approved):

PR #806, opened 2023-12-14
[MODORDERS-964] Update acq-models

alb3rtino:
## Purpose
This PR updates `acq-models` to include the `customFields` attribute to purchase orders and purchase order lines. This is part of the custom fields integration. See [MODORDERS-964](https://issues.folio.org/browse/MODORDERS-964).

## Prerequisite
* [MODORDSTOR PR#374](https://github.com/folio-org/mod-orders-storage/pull/374)
## Related
* [MODGOBI PR#220](https://github.com/folio-org/mod-gobi/pull/220)
* [MODINVOICE PR#460](https://github.com/folio-org/mod-invoice/pull/460)
* [MODINVOSTO PR#167](https://github.com/folio-org/mod-invoice-storage/pull/167)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "orders",
-      "version": "12.0",
+      "version": "12.1",
       "handlers": [
         {
           "methods": [
@@ -143,7 +143,7 @@
     },
     {
       "id": "order-lines",
-      "version": "3.2",
+      "version": "3.3",
       "handlers": [
         {
           "methods": [
@@ -1024,7 +1024,7 @@
     },
     {
       "id": "orders-storage.purchase-orders",
-      "version": "8.0"
+      "version": "8.1"
     },
     {
       "id": "orders-storage.alerts",
@@ -1044,7 +1044,7 @@
     },
     {
       "id": "orders-storage.po-lines",
-      "version": "12.0"
+      "version": "12.3"
     },
     {
       "id": "orders-storage.order-templates",


SerhiiNosko:
@alb3rtino we also need to create PR with updating models for mod-gobi repository, because it also uses the same schemas

alb3rtino:
[MODGOBI PR#220](https://github.com/folio-org/mod-gobi/pull/220)

SerhiiNosko (approved):

SerhiiNosko:
checked that also need to create similar PRs in mod-invoice, mod-invoice-storage
https://github.com/folio-org/mod-invoice/blob/master/descriptors/ModuleDescriptor-template.json#L574

alb3rtino:
I've updated to provided interface versions to
```
order-lines -> 3.3
orders -> 12.1
```
and the required interface versions to
```
orders-storage.po-lines -> 12.3
orders-storage.purchase-orders -> 8.1
```

This requires a merge of [MODORDSTOR PR#374](https://github.com/folio-org/mod-orders-storage/pull/374) first, so i've added the `'do not merge'` tag.

alb3rtino:
@SerhiiNosko I've made updates to the interface versions. Could you please review them? I'm uncertain whether the updated required interfaces require a module major version increase or not.

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "orders",
-      "version": "12.0",
+      "version": "12.1",
```
SerhiiNosko, line 7: As I remember if we increase from 12.0 to 12.1 - Okapi will not throw an error, even if in mod-gobi required version is 12.0 https://github.com/folio-org/mod-gobi/blob/master/descriptors/ModuleDescriptor-template.json#L157 
If we for example we increased from 12.0 to 13.0 and in mod-gobi remain 12.0 - Okapi will throw an error during enabling the module. 
But you can check my words using karate tests(acquisition folder in folio-integration-test) or just locally.

Basically we not always follow the rule to update minor version when adding new field, but in case when you increased it also would be a good practice to increase in these modules as well:
https://github.com/folio-org/mod-invoice/blob/master/descriptors/ModuleDescriptor-template.json#L664
https://github.com/folio-org/mod-gobi/blob/master/descriptors/ModuleDescriptor-template.json#L157
https://github.com/folio-org/mod-ebsconet/blob/master/descriptors/ModuleDescriptor-template.json#L6

SerhiiNosko: Basically I am ok to keep major versions for example in mod-gobi 12.0 and in orders increase to 12.1. It should work

SerhiiNosko (approved):

PR #814, opened 2024-01-05
[MODORDERS-984] Update po line receipt status after updating the piece

damien-git:
## Purpose
[MODORDERS-984](https://issues.folio.org/browse/MODORDERS-984) - Receipt status remains "Fully received" after unreceiving piece

## Approach
Update the po line receipt status after updating the piece.

[Integration test PR](https://github.com/folio-org/folio-integration-tests/pull/1176)

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
+++ b/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
@@ -76,7 +76,7 @@ public class PieceUpdateFlowManager {
       .compose(verifyReceiptStatus -> pieceStorageService.updatePiece(holder.getPieceToUpdate(), requestContext)
         .map(verifyReceiptStatus))
       .map(verifyReceiptStatus -> {
-        if (verifyReceiptStatus) {
+        if (Boolean.TRUE.equals(verifyReceiptStatus)) {
           JsonObject messageToEventBus = new JsonObject();
           messageToEventBus.put("poLineIdUpdate", holder.getPieceToUpdate().getPoLineId());
           pieceService.receiptConsistencyPiecePoLine(messageToEventBus, requestContext);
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
```diff
@@ -61,22 +61,29 @@ public Future<Void> updatePiece(Piece pieceToUpdate, boolean createItem, boolean
...
         if (receivingStatusStorage.compareTo(receivingStatusUpdate) != 0) {
           holder.getPieceToUpdate().setStatusUpdatedDate(new Date());
+          return true;
+        }
+        return false;
+      })
+      .compose(verifyReceiptStatus -> pieceStorageService.updatePiece(holder.getPieceToUpdate(), requestContext)
+        .map(verifyReceiptStatus))
+      .map(verifyReceiptStatus -> {
+        if (verifyReceiptStatus) {
```
azizbekxm, line 79: this line is confusing to reader. it would be better to use Boolean.TRUE.equals(verify...)

SerhiiNosko: also this is causing sonarqube issue

SerhiiNosko:
@damien-git what do you think to delete AsyncUtil.executeBlocking in ReceiptStatusConsistency in scope of this change as well, it is not needed and confuse

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
damien-git: I noticed the warning, but I think this is an error of the tool: given the code just before, values for `verifyReceiptStatus` can only be `TRUE` or `FALSE`, not `null`. And I think "`if (verifyReceiptStatus)`" is much more readable than "`if (Boolean.TRUE.equals(verifyReceiptStatus))`". But since @azizbekxm finds it confusing, I will switch to using `Boolean.TRUE.equals()`.

damien-git:
> @damien-git what do you think to delete AsyncUtil.executeBlocking in ReceiptStatusConsistency in scope of this change as well, it is not needed and confuse

I prefer not to do that in the scope of this ticket, because it is unrelated and also I would have to redo all the tests as it might break something else.

SerhiiNosko (approved):

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
azizbekxm: I agree your point of view, if (isChecked..) is more readable but verifyReceiptStatus is not implying boolean type is more like action so that is why I give recommendation to use Boolean.TRUE.equals().

azizbekxm (approved):

PR #850, opened 2024-03-04
MODORDERS-1053. Missing interface dependencies in module descriptor

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1053

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1019,6 +1019,10 @@
       "version": "1.0"
     },
     {
+      "id": "orders-storage.acquisition-methods",
+      "version": "1.0"
+    },
+    {
       "id": "configuration",
       "version": "2.0"
     },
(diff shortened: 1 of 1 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -1102,10 +1114,18 @@
...
+    },
     {
       "id": "finance.ledgers",
       "version": "2.0"
     },
+    {
+      "id": "finance-storage.ledgers",
+      "version": "4.0"
+    },
     {
```
azizbekxm, line 1129: finance-storage.order-transaction-summaries what about this permissoin, it is required or we can skip this

SerhiiNosko: after introducing new Batch Transaction API we don't use transaction-summaries endpoints from mod-orders any more, so it is not needed. For Q release these endpoints will be considered as deprecated, and for R we will completely remove them

azizbekxm (approved):

PR #852, opened 2024-03-05
MODOERDERS-1028-Add specific error code of barcode uniqueness error e…

yuntianhu:
**Purpose:**

During receiving pieces on the Full Receiving Screen View when populating barcodes the validation of related item barcode uniqueness occurs. In cases when the barcode is not unique, we need to provide a clear message about the reason to users.

**Actual result:**

Currently this error message is shown:

{
  "receivingResults" : [ {
    "poLineId" : "8c9da779-0cf2-4da1-9749-9c1954f368ef",
    "processedSuccessfully" : 0,
    "processedWithError" : 1,
    "receivingItemResults" : [ {
      "pieceId" : "44f2fdaa-b65c-4349-867c-56563761699a",
      "processingStatus" : {
        "type" : "failure",
        "error" : {
          "message" : "The item record failed to be updated",
          "code" : "itemUpdateFailed",
          "parameters" : [ ]
        }
      }
    } ]
  } ],
  "totalRecords" : 1
}
**Expected result:**

Error block is updated to cleary say that item barcode is not unique.

### DIFF
--- a/src/main/java/org/folio/helper/CheckinHelper.java
+++ b/src/main/java/org/folio/helper/CheckinHelper.java
@@ -6,6 +6,7 @@ import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 import static org.folio.orders.utils.HelperUtils.collectResultsOnSuccess;
 import static org.folio.rest.core.exceptions.ErrorCodes.ITEM_UPDATE_FAILED;
+import static org.folio.rest.core.exceptions.ErrorCodes.BARCODE_IS_NOT_UNIQUE;
 import static org.folio.service.inventory.InventoryManager.ITEM_ACCESSION_NUMBER;
 import static org.folio.service.inventory.InventoryManager.ITEM_BARCODE;
 import static org.folio.service.inventory.InventoryManager.ITEM_CHRONOLOGY;
@@ -217,11 +218,21 @@ public class CheckinHelper extends CheckinReceivePiecesHelper<CheckInPiece> {
       })
       // Add processing error if item failed to be updated
       .onFailure(e -> {
-        logger.error("Item associated with piece '{}' cannot be updated", piece.getId());
-        addError(piece.getPoLineId(), piece.getId(), ITEM_UPDATE_FAILED.toError());
+        produceErrorMessage(piece, e.toString());
         promise.complete(false);
       });
     return promise.future();
+  }
+
+  // This function produces error message based the error it receives from the checkinItem()
+  protected void produceErrorMessage(Piece piece, String error) {
+    if (error.contains("Barcode must be unique")) {
+      logger.error("The barcode associate with item '{}' is not unique, it cannot be updated", piece.getId());
+      addError(piece.getPoLineId(), piece.getId(), BARCODE_IS_NOT_UNIQUE.toError());
+    } else if (error != null) {
+      logger.error("Item associated with piece '{}' cannot be updated", piece.getId());
+      addError(piece.getPoLineId(), piece.getId(), ITEM_UPDATE_FAILED.toError());
+    }
   }
 
   private void updatePieceWithCheckinInfo(Piece piece) {
--- a/src/main/java/org/folio/helper/CheckinHelper.java
+++ b/src/main/java/org/folio/helper/CheckinHelper.java
@@ -5,8 +5,6 @@ import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 import static org.folio.orders.utils.HelperUtils.collectResultsOnSuccess;
-import static org.folio.rest.core.exceptions.ErrorCodes.ITEM_UPDATE_FAILED;
-import static org.folio.rest.core.exceptions.ErrorCodes.BARCODE_IS_NOT_UNIQUE;
 import static org.folio.service.inventory.InventoryManager.ITEM_ACCESSION_NUMBER;
 import static org.folio.service.inventory.InventoryManager.ITEM_BARCODE;
 import static org.folio.service.inventory.InventoryManager.ITEM_CHRONOLOGY;
--- a/src/main/java/org/folio/helper/CheckinHelper.java
+++ b/src/main/java/org/folio/helper/CheckinHelper.java
@@ -216,7 +216,7 @@ public class CheckinHelper extends CheckinReceivePiecesHelper<CheckInPiece> {
       })
       // Add processing error if item failed to be updated
       .onFailure(e -> {
-        addErrorForUpdatingItem(piece, e.toString());
+        addErrorForUpdatingItem(piece, e);
         promise.complete(false);
       });
     return promise.future();
--- a/src/main/java/org/folio/helper/CheckinHelper.java
+++ b/src/main/java/org/folio/helper/CheckinHelper.java
@@ -218,21 +218,10 @@ public class CheckinHelper extends CheckinReceivePiecesHelper<CheckInPiece> {
       })
       // Add processing error if item failed to be updated
       .onFailure(e -> {
-        produceErrorMessage(piece, e.toString());
+        addErrorForUpdatingItem(piece, e.toString());
         promise.complete(false);
       });
     return promise.future();
-  }
-
-  // This function produces error message based on the error it receives from the checkinItem()
-  protected void produceErrorMessage(Piece piece, String error) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -217,13 +218,23 @@ protected Future<Boolean> receiveInventoryItemAndUpdatePiece(JsonObject item, Pi
       })
       // Add processing error if item failed to be updated
       .onFailure(e -> {
-        logger.error("Item associated with piece '{}' cannot be updated", piece.getId());
-        addError(piece.getPoLineId(), piece.getId(), ITEM_UPDATE_FAILED.toError());
```
SerhiiNosko, line 221: this invocation of addError method() has 5 occurences with code ITEM_UPDATE_FAILED but only 2 of them are matches with ITEM_UPDATE_FAILED: CheckingHelper and ReceivingHelper.
All occurences in abstract class CheckingReceivePiecesHelper should not use ITEM_UPDATE_FAILED, they should use error codes like po line not found, title not foud, hodling can not be created - but we can improve this in next story.
For this story propose to move new function in base CheckinReceivePiecesHelper and use it in 2 places where ITEM_UPDATE_FAILED error code makes sense(when actualy dealing with item update): CheckingHelper and ReceivingHelper.
image

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -217,13 +218,23 @@ protected Future<Boolean> receiveInventoryItemAndUpdatePiece(JsonObject item, Pi
...
+        produceErrorMessage(piece, e.toString());
         promise.complete(false);
       });
     return promise.future();
   }
 
+  // This function produces error message based on the error it receives from the checkinItem()
```
SerhiiNosko, line 227: propose to rename function to be more self descriptive instead of adding comment  here.
addErrorForUpdatingItem()

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -217,13 +218,23 @@ protected Future<Boolean> receiveInventoryItemAndUpdatePiece(JsonObject item, Pi
...
+  protected void produceErrorMessage(Piece piece, String error) {
+    if (error.contains("Barcode must be unique")) {
```
SerhiiNosko, line 229: seems we don't have more clear way to handle this instead of relying on error message beased on this search results:
https://github.com/search?q=org%3Afolio-org+%22Barcode+must+be+unique%22&type=code&p=1
We can just move this string to constant
private static final String BARCODE_NOT_UNIQUE_MESSAGE = "Barcode must be unique";

src/main/java/org/folio/helper/CheckinHelper.java
SerhiiNosko: or you can do this refactoring to use correct error codes in CheckingReceivePiecesHelper in scope of this story, in this case we will notify UI devs that we have new error codes

yuntianhu: Thank you, Serhii, for point that out! will do

src/main/java/org/folio/helper/CheckinHelper.java
yuntianhu: yep, will do

src/main/java/org/folio/helper/CheckinHelper.java
yuntianhu: sure, will do.

src/main/java/org/folio/rest/core/exceptions/ErrorCodes.java
```diff
@@ -114,8 +114,8 @@ public enum ErrorCodes {
   FUND_LOCATION_RESTRICTION_VIOLATION("fundLocationRestrictionViolation", "One of the locations is restricted to be used by all funds."),
   ENCUMBRANCES_FOR_RE_ENCUMBER_NOT_FOUND("encumbrancesForReEncumberNotFound", "The encumbrances were correctly created during the rollover or have already been updated."),
   CLAIMING_CONFIG_INVALID("claimingConfigInvalid", "Claiming interval should be set and greater than 0 if claiming is active"),
-  TEMPLATE_NAME_ALREADY_EXISTS("templateNameNotUnique", "Template name already exists");
-
+  TEMPLATE_NAME_ALREADY_EXISTS("templateNameNotUnique", "Template name already exists"),
+  BARCODE_IS_NOT_UNIQUE("barcodeIsNotUnique", "Barcode must be unique, this barcode already exists");
```
azizbekxm, line 118: Suggest that "The barcode already exists" or "The barcode already exists. The barcode must be unique"

yuntianhu: cool, will change : ) thank you

src/test/java/org/folio/helper/CheckinHelperTest.java
```diff
@@ -192,8 +196,88 @@ void allCheckingPieceFieldsShouldBePopulated() {
...
+
+    assertNotNull(new CheckinHelper(checkinCollection2, okapiHeadersMock, requestContext.getContext()).getErrors());
+
+    Map<String, List<Piece>> map2 = new HashMap<>();
+    List<Piece> pieces2 = new ArrayList<>();
+    pieces2.add(new Piece().withId(pieceId2).withPoLineId(poLineId2));
+
+    assertNull(map2.put(poLineId2, pieces2));
+  }
+
```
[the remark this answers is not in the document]
yuntianhu: will take a look, thank you for pointing that out!

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -84,7 +85,7 @@ public abstract class CheckinReceivePiecesHelper<T> extends BaseHelper {
   private final Map<String, Map<String, Error>> processingErrors;
   private final Set<String> processedHoldingsParams;
   private final Map<String, String> processedHoldings;
-
+  private static final String BARCODE_NOT_UNIQUE_MESSAGE = "Barcode must be unique";
```
SerhiiNosko, line 88: better to declare constants at the top of class before declaring other variables

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -806,4 +807,14 @@ private static class PoLineAndTitleById {
     Map<String, CompositePoLine> poLineById;
     Map<String, Title> titleById;
   }
+
+  protected void addErrorForUpdatingItem(Piece piece, String error) {
+    if (error.contains(BARCODE_NOT_UNIQUE_MESSAGE)) {
+      logger.error("The barcode associate with item '{}' is not unique, it cannot be updated", piece.getId());
+      addError(piece.getPoLineId(), piece.getId(), BARCODE_IS_NOT_UNIQUE.toError());
+    } else {
+      logger.error("Item associated with piece '{}' cannot be updated", piece.getId());
```
SerhiiNosko, line 816: based on this log we could not see actual reason of failure, improving

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -806,4 +807,14 @@ private static class PoLineAndTitleById {
     Map<String, CompositePoLine> poLineById;
     Map<String, Title> titleById;
   }
+
+  protected void addErrorForUpdatingItem(Piece piece, String error) {
+    if (error.contains(BARCODE_NOT_UNIQUE_MESSAGE)) {
```
SerhiiNosko, line 812: exceptions can be created without message, for example you can try just writing: new IllegalArgumentException(); and it will compile.
So it's required non-null check here, improving

src/test/java/org/folio/helper/CheckinHelperTest.java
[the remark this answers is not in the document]
SerhiiNosko: resolved, tests improved

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
SerhiiNosko: improved

SerhiiNosko:
New error code works locally
image

SerhiiNosko (approved):

damien-git (approved):

PR #853, opened 2024-03-08
[MODORDERS-1043] Remove pending payment encumbrance link when encumbrance is deleted

damien-git:
## Purpose
[MODORDERS-1043](https://folio-org.atlassian.net/browse/MODORDERS-1043) - Remove pending payment encumbrance links when encumbrances are deleted

## Approach
- Pending payments used to be updated in POLInvoiceLineRelationService when an encumbrance was replaced by another one. Pending payment modifications are now part of the global batch modification, and include removal of encumbrance links when they are deleted.
- mod-finance-storage is now allowing updates of released encumbrances. We use that to simplify the batch calls: previously we needed 2 additional calls to unrelease encumbrances first and re-release them afterwards. Now we can simply update them in the single batch operation.
- Pending to pending strategy is now checking for connected invoices and returning an error if we need to remove an encumbrance and there is a connected invoice, unless that invoice is cancelled.
- New unit tests.

See also: [mod-finance PR](https://github.com/folio-org/mod-finance/pull/238), integration test PR.

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
+++ b/src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
@@ -69,14 +69,13 @@ public class EncumbranceService {
   public Future<Void> createOrUpdateEncumbrances(EncumbrancesProcessingHolder holder, RequestContext requestContext) {
     if (holder.getAllEncumbrancesQuantity() == 0)
       return Future.succeededFuture();
-    return unreleaseEncumbrancesFirst(holder, requestContext)
-      .compose(v -> batchProcess(holder, requestContext))
-      .compose(v -> releaseEncumbrancesAfter(holder, requestContext));
+    return batchProcess(holder, requestContext);
   }
 
   public Future<Void> batchProcess(EncumbrancesProcessingHolder holder, RequestContext requestContext) {
     List<Transaction> transactionsToCreate = prepareTransactionsToCreate(holder.getEncumbrancesForCreate());
     List<Transaction> transactionsToUpdate = prepareTransactionsToUpdate(holder);
+    transactionsToUpdate.addAll(holder.getPendingPaymentsToUpdate());
     List<Transaction> transactionsToDelete = prepareTransactionsToDelete(holder.getEncumbrancesForDelete());
     List<String> idsOfTransactionsToDelete =  transactionsToDelete.stream().map(Transaction::getId).toList();
     List<TransactionPatch> transactionPatches = Collections.emptyList();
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/service/finance/transaction/PendingToPendingEncumbranceStrategy.java
```diff
@@ -83,4 +92,35 @@ private List<EncumbranceRelationsHolder> getTransactionsToUpdate(List<Encumbranc
     return toUpdate;
   }
 
+  public Future<EncumbrancesProcessingHolder> addPendingPaymentsToUpdate(EncumbrancesProcessingHolder holder,
+      RequestContext requestContext) {
+    if (holder.getEncumbrancesForDelete().isEmpty()) {
+      return Future.succeededFuture(holder);
+    }
+    List<String> idsOfEncumbrancesToDelete = holder.getEncumbrancesForDelete().stream()
+      .map(EncumbranceRelationsHolder::getOldEncumbrance)
```
Abdulkhakimov, line 101: What do you think about adding filter to avoid NPE? Are you sure that oldEncumbrance in any case will be not null?

damien-git: I think we should check that in `getTransactionsToDelete()`. I am not sure why it was not breaking before...

Abdulkhakimov (approved):

src/main/java/org/folio/service/finance/transaction/PendingToPendingEncumbranceStrategy.java
damien-git: Actually I think I know why it was working before: there is always a new encumbrance created in `EncumbranceRelationsHoldersBuilder` for existing fund distributions (even though it won't be used in the pending to pending strategy). So the two (old and new) encumbrances cannot be both null. That change was not strictly needed, but it doesn't hurt to add a little check in case the implementation for `EncumbranceRelationsHoldersBuilder` changes in the future.

src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
```diff
@@ -384,36 +380,4 @@ protected void checkCustomTransactionError(Throwable fail) throws RuntimeExcepti
     }
   }
 
-  private Future<Void> unreleaseEncumbrancesFirst(EncumbrancesProcessingHolder holder, RequestContext requestContext) {
```
SerhiiNosko, line 387: great, reduced so many calls with this changes, that used in lot of flows
image

SerhiiNosko (approved):

PR #855, opened 2024-03-12
[MODORDERS-1044] - Do not update orderStatus for past fiscal years

imerabishvili:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1044

## Pre-Merge Checklist:

### DIFF
--- a/src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
+++ b/src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
@@ -14,6 +14,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletionException;
 
@@ -385,12 +386,14 @@ public class EncumbranceService {
   }
 
   private Future<String> getFiscalYearId(CompositePoLine poLine, RequestContext requestContext) {
-    String fundId = poLine.getFundDistribution()
+    Optional<String> fundId = poLine.getFundDistribution()
       .stream()
       .findFirst()
-      .map(FundDistribution::getFundId)
-      .orElse("");
-    return fiscalYearService.getCurrentFiscalYearByFundId(fundId, requestContext)
+      .map(FundDistribution::getFundId);
+    if (fundId.isEmpty()) {
+      return Future.succeededFuture("");
+    }
+    return fiscalYearService.getCurrentFiscalYearByFundId(fundId.get(), requestContext)
       .map(FiscalYear::getId);
   }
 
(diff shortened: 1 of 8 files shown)


damien-git:
@imerabishvili `AbstractOrderStatusHandler` does not create a composite order with the lines before calling `updateEncumbrancesOrderStatusAndReleaseIfClosed()`, so your new code is not finding any line to use. You should change `AbstractOrderStatusHandler$convert()` to add the lines to the composite order (they are already passed to `updateOrderStatus`, so you will just need to pass `poLines` as a parameter to `convert`).

damien-git:
Also in `EncumbranceService$getFiscalYearId` you should not try to get the fiscal year without a fund id.

SerhiiNosko:
Karate tests passed after suggested changes, thanks @damien-git for providing them

src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
```diff
@@ -302,20 +322,22 @@ private String buildEncumbrancesByOrderQuery(String orderId) {
...
+      + AND + "fiscalYearId == " + fiscalYearId;
   }
 
   private String buildEncumbrancesByPoLineQuery(String polineId) {
     return ENCUMBRANCE_CRITERIA + AND + "encumbrance.sourcePoLineId==" + polineId;
   }
 
-  private String buildUnreleasedEncumbrancesByPoLineQuery(String polineId) {
+  private String buildUnreleasedEncumbrancesByPoLineQuery(String polineId, String fiscalYearId) {
     return ENCUMBRANCE_CRITERIA
```
SerhiiNosko, line 337: ENCUMBRANCE_CRITERIA is the first here and in this case we are doing almost full scan on the table, if transactions has for example 1 mil recorods and 300k are encumbrances - so pg is doing full scan on 300k that is not performant.
Instead we have indexes already for encumbrance.sourcePurchaseOrderId and encumbrance.sourcePoLineId.
https://github.com/folio-org/mod-finance-storage/blob/master/src/main/resources/templates/db_scripts/schema.json#L419
So we should move this criteria to beginning of the condition and in this case appropriate index should be used and opening order time should be increased a lot

damien-git: [MODFISTO-466](https://folio-org.atlassian.net/browse/MODFISTO-466) would help with that, if we add an index for `transactionType` (which I think we should given how frequently we use it).

damien-git (approved):

src/main/java/org/folio/service/finance/transaction/EncumbranceService.java
SerhiiNosko: even if we add that index for transactionType I am not sure that index for encumbrance.sourcePurchaseOrderId and encumbrance.sourcePoLineId will be used. Because transactionType can split our initial dataset for 7 chunks because we have 7 possible allowed values for transaction type. So if we have 700k records - we need to do full scan on 100k records after applying index for transactionType .
Are you okay if I change order to paste ENCUMBRANCE_CRITERIA to the end?

imerabishvili: done

SerhiiNosko (approved):

PR #870, opened 2024-03-22
MODORDERS-1066. Item records are NOT deleted from Instance when Order is unopened and "Delete items" option is picked

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1066

### DIFF
--- a/src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
+++ b/src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
@@ -185,8 +185,8 @@ public class UnOpenCompositeOrderManager {
   }
 
   private Future<List<Piece>> deleteExpectedPieces(CompositePoLine compPOL, RequestContext requestContext) {
-    if (!PoLineCommonUtil.isItemsUpdateRequired(compPOL)) {
-      logger.info("Skip deleting pieces for Un-Open order line with id: {} because they were not created before", compPOL.getId());
+    if (PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) && Boolean.TRUE.equals(compPOL.getCheckinItems())) {
+      logger.info("Receipt is not required and independent receiving flow is used, skipping deleting pieces, poLineId: {}", compPOL.getId());
       return Future.succeededFuture(Collections.emptyList());
     }
     return pieceStorageService.getExpectedPiecesByLineId(compPOL.getId(), requestContext)
@@ -226,6 +226,8 @@ public class UnOpenCompositeOrderManager {
                     return null;
                   })
                   .onSuccess(v -> logger.debug("Pieces, Items, Holdings deleted after UnOpen order"))
+                  .onFailure(e -> logger.error("Pieces, Items, Holdings failed to be deleted after UnOpen order for order id: {}",
+                    compPOL.getId(), e))
                   .mapEmpty();
               }
               return Future.succeededFuture();
@@ -267,7 +269,7 @@ public class UnOpenCompositeOrderManager {
               return inventoryManager.deleteHoldingById(holdingId, true, requestContext)
                                      .map(v -> Pair.of(holdingId, effectiveLocationId));
             }
-            return Future.succeededFuture();
+            return Future.succeededFuture().map(v -> Pair.of(null, null));
           }));
       }
     });
@@ -364,13 +366,16 @@ public class UnOpenCompositeOrderManager {
   public Future<Void> deletePieceWithItem(String pieceId, RequestContext requestContext) {
     PieceDeletionHolder holder = new PieceDeletionHolder().withDeleteHolding(true);
     return pieceStorageService.getPieceById(pieceId, requestContext)
-      .onSuccess(holder::setPieceToDelete)
-      .compose(aVoid -> purchaseOrderLineService.getOrderLineById(holder.getPieceToDelete().getPoLineId(), requestContext)
-        .compose(poLine -> purchaseOrderStorageService.getPurchaseOrderById(poLine.getPurchaseOrderId(), requestContext)
-          .map(purchaseOrder -> {
-            holder.withOrderInformation(purchaseOrder, poLine);
-            return null;
-          })))
+      .map(piece-> {
+        holder.setPieceToDelete(piece);
+        return holder;
+      })
+      .compose(aHolder -> purchaseOrderLineService.getOrderLineById(holder.getPieceToDelete().getPoLineId(), requestContext))
+      .compose(poLine -> purchaseOrderStorageService.getPurchaseOrderById(poLine.getPurchaseOrderId(), requestContext)
+      .map(purchaseOrder -> {
+        holder.withOrderInformation(purchaseOrder, poLine);
+        return null;
+      }))
       .compose(purchaseOrder -> protectionService.isOperationRestricted(holder.getOriginPurchaseOrder().getAcqUnitIds(), DELETE, requestContext))
       .compose(vVoid -> canDeletePieceWithItem(holder.getPieceToDelete(), requestContext))
       .compose(aVoid -> pieceStorageService.deletePiece(pieceId, requestContext))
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -78,12 +78,12 @@ public Future<Void> process(CompositePurchaseOrder compPO, CompositePurchaseOrde
     return updateAndGetOrderWithLines(compPO, requestContext)
       .map(aVoid -> encumbranceWorkflowStrategyFactory.getStrategy(OrderWorkflowType.OPEN_TO_PENDING))
       .compose(strategy -> strategy.processEncumbrances(compPO, poFromStorage, requestContext))
+      .compose(ok -> processInventory(compPO.getCompositePoLines(), deleteHoldings, requestContext))
```
SerhiiNosko, line 81: changed order because making order as pending should go after all processing steps. For example without this change receiptStatus always was pending in processeInventory() method.

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -189,17 +185,18 @@ private Future<Void> processInventoryOnlyWithItems(CompositePoLine compPOL, Requ
...
-    if (!PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) && Boolean.FALSE.equals(compPOL.getCheckinItems())) {
-      return pieceStorageService.getExpectedPiecesByLineId(compPOL.getId(), requestContext)
-        .compose(pieceCollection -> {
-          if (isNotEmpty(pieceCollection.getPieces())) {
-            return pieceStorageService.deletePiecesByIds(pieceCollection.getPieces().stream().map(Piece::getId).collect(toList()), requestContext)
-                                .map(v -> pieceCollection.getPieces());
-          }
-          return Future.succeededFuture(Collections.emptyList());
-        });
+    if (!PoLineCommonUtil.isItemsUpdateRequired(compPOL)) {
```
SerhiiNosko, line 188: changed to more generic check here to be the same in other places.
Also for example in UI when creating POL with ReceiptStatus = 'Receipt not required status', the receiving workflow becomes Independent automatically, so these 2 checks are replaced to simplify the code.
Also before this change ReceiptStatus always was Pending because of the issue with order of operations described earlier.

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -169,10 +169,6 @@ private Future<Void> processInventoryOnlyWithHolding(CompositePoLine compPOL, Re
   }
 
   private Future<Void> processInventoryOnlyWithItems(CompositePoLine compPOL, RequestContext requestContext) {
-    if (!PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus())) {
-      logger.info("Skip deleting items and pieces for Un-Open order with id: {} because receipt can be required", compPOL.getId());
-      return Future.succeededFuture();
-    }
```
SerhiiNosko, line 175: Check of PoLineCommonUtil.isItemsUpdateRequired(compPOL) is made before this method invocation, so not needed here

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -169,10 +169,6 @@ private Future<Void> processInventoryOnlyWithHolding(CompositePoLine compPOL, Re
   }
 
   private Future<Void> processInventoryOnlyWithItems(CompositePoLine compPOL, RequestContext requestContext) {
-    if (!PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus())) {
```
imerabishvili, line 172: **Question**
Could you please explain removing this check?

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -189,17 +185,18 @@ private Future<Void> processInventoryOnlyWithItems(CompositePoLine compPOL, Requ
...
-    if (!PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) && Boolean.FALSE.equals(compPOL.getCheckinItems())) {
-      return pieceStorageService.getExpectedPiecesByLineId(compPOL.getId(), requestContext)
-        .compose(pieceCollection -> {
-          if (isNotEmpty(pieceCollection.getPieces())) {
-            return pieceStorageService.deletePiecesByIds(pieceCollection.getPieces().stream().map(Piece::getId).collect(toList()), requestContext)
-                                .map(v -> pieceCollection.getPieces());
-          }
-          return Future.succeededFuture(Collections.emptyList());
-        });
+    if (PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) && Boolean.TRUE.equals(compPOL.getCheckinItems())) {
```
imerabishvili, line 188: Good!

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
SerhiiNosko: it is conflicting check with the check that we have in deleteExpectedPieces, and after fixing to make it similar it is not makes sense to doube it with deleteExpectedPieces.
This check was added previosly in scope of some refactoring and it was wrong
Also please see https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/service/pieces/flows/strategies/ProcessInventoryStrategy.java#L56 - this check is used to create/update pieces, so for deleting them we also should have the similar. I doubted between OR or AND but remained AND as it was previously

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
damien-git, line 188: This change means that even if the receiving workflow is "Independent order and receipt quantity", pieces can be deleted when unopening an order. I don't see that change in logic reflected in the ticket or comments. Also the dialog mentions "Synchronized order and receipt quantity", so I imagine it could confuse users. Could you explain it ?

damien-git (changes requested):

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
damien-git: But it is not: the opposite of
`!PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) && Boolean.FALSE.equals(compPOL.getCheckinItems())`
is:
`PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) || Boolean.TRUE.equals(compPOL.getCheckinItems())`

SerhiiNosko: thanks for this catch, initial logic was a bit hard to read, fixing

SerhiiNosko: Now it is the same as for creating/updating piece as should be:
https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/service/pieces/flows/strategies/ProcessInventoryStrategy.java#L56

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -189,17 +185,18 @@ private Future<Void> processInventoryOnlyWithItems(CompositePoLine compPOL, Requ
...
-      return pieceStorageService.getExpectedPiecesByLineId(compPOL.getId(), requestContext)
-        .compose(pieceCollection -> {
-          if (isNotEmpty(pieceCollection.getPieces())) {
-            return pieceStorageService.deletePiecesByIds(pieceCollection.getPieces().stream().map(Piece::getId).collect(toList()), requestContext)
-                                .map(v -> pieceCollection.getPieces());
-          }
-          return Future.succeededFuture(Collections.emptyList());
-        });
+    if (PoLineCommonUtil.isReceiptNotRequired(compPOL.getReceiptStatus()) || Boolean.TRUE.equals(compPOL.getCheckinItems())) {
+      logger.info("Receipt is not required and independent receiving flow is used, skipping deleting pieces, poLineId: {}", compPOL.getId());
```
damien-git, line 189: it should be "or" also in the log

SerhiiNosko: fixed

damien-git (approved):

PR #874, opened 2024-03-28
[MODORDERS-1026] - Implemented API to execute mod-template-engine request

azizbekxm:
## Purpose

https://folio-org.atlassian.net/browse/MODORDERS-1026
## Approach

- Fetched routing-list by id.
- Fetching users by userIds of routing-list
- Creating template-request
- Executing request and forwarding response

### Overall process : https://excalidraw.com/#json=obkLNEZh-lTWEnCbqLqQN,Hfkp0e-FXO1fsBOFp8xR4w

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -86,13 +86,7 @@
             "orders.item.reopen"
           ],
           "modulePermissions": [
-            "modperms.orders.item.put",
-            "orders-storage.titles.item.get",
-            "orders-storage.configuration.prefixes.collection.get",
-            "orders-storage.configuration.suffixes.collection.get",
-            "finance.funds.item.get",
-            "finance.transactions.collection.get",
-            "finance.transactions.batch"
+            "modperms.orders.item.put"
           ]
         },
         {
@@ -267,6 +261,7 @@
             "organizations-storage.organizations.collection.get",
             "invoice.invoice-lines.collection.get",
             "invoice.invoice-lines.item.put",
+            "invoices.bypass-acquisition-units",
             "orders-storage.order-invoice-relationships.collection.get",
             "user-tenants.collection.get",
             "consortia.sharing-instances.item.post"
@@ -1475,7 +1470,7 @@
     {
       "permissionName": "orders.bypass-acquisition-units",
       "displayName": "Bypass acquisition units checks",
-      "description": "Backend internal permission to bypass acquisition units checks"
+      "description": "Backend internal permission to bypass order acquisition units checks"
     },
     {
       "permissionName": "orders.acquisitions-units-assignments.assign",
@@ -1774,6 +1769,8 @@
         "acquisitions-units-storage.units.collection.get",
         "acquisitions-units-storage.memberships.collection.get",
         "circulation.requests.collection.get",
+        "orders-storage.configuration.prefixes.collection.get",
+        "orders-storage.configuration.suffixes.collection.get",
         "orders-storage.purchase-orders.collection.get",
         "orders-storage.purchase-orders.item.put",
         "orders-storage.purchase-orders.item.get",
@@ -1796,10 +1793,12 @@
         "orders-storage.reporting-codes.item.put",
         "orders-storage.reporting-codes.item.delete",
         "orders-storage.titles.collection.get",
+        "orders-storage.titles.item.get",
         "orders-storage.order-invoice-relationships.collection.get",
         "configuration.entries.collection.get",
         "finance.funds.budget.item.get",
         "finance.funds.collection.get",
+        "finance.funds.item.get",
         "finance.budgets.collection.get",
         "finance.transactions.batch",
         "finance.transactions.collection.get",
@@ -1833,6 +1832,7 @@
         "inventory-storage.identifier-types.collection.get",
         "invoice.invoices.collection.get",
         "invoice.invoice-lines.collection.get",
+        "invoices.bypass-acquisition-units",
         "organizations-storage.organizations.collection.get",
         "organizations-storage.organizations.item.get",
         "isbn-utils.convert-to-13.get"
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1003,7 +1003,7 @@
       "version": "1.0",
       "handlers": [
         {
-          "methods": ["POST"],
+          "methods": ["GET"],
           "pathPattern": "/orders/routing-lists/{id}/template",
           "permissionsRequired": [
             "orders.routing-list-template.item.get"
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -155,6 +155,9 @@
           "permissionsRequired": [
             "orders.po-lines.collection.get"
           ],
+          "permissionsDesired": [
+            "orders.bypass-acquisition-units"
+          ],
           "modulePermissions": [
             "acquisitions-units-storage.units.collection.get",
             "acquisitions-units-storage.memberships.collection.get",
@@ -195,6 +198,9 @@
           "permissionsRequired": [
             "orders.po-lines.item.get"
           ],
+          "permissionsDesired": [
+            "orders.bypass-acquisition-units"
+          ],
           "modulePermissions": [
             "orders-storage.po-lines.item.get",
             "orders-storage.alerts.item.get",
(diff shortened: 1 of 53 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -998,6 +998,19 @@
...
+    {
+      "id": "orders.routing-list",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/routing-lists/{id}/template",
+          "permissionsRequired": [
+            "orders.routing-list-template.item.get"
+          ]
```
SerhiiNosko, line 1010: we should add module permission to be able to call mod-template-engine requests from non-admin user

descriptors/ModuleDescriptor-template.json
```diff
@@ -998,6 +998,19 @@
         }
       ]
     },
```
SerhiiNosko, line 1000: we should add template-engine as required interface

descriptors/ModuleDescriptor-template.json
```diff
@@ -1739,8 +1757,8 @@
         "orders.rollover.item.post",
         "orders.holding-summary.collection.get",
         "orders.acquisition-methods.all",
-        "orders.export-history.all"
-
+        "orders.export-history.all",
+        "orders.routing-list-template.item.get"
```
SerhiiNosko, line 1761: we should add this permission in routing-lists.all when this PR will be merged https://github.com/folio-org/mod-orders/pull/879

descriptors/ModuleDescriptor-template.json
SerhiiNosko: and users interface as well

src/main/java/org/folio/service/UserService.java
```diff
@@ -1,12 +1,35 @@
...
+    this.restClient = restClient;
+  }
+
   public static String getCurrentUserId(Map<String, String> okapiHeaders) {
     return okapiHeaders.get(OKAPI_USERID_HEADER);
   }
+
+  public Future<UserCollection> getUsersByIds(List<String> userIds, RequestContext requestContext) {
+    var requestEntry = new RequestEntry(USERS_ENDPOINT).withOffset(0).withLimit(Integer.MAX_VALUE)
+      .withQuery(convertIdsToCqlQuery(userIds, "id"));
```
SerhiiNosko, line 32: we should use this approach to fetch users by chunks, because GET request length is limited
https://github.com/folio-org/mod-invoice/pull/484/files#diff-25c18aacbd2be683de8e7ddf920dc739213b839acf416059c63813e54e5587e3R296

src/main/java/org/folio/service/RoutingListsService.java
```diff
@@ -0,0 +1,109 @@
...
+    return userCollection.getUsers().stream()
+      .map(UserCollection.User::getPersonal)
+      .filter(Objects::nonNull)
+      .map(personalData -> {
+          var userForContext = new TemplateProcessingRequest.User()
+            .withFirstName(personalData.getFirstName())
+            .withLastName(personalData.getLastName());
+          var addresses = personalData.getAddresses();
+          if (addresses != null && !addresses.isEmpty()){
+            userForContext.withRoutingAddress(addresses.get(0).getAddressLine1());
```
SerhiiNosko, line 91: we could not get addresses.get(0). Logic should be the following: in orders-storage we have settings API where user can select preferred address type: https://folio-org.atlassian.net/browse/UIREC-300
So here we need to iterate through addresses and choose that address where address type is matched with settings value.

src/main/java/org/folio/models/TemplateProcessingRequest.java
```diff
@@ -0,0 +1,156 @@
...
+      return this;
+    }
+
+    public Context withUsers(List<User> users) {
+      this.users = users;
+      return this;
+    }
+  }
+
+  public static class RoutingList {
```
SerhiiNosko, line 80: can we use RoutingList from acq-models, that we already merged?

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -434,6 +436,16 @@ CompositeOrderDynamicDataPopulateService combinedPopulateService(CompositeOrderR
     return new CombinedOrderDataPopulateService(compositeOrderRetrieveHolderBuilder, populateServices);
   }
 
+  @Bean
+  RoutingListsService routingListService(RestClient restClient, UserService userService) {
```
SerhiiNosko, line 440: propose to name service as RoutingListService, because most of services we have in singular
image

descriptors/ModuleDescriptor-template.json
azizbekxm: thanks, done

descriptors/ModuleDescriptor-template.json
azizbekxm: thanks, done

src/main/java/org/folio/service/UserService.java
azizbekxm: implemented

src/main/java/org/folio/service/RoutingListsService.java
azizbekxm: this part is on hold for now until it will be solved

src/main/java/org/folio/models/TemplateProcessingRequest.java
azizbekxm: used, thanks

src/main/java/org/folio/config/ApplicationConfig.java
azizbekxm: changed

descriptors/ModuleDescriptor-template.json
SerhiiNosko: forgot to mention, we also need to add module permission to get routing list from storage and to get users

descriptors/ModuleDescriptor-template.json
```diff
@@ -998,6 +998,22 @@
         }
       ]
     },
+    {
+      "id": "orders.routing-list",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/routing-lists/{id}/template",
```
SerhiiNosko, line 1007: It sounds like just getting template stored in mod-template-engine.
What do you think about POST /orders/routing-lists/{id}/process-template ?

azizbekxm: First I thought in that way, the next, since api is no longer doing modification operations, get make more sense to me. 

POST is not [idempotent ](https://restfulapi.net/idempotent-rest-apis/) method, but our api is doing work that is idempotenent. That is mean, multiple request have same effect and don't change the db records.

regarding api -> user shouldn't need to know what should be done to get template, so /template this name come to mind. more like hiding internar dependencies.

Overall, for me, api tell the user enter routing list id and get required template.

src/test/java/org/folio/rest/impl/TitlesApiTest.java
```diff
@@ -121,20 +118,20 @@ void testPostTitle() {
     assertThat(postTitleRq.getId(), nullValue());
 
     Title postTitleRs = verifyPostResponse(TITLES_ENDPOINT, JsonObject.mapFrom(postTitleRq).encode(),
-      prepareHeaders(EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10, X_OKAPI_USER_ID_WITH_ACQ_UNITS, ALL_DESIRED_PERMISSIONS_HEADER), APPLICATION_JSON, HttpStatus.HTTP_CREATED.toInt()).as(Title.class);
+      prepareHeaders(EXIST_CONFIG_X_OKAPI_TENANT_LIMIT_10, X_OKAPI_USER_ID_WITH_ACQ_UNITS, ALL_DESIRED_ACQ_PERMISSIONS_HEADER), APPLICATION_JSON, HttpStatus.HTTP_CREATED.toInt()).as(Title.class);
```
SerhiiNosko, line 121: why we need changes in Title Api tests?

azizbekxm: It is improvement, ⁠I read book that said the Boy Scout rule "Leave the campground cleaner than you found it" :)

SerhiiNosko (approved):

Saba-Zedginidze-EPAM (approved):

azizbekxm:
Tested on rancher:
image

PR #878, opened 2024-04-02
[MODORDERS-1078] Avoid onSuccess terminal operators in the middle of executing chain

SerhiiNosko:
## Purpose
The handler used in onSuccess was called async and the method was not waiting for it to return. But also the API says:
WARNING: this is a terminal operation. If several handlers are registered, there is no guarantee that they will be invoked in order of registration.

We need to identify places where usage of onSuccess was incorrect and replace to use of map or compose methods
https://folio-org.atlassian.net/browse/MODORDERS-1078

### DIFF
--- a/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
+++ b/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
@@ -53,7 +53,7 @@ public class PieceUpdateFlowInventoryManager {
   private Future<Void> nonPackagePoLineUpdateInventory(PieceUpdateHolder holder, RequestContext requestContext) {
     return nonPackageUpdateTitleWithInstance(holder, requestContext)
       .map(holder::withInstanceId)
-      .compose(instanceId -> handleHolding(holder, requestContext))
+      .compose(aHolder -> handleHolding(holder, requestContext))
       .compose(holdingId -> handleItem(holder, requestContext))
       .map(itemId -> {
         Optional.ofNullable(itemId).ifPresent(holder.getPieceToUpdate()::withItemId);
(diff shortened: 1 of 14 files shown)


src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
```diff
@@ -52,10 +52,13 @@ public Future<Void> processInventory(PieceUpdateHolder holder, RequestContext re
 
   private Future<Void> nonPackagePoLineUpdateInventory(PieceUpdateHolder holder, RequestContext requestContext) {
     return nonPackageUpdateTitleWithInstance(holder, requestContext)
-      .onSuccess(holder::withInstanceId)
+      .map(holder::withInstanceId)
       .compose(instanceId -> handleHolding(holder, requestContext))
```
damien-git, line 56: `withInstanceId` returns a `PieceUpdateHolder`

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
```diff
@@ -64,10 +67,13 @@ private Future<Void> nonPackagePoLineUpdateInventory(PieceUpdateHolder holder, R
   private Future<Void> packagePoLineUpdateInventory(PieceUpdateHolder holder, RequestContext requestContext) {
     return titlesService.getTitleById(holder.getPieceToUpdate().getTitleId(), requestContext)
       .compose(title -> packageUpdateTitleWithInstance(title, requestContext))
-      .onSuccess(title -> holder.withInstanceId(title.getInstanceId()))
+      .map(title -> holder.withInstanceId(title.getInstanceId()))
       .compose(title -> handleHolding(holder, requestContext))
```
damien-git, line 71: same here

damien-git (changes requested):

SerhiiNosko:
@damien-git naming has been improved per your suggestions, please re-review again

damien-git:
> @damien-git naming has been improved per your suggestions, please re-review again

Thanks. You missed the first one.

damien-git (approved):

yuntianhu (approved):

SerhiiNosko:
Karate tests passed after the changes:
image

PR #879, opened 2024-04-02
[MODORDERS-1006] Create Routing Lists API

Saba-Zedginidze-EPAM:
## Purpose
[https://folio-org.atlassian.net/browse/MODORDERS-1006](https://folio-org.atlassian.net/browse/MODORDERS-1006)
## Approach
- Add new API
- Implement a service for API
- Add validation logic for creating routing lists


src/main/java/org/folio/service/routinglists/validators/RoutingListValidatorUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import org.folio.rest.jaxrs.model.RoutingListCollection;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+public class RoutingListValidatorUtil {
+
+  public static List<Error> validateRoutingList(RoutingListCollection rListExisting, PoLine poLine) {
```
SerhiiNosko, line 15: thinking that better to include this code into service layer, it used only in single place, so new class is not needed

ramls/routing-lists.raml
```diff
@@ -1,47 +1,88 @@
 #%RAML 1.0
-title: "RoutingList"
+title: Routing Lists
```
SerhiiNosko, line 2: module descriptor file also should be updated, see comment here
https://github.com/folio-org/mod-orders/pull/874#discussion_r1549516070

src/main/java/org/folio/rest/core/exceptions/ErrorCodes.java
```diff
@@ -117,7 +117,10 @@ public enum ErrorCodes {
   CLAIMING_CONFIG_INVALID("claimingConfigInvalid", "Claiming interval should be set and greater than 0 if claiming is active"),
   TEMPLATE_NAME_ALREADY_EXISTS("templateNameNotUnique", "Template name already exists"),
   BARCODE_IS_NOT_UNIQUE("barcodeIsNotUnique", "The barcode already exists. The barcode must be unique"),
-  DELETE_WITH_EXPENDED_AMOUNT("deleteWithExpendedAmount", "Cannot delete en encumbrance with an expended amount");
+  DELETE_WITH_EXPENDED_AMOUNT("deleteWithExpendedAmount", "Cannot delete an encumbrance with an expended amount"),
+  INVALID_ROUTING_LIST_FOR_PO_LINE_FORMAT("invalidRoutingListForPoLineFormat", "Cannot create routing list for POL without 'Physical' or 'P/E Mix' order format"),
+  ROUTING_LIST_LIMIT_REACHED_FOR_PO_LINE("routingListLimitReachedForPoLine", "Cannot create routing list for POL as the associated lists' amount is not less than Physical copies"),
```
SerhiiNosko, line 122: small typo in "lists' amount", should be "list's amount"

src/main/java/org/folio/service/routinglists/RoutingListService.java
```diff
@@ -1,51 +1,120 @@
...
+  public Future<RoutingList> getRoutingList(String rListId, RequestContext requestContext) {
+    RequestEntry requestEntry = new RequestEntry(ROUTING_LIST_BY_ID_ENDPOINT).withId(rListId);
+    return restClient.get(requestEntry, RoutingList.class, requestContext);
+  }
+
+  public Future<Void> updateRoutingList(RoutingList routingList, RequestContext requestContext) {
+    try {
+      validateRoutingList(routingList, requestContext);
+    }
+    catch (HttpException e) {
```
SerhiiNosko, line 70: per code convention catch should be on the same level as closing bracket

src/main/java/org/folio/service/routinglists/RoutingListService.java
```diff
@@ -1,51 +1,120 @@
...
+    }
+    RequestEntry requestEntry = new RequestEntry(ROUTING_LIST_ENDPOINT);
+    return restClient.post(requestEntry, routingList, RoutingList.class, requestContext);
+  }
+
+  public Future<RoutingListCollection> getRoutingLists(int limit, int offset, String query, RequestContext requestContext) {
+    RequestEntry requestEntry = new RequestEntry(ROUTING_LIST_ENDPOINT)
+      .withQuery(query)
+      .withOffset(offset)
+      .withLimit(limit);
```
SerhiiNosko, line 96: would be more readable when preserving the order , if we have parameters limit, offset, then we should declare in the same order in RequestEntry builder or change order in method parameters

src/main/java/org/folio/service/routinglists/validators/RoutingListValidatorUtil.java
```diff
@@ -0,0 +1,39 @@
...
+    return errors.stream().map(ErrorCodes::toError).toList();
+  }
+
+  private static boolean isPoLineFormatValid(PoLine poLine) {
+    return poLine.getOrderFormat() == PoLine.OrderFormat.PHYSICAL_RESOURCE
+      || poLine.getOrderFormat() == PoLine.OrderFormat.P_E_MIX;
+  }
+
+  private static boolean isRoutingListsLimitReached(RoutingListCollection rListExisting, PoLine poLine) {
+    return getQuantityPhysicalTotal(poLine) <= rListExisting.getTotalRecords();
```
SerhiiNosko, line 32: to match the name of method comparision logic should be like this, do you agree?
```suggestion
    return rListExisting.getTotalRecords() >= getQuantityPhysicalTotal(poLine);
```

src/test/java/org/folio/rest/impl/RoutingListsApiTest.java
```diff
@@ -13,51 +51,56 @@
...
-import org.folio.rest.core.models.RequestContext;
-import org.folio.service.RoutingListService;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-import org.mockito.MockitoAnnotations;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Bean;
```
SerhiiNosko, line 47: you need to use the same import order as in existing files, please contact with @azizbekxm or @imerabishvili for providing settings file with standartize import order

src/main/java/org/folio/rest/core/exceptions/ErrorCodes.java
Saba-Zedginidze-EPAM: In this context lists are plural, and for plural words correct case would be lists' (list's mean only for a list) but it's not a big deal, can change if it would be a better option still.

src/main/java/org/folio/service/routinglists/RoutingListService.java
Saba-Zedginidze-EPAM: Got it

src/main/java/org/folio/rest/impl/RoutingListsAPI.java
```diff
@@ -1,22 +1,22 @@
...
 import org.folio.rest.jaxrs.resource.OrdersRoutingLists;
-import org.folio.service.RoutingListService;
+import org.folio.service.routinglists.RoutingListService;
 import org.folio.spring.SpringContextUtil;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import javax.ws.rs.core.Response;
+import java.util.Map;
+
+import static io.vertx.core.Future.succeededFuture;
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Thanks, I did not know that, updated my settings according to your screenshots

src/main/java/org/folio/service/routinglists/validators/RoutingListValidatorUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+
+import org.folio.rest.core.exceptions.ErrorCodes;
+import org.folio.rest.jaxrs.model.Error;
+import org.folio.rest.jaxrs.model.Location;
+import org.folio.rest.jaxrs.model.PoLine;
+import org.folio.rest.jaxrs.model.RoutingListCollection;
+
+public class RoutingListValidatorUtil {
+
```
SerhiiNosko, line 13: if it is utility class than private constructor can be created to avoid initialization from outer places. See examples of other utility classes

descriptors/ModuleDescriptor-template.json
```diff
@@ -1003,6 +1003,31 @@
...
       "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/routing-lists",
+          "permissionsRequired": ["orders-storage.routing-lists.collection.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/orders/routing-lists",
+          "permissionsRequired": ["orders-storage.routing-lists.item.post"]
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Okay I see now. I thought one would be needing additional perms to access different modules aswell.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1736,6 +1761,19 @@
       "displayName" : "orders routing-list-template item get",
       "description" : "Orders routing-list-template item get"
```
azizbekxm, line 1762: each new permission should have these similar description

descriptors/ModuleDescriptor-template.json
```diff
@@ -1003,6 +1003,36 @@
       "id": "orders.routing-list",
       "version": "1.0",
       "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/orders/routing-lists",
+            "permissionsRequired": ["orders.routing-lists.collection.get"],
+            "modulePermissions": ["orders-storage.routing-lists.collection.get"]
```
SerhiiNosko, line 1010: formatting issue with spaces

SerhiiNosko (approved):

azizbekxm (approved):

PR #880, opened 2024-04-03
MODORDERS-1078. Revert back rollover changes to start it asynchronously

SerhiiNosko:
## Purpose
Revert back change made in https://github.com/folio-org/mod-orders/pull/878 in OrderRolloverService because it was the desired behavior.
Rollover should start asynchronously and update status of completing by additional rest call to mod-finance. It is made in this way to avoid gateway timeout issues when mod-finance http request waits when the order rollover part completes.
Related ticket: https://folio-org.atlassian.net/browse/MODORDERS-865

### DIFF
--- a/src/main/java/org/folio/service/orders/OrderRolloverService.java
+++ b/src/main/java/org/folio/service/orders/OrderRolloverService.java
@@ -108,8 +108,8 @@ public class OrderRolloverService {
 
   public Future<Void> rollover(LedgerFiscalYearRollover ledgerFYRollover, RequestContext requestContext) {
     return prepareRollover(ledgerFYRollover, requestContext)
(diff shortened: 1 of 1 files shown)


Abdulkhakimov (approved):

azizbekxm (approved):

damien-git:
I think we should at least add a comment when we use `onSuccess` to start an async child, or maybe find a different way to do it. Otherwise we risk making the same change again.

SerhiiNosko:
Thanks @damien-git , added comment about it in currently opened PR: https://github.com/folio-org/mod-orders/pull/881
And thanks @Abdulkhakimov for catching this so important thing

PR #905, opened 2024-04-19
Delete received pieces in bulk

yuntianhu:
Purpose: Libraries that use the Receiving app for receipt of print serials may wish to delete pieces that precede a specific date. Currently, FOLIO supports the deletion of pieces, but one at a time. When a library wishes to purge receiving records for multiple years on an ongoing workflow, the ability to delete pieces in bulk will streamline this workflow. 

Weeding old journal issues; receiving records no longer necessary since items have been withdrawn

User story statement(s):

As a serials librarian,
I want to delete multiple pieces at one time
so that the continuous addition of pieces does not eventually impact performance of the Receiving app.

Scenarios:

Bulk edit received/expected pieces:

Given user has received multiple piece records for a given receiving title

When expanding received/expected pieces accordion

Then user has option in action menu for Bulk edit

Select one or more pieces:

Given user clicks Bulk edit

When toggles appear in table

Then user can select one or more pieces

Edit pieces:

Given user has selected one or more pieces

When user clicks edit piece button

Then Bulk edit piece form is displayed

Delete all pieces

Given user clicks edit piece button

When Bulk edit piece form is displayed

Then "Delete all" button is active

Delete all pieces confirmation

Given user is viewing Bulk edit piece form

When user clicks "Delete all"

Then confirmation modal is shown

Message: "This will permanently remove all the pieces you have selected. Any related items with a status of "On order" will also be deleted. Would you like to proceed?

AND user can confirm or cancel

Library enhancement suggested by Lafayette College (May 2022 adopter)

### DIFF
--- a/src/main/java/org/folio/rest/impl/PiecesAPI.java
+++ b/src/main/java/org/folio/rest/impl/PiecesAPI.java
@@ -2,10 +2,14 @@ package org.folio.rest.impl;
 
 import static io.vertx.core.Future.succeededFuture;
 
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.core.Response;
 
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -96,4 +100,26 @@ public class PiecesAPI extends BaseApi implements OrdersPieces {
       .onSuccess(ok -> asyncResultHandler.handle(succeededFuture(buildNoContentResponse())))
       .onFailure(fail -> handleErrorResponse(asyncResultHandler, fail));
   }
+
+  @Override
+  @Validate
+  public void deleteOrdersPiecesBulk(List<String> ids, boolean deleteHolding, Map<String, String> okapiHeaders,
+                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    List<Future> deleteFutures = ids.stream()
+      .map(id -> pieceDeleteFlowManager.deletePiece(id, deleteHolding, new RequestContext(vertxContext, okapiHeaders))
+        .onFailure(t -> {
+          logger.error("Failed to delete piece with ID: " + id, t);
+        })
+        .mapEmpty())
+      .collect(Collectors.toList());
+
+    CompositeFuture.join(deleteFutures).onComplete(ar -> {
+      if (ar.succeeded()) {
+        asyncResultHandler.handle(Future.succeededFuture(buildNoContentResponse()));
+      } else {
+        handleErrorResponse(asyncResultHandler, new Throwable("Error deleting multiple pieces"));
+      }
+    });
+  }
 }
+


src/main/java/org/folio/rest/impl/PiecesAPI.java
```diff
@@ -91,9 +99,21 @@ public void putOrdersPiecesById(String pieceId, boolean createItem, boolean dele
...
+
+  }
+
+  @Override
+  @Validate
+  public void deleteOrdersPiecesBatch(PieceCollection entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    List<String> pieceIds = new ArrayList<>();
+    for (Piece piece : entity.getPieces()) {
+      pieceIds.add(piece.getId());
+    }
```
azizbekxm, line 115: I believe we can do by using steam instead of for loop. that will be consice

src/main/java/org/folio/rest/impl/PiecesAPI.java
```diff
@@ -66,6 +72,8 @@ public void postOrdersPieces(boolean createItem, Piece entity, Map<String, Strin
       .onFailure(t -> handleErrorResponse(asyncResultHandler, t));
   }
 
+
+
```
azizbekxm, line 76: please reformat this file one more time

src/main/java/org/folio/rest/impl/PiecesAPI.java
yuntianhu: good catch! will do

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -140,4 +143,26 @@ private Future<JsonObject> getOnOrderItemForPiece(Piece piece, RequestContext re
       return Future.succeededFuture();
     }
   }
+
+ public Future<Void> batchDeletePiece(List<String> ids, RequestContext requestContext) {
+   PieceDeletionHolder holder = new PieceDeletionHolder().withDeleteHolding(false);
+
+   List<Future> deleteFutures = ids.stream()
+     .map(id -> pieceStorageService.getPieceById(id, requestContext)
```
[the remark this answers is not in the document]
yuntianhu: I will try it. I see there is a max number of get pieces in  batch is 15. I am not sure why. but, I will try it first, and see how it goes.

src/main/java/org/folio/rest/impl/PiecesAPI.java
yuntianhu: good catch, will do!

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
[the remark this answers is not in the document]
SerhiiNosko: it is some constant to prevent long url query size, the max size of url length - 4 kb, historically we use chunks by 15 elems to avoid this issue.

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -140,4 +143,26 @@ private Future<JsonObject> getOnOrderItemForPiece(Piece piece, RequestContext re
...
+       .map(pieceToDelete -> {
+         holder.withPieceToDelete(pieceToDelete);
+         return null;
+       })
+       .compose(aHolder -> basePieceFlowHolderBuilder.updateHolderWithOrderInformation(holder, requestContext))
```
SerhiiNosko, line 156: The same logic also is in deletePiece method. We should avoid code duplication, logic should be in 1 places

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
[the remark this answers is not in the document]
yuntianhu: Thank you for the info, Serhii

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
yuntianhu: thank you Serhii, I will rewrite this part

PR #909, opened 2024-04-22
[MODORDERS-1081] Create instance when user adds receiving title to package order line

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1081] Create instance when user adds receiving title to package order line](https://folio-org.atlassian.net/browse/MODORDERS-1081)

## Approach
* Prevent creating shared instances upon title creation
* Create inventory instances upon title creation only in central tenant
* Create shared instances upon Piece creation/receiving


Saba-Zedginidze-EPAM:
Will be fixing unit tests soon, open for review

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -168,10 +168,8 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
   public Future<Void> openOrderUpdateInventory(CompositePoLine compPOL, Piece piece, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
     if (Boolean.TRUE.equals(compPOL.getIsPackage())) {
       return titlesService.getTitleById(piece.getTitleId(), requestContext)
-        .compose(title -> inventoryInstanceManager.openOrderHandlePackageLineInstance(title, isInstanceMatchingDisabled, requestContext))
```
SerhiiNosko, line 171: If this method openOrderHandlePackageLineInstance() is not used in inventoryInstanceManager it can be deleted

src/main/java/org/folio/service/titles/TitlesInstanceService.java
```diff
@@ -0,0 +1,43 @@
+package org.folio.service.titles;
+
+import org.folio.rest.core.models.RequestContext;
+import org.folio.rest.jaxrs.model.Title;
+import org.folio.rest.tools.utils.TenantTool;
+import org.folio.service.inventory.InventoryInstanceManager;
+
+import io.vertx.core.Future;
+
+public class TitlesInstanceService {
```
SerhiiNosko, line 10: propose to rename to TitleInstanceService, it is one to one association

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -48,12 +48,11 @@ public TitlesService(RestClient restClient, ProtectionService protectionService,
   }
 
   public Future<Title> createTitle(Title title, RequestContext requestContext) {
-    String targetTenantId = TenantTool.tenantId(requestContext.getHeaders());
     return protectionService.validateAcqUnitsOnCreate(title.getAcqUnitIds(), TITLES_ASSIGN, requestContext)
-      .compose(v -> inventoryInstanceManager.createShadowInstanceIfNeeded(title.getInstanceId(), targetTenantId, requestContext))
-      .compose(shadowInstance -> {
+      .compose(v -> inventoryInstanceManager.getOrCreateInstanceRecord(title, requestContext))
+      .compose(instId -> {
```
SerhiiNosko, line 53: I think we need the following logic here:

- firstly invoke createShadowInstanceIfNeeded(). In this case when we are in ECS and shared instance is exists in central tenant - we would not need to create instance with another id. Instead we will use shadow instance with the same instanceId as in central tenant 
- after this invoke inventoryInstanceManager.getOrCreateInstanceRecord() only in case when shadow instance from previous step was not created.
So for ECS envs we will use shadow instance, for non ECS - create new instance that is our desired result.
We can encapsulate this logic in TitleInstanceService

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -73,11 +72,15 @@ public Future<Title> getTitleById(String titleId, RequestContext requestContext)
   }
 
   public Future<Void> saveTitle(Title title, RequestContext requestContext) {
-    String targetTenantId = TenantTool.tenantId(requestContext.getHeaders());
-    return inventoryInstanceManager.createShadowInstanceIfNeeded(title.getInstanceId(), targetTenantId, requestContext)
-      .compose(shadowInstance -> {
+    return saveTitleWithInstance(title, false, requestContext).mapEmpty();
```
SerhiiNosko, line 75: here should be the same logic, as in previous create() method, firstly invoke createShadowInstanceIfNeeded - if instance has not been created because non ECS mode - create instance with inventoryInstanceManager.getOrCreateInstanceRecord

src/main/java/org/folio/service/titles/TitleInstanceService.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  private Future<String> createTitleInventoryInstance(Title title, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
+    if (title.getInstanceId() != null) {
+      return Future.succeededFuture(title.getInstanceId());
+    }
+    return titlesService.saveTitleWithInstance(title, isInstanceMatchingDisabled, requestContext);
+  }
+
+  private Future<String> createTitleShadowInstance(String instanceId, RequestContext requestContext) {
```
SerhiiNosko, line 36: this method create only shadow instance, name createTitleShadowInstance a bit confuses, because title is not created here, maybe it can be just removed

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Removed it, forgot to remove after rebasing

src/main/java/org/folio/service/titles/TitlesInstanceService.java
Saba-Zedginidze-EPAM: Renamed

src/main/java/org/folio/service/titles/TitlesService.java
Saba-Zedginidze-EPAM: Modified the logic

src/main/java/org/folio/service/titles/TitleInstanceService.java
Saba-Zedginidze-EPAM: Renamed all methods

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -168,10 +164,8 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
```
SerhiiNosko, line 171: Just realized that this method openOrderUpdateInventory() is used for non-package POLs. We don't need any changes in flow for non-package orders, so please revert changes in this class.

When adding title to Package POL TitleService is invoked so we need change only in this service for package orders

src/main/java/org/folio/service/titles/TitleInstanceService.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public Future<String> getOrCreateInstance(Title title, RequestContext requestContext) {
+    return getOrCreateInstance(title, false, requestContext);
+  }
+
+  public Future<String> getOrCreateInstance(Title title, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
+    return createShadowInstance(title.getInstanceId(), requestContext)
+      .compose(shadowInstId -> createInventoryInstance(shadowInstId, title, isInstanceMatchingDisabled, requestContext));
+  }
```
azizbekxm, line 25: I assume true case can be used in the future, because all places is using false case and increase code chain.

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
SerhiiNosko: if we revert this change shadow instance can not be used in this place and new instance will be created that is not desired behaviour.
So you fixed here issue when createShadowInstanceIfNeeded not invoked

src/main/java/org/folio/service/titles/TitleInstanceService.java
Saba-Zedginidze-EPAM: @azizbekxm so do I get this right - you are suggesting passing the flag with every call and remove the method with false as default?

azizbekxm: maybe action not required. I am going to say is that If true isn't used in the future, it would be better to simply methods by removing boolean attributes and just calling directly with false.

azizbekxm (approved):

src/main/java/org/folio/service/titles/TitleInstanceService.java
```diff
@@ -0,0 +1,42 @@
...
+      return Future.succeededFuture(title.getInstanceId());
+    }
+    return inventoryInstanceManager.getOrCreateInstanceRecord(title, isInstanceMatchingDisabled, requestContext);
+  }
+
+  private Future<String> createShadowInstance(String instanceId, RequestContext requestContext) {
+    String targetTenantId = TenantTool.tenantId(requestContext.getHeaders());
+    return inventoryInstanceManager.createShadowInstanceIfNeeded(instanceId, targetTenantId, requestContext)
+      .map(sharingInstance -> sharingInstance != null ? instanceId : null);
+  }
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Moved it

SerhiiNosko (approved):

PR #929, opened 2024-05-01
[MODORDERS-1104] Integrate with Transfer API for moving circulation requests to new item

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1104] Integrate with Transfer API for moving circulation requests to new item](https://folio-org.atlassian.net/browse/MODORDERS-1104)

## Approach
* Create InventoryBindingManager to move or cancel requests for specified piece items
* Check which pieces have an outstanding Request


src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -113,17 +115,40 @@ public Future<JsonObject> getItemRecordById(String itemId, boolean skipThrowNorF
...
+      .map(json -> json.getInteger(REQUESTS_TOTAL));
+  }
+
+  /**
+   * Returns list of requestIds for specified item.
+   *
+   * @param itemId id of Item
+   * @return future with list of requestIds
+   */
+  public Future<List<String>> getRequestIdsByItemId(String itemId, RequestContext requestContext) {
```
SerhiiNosko, line 137: let's move these 2 methods connected to Requests into separate service:

- getNumberOfRequestsByItemId

- getRequestIdsByItemId

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
```diff
@@ -0,0 +1,98 @@
...
+        JsonObject jsonObject =  JsonObject.of(REQUEST_JSON_KEY, itemId);
+        RequestEntry requestEntry = new RequestEntry(String.format(REQUEST_MOVE_ENDPOINT, reqId));
+        return restClient.postJsonObject(requestEntry, jsonObject, requestContext);
+      })
+      .mapEmpty();
+
+  }
+
+  private Future<String> createNewItemForRequest(String requestId, RequestContext requestContext) {
+    throw new RuntimeException("Not implemented");
```
SerhiiNosko, line 95: Item would be created earlier in the flow, @azizbekxm is working on it, so in transferRequest() we can add param itemId

src/main/java/org/folio/service/inventory/InventoryItemManager.java
Saba-Zedginidze-EPAM: Created it

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  public Future<List<Piece>> getPiecesWithActiveRequests(List<Piece> pieces, RequestContext requestContext) {
+    if (logger.isDebugEnabled()) {
+      logger.debug("Filtering pieces with active requests: {}", pieces.stream().map(Piece::getId).toList());
+    }
+    var futures = pieces.stream()
+      .map(piece -> getPieceWithRequests(piece, requestContext))
+      .collect(Collectors.toList());
+    return GenericCompositeFuture.all(futures)
```
SerhiiNosko, line 48: here we can try optimize, because now if we have 10 pieces, we will do 10 calls.
We can introduce new method in CirculationRequestsRetriever to fetch requests by multiple item ids. Interesting do we have in response itemId to map number of requests to itemId.
To avoid very long GET queries we use pattern to split by 15 elements, here is releated constant - MAX_IDS_FOR_GET_RQ_15

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private final CirculationRequestsRetriever circulationRequestsRetriever;
+  private final RestClient restClient;
+
+  public InventoryBindingManager(RestClient restClient, CirculationRequestsRetriever circulationRequestsRetriever) {
+    this.restClient = restClient;
+    this.circulationRequestsRetriever = circulationRequestsRetriever;
+  }
+
+  public Future<List<Piece>> getPiecesWithActiveRequests(List<Piece> pieces, RequestContext requestContext) {
```
SerhiiNosko, line 41: Propose to not mix pieces, items, and requests in the same service.
Instead of pieces we can use itemIds.

The caller side will extract itemIds from pieces. In this case we can use this service only in other further places that do not know about pieces, only about items.

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.folio.rest.core.RestClient;
+import org.folio.rest.core.models.RequestContext;
+import org.folio.rest.core.models.RequestEntry;
+import org.folio.rest.jaxrs.model.Piece;
+import org.folio.service.CirculationRequestsRetriever;
+
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+public class InventoryBindingManager {
```
SerhiiNosko, line 23: Propose to rename to InventoryItemRequestManager

src/main/java/org/folio/service/CirculationRequestsRetriever.java
```diff
@@ -0,0 +1,64 @@
...
+import org.folio.rest.core.models.RequestContext;
+import org.folio.rest.core.models.RequestEntry;
+
+import io.vertx.core.Future;
+
+public class CirculationRequestsRetriever {
+
+  private static final String REQUESTS_RECORDS = "requests";
+  private static final String REQUESTS_TOTAL = "totalRecords";
+  private static final String REQUEST_ID = "id";
```
azizbekxm, line 19: Also consider replacing with Enum
In Effective java suggest to use Enums over this type of patter.
https://medium.com/@swang354/effective-java-7357d032dd75
it is up to you. If you are going to leave as it was write some reasons to explain situation.

src/main/java/org/folio/service/CirculationRequestsRetriever.java
```diff
@@ -0,0 +1,64 @@
...
+  public CirculationRequestsRetriever(RestClient restClient) {
+    this.restClient = restClient;
+  }
+
+  /**
+   * Returns number of requests for specified item.
+   *
+   * @param itemId id of Item
+   * @return future with number of requests
+   */
```
azizbekxm, line 32: These comments seems unhelpful, make it more informative or remove
reference to learn about comments: https://medium.com/codex/clean-code-comments-833e11a706dc
Noise Comments Section
Or Clean Code comments chapter

src/main/java/org/folio/service/inventory/InventoryItemRequestManager.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class InventoryItemRequestManager {
+
+  private static final Logger logger = LogManager.getLogger(InventoryItemRequestManager.class);
+
+  private static final String REQUEST_CANCEL_MESSAGE = "Cancelling Request with id='{}'";
+  private static final String REQUEST_MOVE_MESSAGE = "Moving Request with id='{}' to item with id='{}'";
+  private static final String REQUEST_ENDPOINT = INVENTORY_LOOKUP_ENDPOINTS.get(REQUESTS) + "/%s";
+  private static final String REQUEST_MOVE_ENDPOINT = REQUEST_ENDPOINT + "/move";
+  private static final String REQUEST_JSON_KEY = "destinationItemId";
```
azizbekxm, line 30: https://github.com/folio-org/mod-orders/pull/929#discussion_r1587758084

src/main/java/org/folio/service/inventory/InventoryItemRequestManager.java
```diff
@@ -0,0 +1,89 @@
...
+  private final RestClient restClient;
+
+  public InventoryItemRequestManager(RestClient restClient, CirculationRequestsRetriever circulationRequestsRetriever) {
+    this.restClient = restClient;
+    this.circulationRequestsRetriever = circulationRequestsRetriever;
+  }
+
+  public Future<List<String>> getItemsWithActiveRequests(List<String> itemIds, RequestContext requestContext) {
+    if (logger.isDebugEnabled()) {
+      logger.debug("Filtering itemIds with active requests: {}", itemIds);
```
azizbekxm, line 42: it would be great start message with method name
 getItemsWithActiveRequests:: Filtering itemIds with active requests: {}", itemIds);
Comment
The same goes to other places as well

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
Saba-Zedginidze-EPAM: Modified it

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
Saba-Zedginidze-EPAM: I believe now it should work correctly the way you explained

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
Saba-Zedginidze-EPAM: Renamed

src/main/java/org/folio/service/inventory/InventoryBindingManager.java
Saba-Zedginidze-EPAM: Removed pieces altogether and moved requests related functionality in a different service

src/main/java/org/folio/service/CirculationRequestsRetriever.java
Saba-Zedginidze-EPAM: Yes you are correct, the method is very straightforward, and its name speaks for it, removed unnecessary comments

src/main/java/org/folio/service/inventory/InventoryItemRequestManager.java
Saba-Zedginidze-EPAM: Created a separate enum for such fields, good point

src/main/java/org/folio/service/inventory/InventoryItemRequestManager.java
Saba-Zedginidze-EPAM: Added it

imerabishvili (approved):

src/main/java/org/folio/service/inventory/InventoryItemRequestService.java
```diff
@@ -0,0 +1,82 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+public class InventoryItemRequestService {
+
+  private static final Logger logger = LogManager.getLogger(InventoryItemRequestService.class);
+
+  private static final String CANCEL_MESSAGE = "cancelRequest:: Cancelling Request with id='{}'";
+  private static final String MOVE_MESSAGE = "transferRequest:: Moving Request with id='{}' to item with id='{}'";
+  private static final String FILTER_ITEMS_MESSAGE = "getItemsWithActiveRequests:: Filtering itemIds with active requests: {}";
```
azizbekxm, line 28: these three messages used only once, so I propose to move back to logs

src/test/java/org/folio/service/CirculationRequestsRetrieverTest.java
```diff
@@ -0,0 +1,186 @@
+package org.folio.service;
+
+import static org.folio.TestConfig.autowireDependencies;
```
azizbekxm, line 3: I didn't want to confuse you, but we should move to Intellij idea default import style. The old files it is better to stick old style, for the new files, it is better to move to new style. I saw many pr, where developer are using default java import style

just open settings and change to Project that will respect to existing import style if there is, otherwise it will use default import style

SerhiiNosko (approved):

src/main/java/org/folio/service/CirculationRequestsRetriever.java
```diff
@@ -0,0 +1,76 @@
...
+public class CirculationRequestsRetriever {
+
+  private final RestClient restClient;
+
+  public CirculationRequestsRetriever(RestClient restClient) {
+    this.restClient = restClient;
+  }
+
+  public Future<Integer> getNumberOfRequestsByItemId(String itemId, RequestContext requestContext) {
+    String query = String.format("(itemId==%s and status=\"*\")", itemId);
```
SerhiiNosko, line 35: regarding status filters, we will ask on refinement do we filter by open ones only.
Here is list of all available statuses:
"Open - Not yet filled",
"Open - Awaiting pickup",
"Open - In transit",
"Open - Awaiting delivery",
"Closed - Filled",
"Closed - Cancelled",
"Closed - Unfilled",
"Closed - Pickup expired"

Saba-Zedginidze-EPAM: Tested and works for each Open status now

src/test/java/org/folio/service/CirculationRequestsRetrieverTest.java
Saba-Zedginidze-EPAM: Got it, thanks, I believe it should be correct now

src/main/java/org/folio/service/inventory/InventoryItemRequestService.java
Saba-Zedginidze-EPAM: Moved them, thanks

PR #930, opened 2024-05-01
Modorders 1103

imerabishvili:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1103

## Approach
Update `inventoryHoldingManager` and `inventoryItemManager` usages to use tenants from PO line locations.

## Pre-Merge Checklist:


src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -93,69 +93,80 @@ public InventoryItemManager(RestClient restClient,
   }
 
 
-  /**
-   * Returns list of item records for specified id's.
-   *
-   * @param ids List of item id's
-   * @return future with list of item records
-   */
```
SerhiiNosko, line 101: why java docs deleted, you you want to extract interfaces and move java docs to interfaces?

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -250,6 +249,40 @@ public Future<List<Piece>> handleItemRecords(CompositePoLine compPOL, Location l
       });
```
SerhiiNosko, line 249: handleItemRecords() does not have usages, so can be deleted

src/main/java/org/folio/helper/PoNumberHelper.java
```diff
@@ -42,6 +48,20 @@ public Future<Void> checkPONumberUnique(PoNumber poNumber, RequestContext reques
...
+      .map(po -> {
+        if (po.getInteger("totalRecords") != 0) {
+          logger.error("Exception validating PO Number existence");
+          throw new HttpException(400, ErrorCodes.PO_NUMBER_ALREADY_EXISTS);
+        }
+        return null;
+      });
+  }
+
   public Future<PoNumber> getPoNumber(RequestContext requestContext) {
```
SerhiiNosko, line 65: Is not used, can be deleted

src/main/java/org/folio/orders/utils/RequestContextUtil.java
```diff
@@ -10,6 +10,9 @@ private RequestContextUtil() {
   }
 
   public static RequestContext createContextWithNewTenantId(RequestContext requestContext, String tenantId) {
+    if (tenantId == null) {
```
SerhiiNosko, line 13: better to use StringUtils

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -93,69 +93,80 @@ public InventoryItemManager(RestClient restClient,
...
     return restClient.getAsJsonObject(requestEntry, skipThrowNorFoundException, requestContext);
   }
 
-  /**
-   * Returns list of requests for specified item.
-   *
-   * @param itemId id of Item
-   * @return future with list of requests
-   */
-  public Future<Integer> getNumberOfRequestsByItemId(String itemId, RequestContext requestContext) {
```
SerhiiNosko, line 121: Requests functionality will be moved outside of InventoryItemManager in scope of https://github.com/folio-org/mod-orders/pull/929

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -144,25 +160,26 @@ private Future<Void> processInventory(CompositePoLine compPOL, boolean deleteHol
...
-      } else { // synchronized workflow
-        return (deleteHoldings) ? processInventoryHoldingWithItems(compPOL, requestContext) : processInventoryOnlyWithItems(compPOL, requestContext);
-      }
+    boolean itemsUpdateRequired = PoLineCommonUtil.isItemsUpdateRequired(compPOL);
+    boolean holdingsUpdateRequired = PoLineCommonUtil.isHoldingsUpdateRequired(compPOL);
+    // itemsUpdateRequired => holdingsUpdateRequired
+    if (!itemsUpdateRequired && !holdingsUpdateRequired) {
+      return Future.succeededFuture();
     }
-    if (PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)) {
```
SerhiiNosko, line 154: we need to check holdingsUpdateRequired here, correct?

SerhiiNosko: seems previous version of this method is more readable

src/main/java/org/folio/service/inventory/InventoryItemManager.java
imerabishvili: Yes now I see it can be deleted however for `InventoryHoldingManager::getHoldingsForAllLocationTenants` it's used only in tests but we should refactor unit test and then remove it.

src/main/java/org/folio/helper/PoNumberHelper.java
imerabishvili: `PoNumberAPI::getOrdersPoNumber` uses it. Can you double check?

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -93,69 +93,80 @@ public InventoryItemManager(RestClient restClient,
...
       .withQuery(query).withOffset(0).withLimit(Integer.MAX_VALUE);
     return restClient.getAsJsonObject(requestEntry, requestContext)
-      .map(response -> extractEntities(response, ITEMS));
+      .map(this::extractEntities);
+  }
+
+  public Future<Integer> getNumberOfRequestsByItemId(String itemId, RequestContext requestContext) {
+    String query = String.format("(itemId==%s and status=\"*\")", itemId);
+    RequestEntry requestEntry = new RequestEntry(INVENTORY_LOOKUP_ENDPOINTS.get(REQUESTS))
+      .withQuery(query).withOffset(0).withLimit(0);
```
azizbekxm, line 138: I think Limit is supposed to be Integer Max Value

imerabishvili: I asked this as well. And why current version is working (I didn't modified, just replaced)?

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
imerabishvili: Agree. I'll try to farther improve or rollback to previous version.

src/main/java/org/folio/service/inventory/InventoryItemManager.java
SerhiiNosko: @Saba-Zedginidze-EPAM please take a look is it expected that we use limit = 0 here as you cyrrently busy with requests integration

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
SerhiiNosko: I see you back to previous implementation, resolvign comment

SerhiiNosko (approved):

src/main/java/org/folio/orders/utils/RequestContextUtil.java
SerhiiNosko: RMB is written just with null check, so I am okay to leave as is also with null check. It's personally my preference to use StringUtils in every place, but we can leave as is for current place.
image

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -184,44 +201,81 @@ private Future<Void> processInventoryOnlyWithHolding(CompositePoLine compPOL, Re
...
-          .compose(onOrderItems -> {
-            if (CollectionUtils.isNotEmpty(onOrderItems)) {
-              List<String> itemIds = onOrderItems.stream().map(item -> item.getString(ID)).collect(toList());
-              return inventoryItemManager.deleteItems(itemIds, false, requestContext);
-            }
-            return Future.succeededFuture();
-          })
-          .onSuccess(v -> logger.info("Items and pieces deleted after Un-Open order with id: {}", compPOL.getId()))
-          .onFailure(e -> logger.error("Items and pieces deletion failed after Un-Open order with id: {}", compPOL.getId(), e))
+        GenericCompositeFuture.all(
```
SerhiiNosko, line 240: found one more thing, previous implementation was waiting when item deleted and only after that returned completed future, I think we need wait also to keep the same bahaviour, so need to add return before GenericCompositeFuture

SerhiiNosko: If we have 60 tenants and need update inventory objects in every of them, it can slow down process of un-opening the order, so probably we should remain as is with ignoring this comment, adding also @okolawole-ebsco to this discussion

SerhiiNosko: lets start with approach to wait until all items updated to not mask errors that can be produced by inventory, and after discussion will back to this and change if needed

src/main/java/org/folio/service/inventory/InventoryItemManager.java
Saba-Zedginidze-EPAM: As I know, this results in the record bodies being empty but the totalRecords shows how many entities were originally selected. I noticed this issue in my PR aswell and am currently modifying the code

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -250,6 +248,40 @@ public Future<List<Piece>> handleItemRecords(CompositePoLine compPOL, Location l
...
   }
 
+  private Future<List<JsonObject>> searchStorageExistingItems(String poLineId, String holdingId, int expectedQuantity,
+                                                              RequestContext requestContext) {
+    String query = String.format(LOOKUP_ITEM_QUERY, poLineId, holdingId);
+    RequestEntry requestEntry = new RequestEntry(ITEM_STOR_ENDPOINT).withQuery(query).withOffset(0).withLimit(expectedQuantity);
+    return restClient.getAsJsonObject(requestEntry, requestContext)
+      .map(itemsCollection -> {
+        List<JsonObject> items = extractEntities(itemsCollection);
+        logger.debug("{} existing items found out of {} for PO Line with '{}' id", items.size(), expectedQuantity, poLineId);
```
azizbekxm, line 258: it would be great to start with method name like in other paces

azizbekxm: I saw other places, please use common style other places as well

src/main/java/org/folio/orders/utils/PoLineCommonUtil.java
```diff
@@ -191,6 +203,14 @@ public static Map<String, List<Location>> groupLocationsByHoldingId(CompositePoL
       .collect(Collectors.groupingBy(Location::getHoldingId));
   }
 
+  public static List<String> getLocationTenants(CompositePoLine poLine) {
```
azizbekxm, line 206: Propose to rename to getTenantsFromPoLineLocations or other you can come up with. Current nime like returining Pair locations and tenants

src/main/java/org/folio/service/inventory/InventoryItemManager.java
imerabishvili: Please note that first commit in the PR is `move private methods with one usage below their callers`.
Although I agree with your suggestion we can't fix logging and other usages in all replaced code parts. From this perspective please review only consecutive commits

src/main/java/org/folio/orders/utils/PoLineCommonUtil.java
imerabishvili: The method name should be reading like `getOrderLinePieces`: OrderLine is an object and it contains pieces. Same way Location is an object and it contains tenant. WDYT?

imerabishvili: Maybe `getTenantsFromLocations` as `CompositePoLine` is a method parameter?

azizbekxm: Your explanation make sense, you can leave as it was or change to other name specific comments. When invoking method considering it is static method, reading getLocationTenants kind of strange, so getPoLinLocationTenants make descriptive, it is suggest that it will get poLine object. It is up to you. Both version seems good to me.

src/main/java/org/folio/service/inventory/InventoryItemManager.java
azizbekxm: Ok,

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -184,44 +201,81 @@ private Future<Void> processInventoryOnlyWithHolding(CompositePoLine compPOL, Re
...
+          PoLineCommonUtil.getTenantsFromLocations(compPOL)
```
SerhiiNosko, line 241: if we are working in non ECS env - Location object will not have tenantId populated at all, it means we will not invoke this logic to process inventory

imerabishvili: `getTenantsFromLocations` list might contain `null` and `createContextWithNewTenantId` will return same context:
```
    if (tenantId == null) {
      return requestContext;
    }
```
That's why we can't use here isEmpty check. Tenant Id list should not contain empty string, only `null` allowed

SerhiiNosko: pushed the change, resolving

PR #931, opened 2024-05-03
MODORDERS-1105. Update logic to handle item status "On order" "Order closed" in correct tenant

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1105

### DIFF
--- a/src/main/java/org/folio/orders/utils/RequestContextUtil.java
+++ b/src/main/java/org/folio/orders/utils/RequestContextUtil.java
@@ -18,7 +18,7 @@ public class RequestContextUtil {
     }
     var modifiedHeaders = new CaseInsensitiveMap<>(requestContext.getHeaders());
     modifiedHeaders.put(XOkapiHeaders.TENANT, tenantId);
-    logger.info("Request context has been changes with new tenant: {}", tenantId);
+    logger.info("Request context has been changed with new tenant: {}", tenantId);
     return new RequestContext(requestContext.getContext(), modifiedHeaders);
   }
 
--- a/src/main/java/org/folio/orders/utils/RequestContextUtil.java
+++ b/src/main/java/org/folio/orders/utils/RequestContextUtil.java
@@ -1,10 +1,14 @@
 package org.folio.orders.utils;
 
 import org.apache.commons.collections4.map.CaseInsensitiveMap;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.helper.PurchaseOrderHelper;
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.rest.core.models.RequestContext;
 
 public class RequestContextUtil {
+  private static final Logger logger = LogManager.getLogger(PurchaseOrderHelper.class);
 
   private RequestContextUtil() {
   }
@@ -15,6 +19,7 @@ public class RequestContextUtil {
     }
     var modifiedHeaders = new CaseInsensitiveMap<>(requestContext.getHeaders());
     modifiedHeaders.put(XOkapiHeaders.TENANT, tenantId);
+    logger.info("Request context has been changes with new tenant: {}", tenantId);
     return new RequestContext(requestContext.getContext(), modifiedHeaders);
   }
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/orders/utils/RequestContextUtil.java
```diff
@@ -15,6 +18,7 @@ public static RequestContext createContextWithNewTenantId(RequestContext request
     }
     var modifiedHeaders = new CaseInsensitiveMap<>(requestContext.getHeaders());
     modifiedHeaders.put(XOkapiHeaders.TENANT, tenantId);
+    logger.info("Request context has been changes with new tenant: {}", tenantId);
```
imerabishvili, line 21: Really useful, nice!

src/main/java/org/folio/service/inventory/InventoryItemStatusSyncService.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return result;
+  }
+
+  private List<JsonObject> updateItemStatuses(List<JsonObject> items, ItemStatus status) {
+    items.forEach(item -> updateItemStatus(item, status));
+    return items;
+  }
+
+  private JsonObject updateItemStatus(JsonObject item, ItemStatus itemStatus) {
```
imerabishvili, line 135: **Minor**
Actually we do not create a new item but modify existing one so I'd make method void and update line 66 accordingly.

src/main/java/org/folio/service/inventory/InventoryItemStatusSyncService.java
```diff
@@ -0,0 +1,147 @@
+package org.folio.service.inventory;
```
imerabishvili, line 1: Can you please format the file? IDEA updates javadocs for me.

src/main/java/org/folio/service/inventory/InventoryItemStatusSyncService.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.folio.helper.BaseHelper.MAX_REPEAT_ON_FAILURE;
+import static org.folio.rest.RestConstants.MAX_IDS_FOR_GET_RQ_15;
+
+public class InventoryItemStatusSyncService {
```
imerabishvili, line 28: Do you agree that methods like `updateItemWithPieceFields` and baisaclly any `InventoryItemManager` method which is to complex and more than CRUD is a candidate for moving in this class?

imerabishvili (approved):

src/test/java/org/folio/service/inventory/InventoryItemStatusSyncServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+    JsonObject item = new JsonObject()
+      .put(InventoryItemManager.ITEM_PURCHASE_ORDER_LINE_IDENTIFIER, PO_LINE_ID);
+    when(inventoryItemManager.getItemsByPoLineIdsAndStatus(eq(List.of(PO_LINE_ID)), eq(ItemStatus.ON_ORDER.value()), any(RequestContext.class)))
+      .thenReturn(Future.succeededFuture(List.of(item)));
+    when(inventoryItemManager.updateItem(eq(item), any(RequestContext.class))).thenReturn(Future.succeededFuture());
+    Location location1 = new Location()
+      .withTenantId("tenant1");
+    Location location2 = new Location()
+      .withTenantId("tenant2");
+    List<Location> locations = List.of(location1, location2);
```
Saba-Zedginidze-EPAM, line 84: You could create a method that can generate Locations and choose whether to populate TenantId or not and use it in other locations. Would be better to avoid copy-pastes.

Or have them created in BeforeEach maybe

Saba-Zedginidze-EPAM (approved):

src/main/java/org/folio/service/inventory/InventoryItemStatusSyncService.java
SerhiiNosko: I just did not want to couple new service to Piece entity, in this case we can use InventoryItemStatusSyncService in places where pieces are not presented.

src/main/java/org/folio/service/inventory/InventoryItemStatusSyncService.java
SerhiiNosko: reformatted

src/test/java/org/folio/service/inventory/InventoryItemStatusSyncServiceTest.java
SerhiiNosko: intoroduced new private method

PR #936, opened 2024-05-10
[MODORDERS-1114] Add transferRequests flag to validate binding

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1114] Enhance bind-pieces endpoint with flag to transfer circulation requests](https://folio-org.atlassian.net/browse/MODORDERS-1114)

## Approach
* Add enum TransferRequests field to determine in case of outstanding requests for item to transfer these or do nothing
* Support binding of pieces with items in different tenants


src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -57,7 +66,9 @@ public Future<ReceivingResults> bindPieces(BindPiecesCollection bindPiecesCollec
   private Future<ReceivingResults> processBindPieces(BindPiecesCollection bindPiecesCollection, RequestContext requestContext) {
     // 1. Get piece records from storage
     return retrievePieceRecords(requestContext)
-      // 2. Update piece isBound flag
+      // 2. Check if there are any outstanding requests for items
+      .compose(piecesGroupedByPoLine -> checkRequestsForPieceItems(piecesGroupedByPoLine, bindPiecesCollection, requestContext))
+      // 3. Update piece isBound flag
       .map(this::updatePieceRecords)
       // 3. Update currently associated items
```
SerhiiNosko, line 73: comments should be update, so it should be 4

Saba-Zedginidze-EPAM: Forgot to do it after rebasing, fixed now

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -71,10 +82,25 @@ private Future<ReceivingResults> processBindPieces(BindPiecesCollection bindPiec
       .map(piecesGroupedByPoLine -> prepareResponseBody(piecesGroupedByPoLine, bindPiecesCollection));
   }
 
+  private Future<Map<String, List<Piece>>> checkRequestsForPieceItems(Map<String, List<Piece>> piecesGroupedByPoLine, BindPiecesCollection bindPiecesCollection, RequestContext requestContext) {
+    var itemIds = extractAllPieces(piecesGroupedByPoLine)
+      .map(Piece::getItemId)
+      .filter(StringUtils::isNotEmpty)
+      .toList();
+
+    return inventoryItemRequestService.getItemsWithActiveRequests(itemIds, requestContext)
```
SerhiiNosko, line 91: items that can be located in other tenants: For example item1 in tenantA, item2 in tenantB
Earch piece has receivingTenantId that we can use to fetch items from correct tenant.
Basically we did not have requirement for R to transfer requests across tenants. So for now we can add validation if pieces have differrent Piece::receivingTenantId - throw an error

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -112,11 +138,16 @@ private Future<Map<String, List<Piece>>> createItemForPiece(Map<String, List<Pie
       .map(PoLineCommonUtil::convertToCompositePoLine)
       .compose(compPOL ->
         inventoryItemManager.createBindItem(compPOL, holdingIds.get(0), bindPiecesCollection.getBindItem(), requestContext))
-      .map(itemId -> {
-          piecesGroupedByPoLine.get(poLineId).forEach(piece -> piece.setItemId(itemId));
-          return piecesGroupedByPoLine;
+      .map(newItemId -> {
+        // Transfer requests if flag is on
+        if (TRANSFER.equals(bindPiecesCollection.getTransferRequests())) {
```
SerhiiNosko, line 143: getTransderRequests sounds like collection of objects.
We should rename this to somethig like this: requestsAction: TRANSFER, CANCEL, DO_NOTHING

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -57,24 +66,41 @@ public Future<ReceivingResults> bindPieces(BindPiecesCollection bindPiecesCollec
...
+  private Future<Map<String, List<Piece>>> checkRequestsForPieceItems(Map<String, List<Piece>> piecesGroupedByPoLine, BindPiecesCollection bindPiecesCollection, RequestContext requestContext) {
+    var itemIds = extractAllPieces(piecesGroupedByPoLine)
+      .map(Piece::getItemId)
+      .filter(StringUtils::isNotEmpty)
+      .toList();
+
+    return inventoryItemRequestService.getItemsWithActiveRequests(itemIds, requestContext)
+      .map(items -> {
+        if (!items.isEmpty() && Objects.isNull(bindPiecesCollection.getTransferRequests())) {
+          logger.debug("checkRequestsForPieceItems:: Found outstanding requests on items with ids: {}", items);
```
azizbekxm, line 94: lets use warn or error level log

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Added another validation step

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Renamed, good point

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Changed the level

src/main/java/org/folio/rest/core/exceptions/ErrorCodes.java
```diff
@@ -83,6 +83,7 @@ public enum ErrorCodes {
   PIECES_TO_BE_CREATED("piecesNeedToBeCreated", "Pieces need to be created"),
   LOCATION_CAN_NOT_BE_MODIFIER_AFTER_OPEN("locationCannotBeModifiedAfterOpen", "Please use the receiving App to update pieces and locations"),
   REQUEST_FOUND("thereAreRequestsOnItem", "There are requests on item"),
+  REQUESTS_FOUND_WITH_TRANSFER_DISABLED("thereAreRequestsOnItemsAndTransferIsDisabled", "There are requests on items and transfer requests is disabled"),
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Modified the message

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Should this validation happen in which cases:
1) When requests are found and requestAction is TRANSFER
2) When requests are found and requestAction is not TRANSFER
3) When requests are not found

I am thinking only the 1st case should be applicable for this validation

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -124,13 +124,9 @@ private ReceivingResults prepareResponseBody(CheckinCollection checkinCollection
       results.getReceivingResults().add(result);
 
       // Get all processed piece records for PO Line
-      Map<String, Piece> processedPiecesForPoLine = StreamEx
-        .of(piecesGroupedByPoLine.getOrDefault(poLineId, Collections.emptyList()))
-        .toMap(Piece::getId, piece -> piece);
+      Map<String, Piece> processedPiecesForPoLine = getProcessedPiecesForPoLine(poLineId, piecesGroupedByPoLine);
```
SerhiiNosko, line 127: some unused imports are introduced, you can find them in sonar report:
https://sonarcloud.io/project/issues?id=org.folio%3Amod-orders&pullRequest=936&resolved=false&sinceLeakPeriod=true

src/main/java/org/folio/helper/BindHelper.java
SerhiiNosko: I think all actions except Do nothing.
Even if Cancel action is not used now, but we dont support it in multitenant env now, so should throw an error

Saba-Zedginidze-EPAM: Okay then should we validate it when requests are found or not?

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -55,26 +64,54 @@ public Future<ReceivingResults> bindPieces(BindPiecesCollection bindPiecesCollec
...
-      // 7. Return results to the client
+      // 8. Return results to the client
       .map(piecesGroupedByPoLine -> prepareResponseBody(piecesGroupedByPoLine, bindPiecesCollection));
   }
 
+  private Future<Map<String, List<Piece>>> checkRequestsForPieceItems(Map<String, List<Piece>> piecesGroupedByPoLine, BindPiecesCollection bindPiecesCollection, RequestContext requestContext) {
+    var itemIds = extractAllPieces(piecesGroupedByPoLine)
+      .map(Piece::getItemId)
+      .filter(StringUtils::isNotEmpty)
+      .toList();
```
SerhiiNosko, line 89: you should have some map here where tenantId will be a key and list of itemIds would be a value, and for each tenantId make request inventoryItemRequestService.getItemsWithActiveRequests(itemIds, requestContext).
To make request in another tenant you should modify requestContext, here is method that can help to do it: RequestContextUtil::createContextWithNewTenantId

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -55,26 +64,54 @@ public Future<ReceivingResults> bindPieces(BindPiecesCollection bindPiecesCollec
...
+
```
SerhiiNosko, line 90: the similar we should do here to update item status:
Each piece has receivingTenantId property that you can use to fetch items from correct tenant 
image

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Modified the methods to fetch items from different tenants

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Addresses this case aswell

SerhiiNosko: looks good

SerhiiNosko (approved):

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -783,6 +784,30 @@ protected void addErrorForUpdatingItem(Piece piece, Throwable e) {
     }
   }
 
+  protected Map<String, Piece> getProcessedPiecesForPoLine(String poLineId, Map<String, List<Piece>> piecesGroupedByPoLine) {
+    return StreamEx
+      .of(piecesGroupedByPoLine.getOrDefault(poLineId, Collections.emptyList()))
+      .toMap(Piece::getId, piece -> piece);
+  }
+
+  protected Map<String, Integer> getEmptyResultCounts() {
```
imerabishvili, line 793: Maybe (if map is not modified later)
```
    return Map.of(
      ProcessingStatus.Type.SUCCESS.toString(), 0,
      ProcessingStatus.Type.FAILURE.toString(), 0
    );
```

Saba-Zedginidze-EPAM: The values are updated, that's why didn't use .of()

imerabishvili: Actually map is modified later. Why we have `Map<String, int>` instead of `Map<ProcessingStatus.Type, int>`?

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -783,6 +784,30 @@ protected void addErrorForUpdatingItem(Piece piece, Throwable e) {
     }
   }
 
+  protected Map<String, Piece> getProcessedPiecesForPoLine(String poLineId, Map<String, List<Piece>> piecesGroupedByPoLine) {
```
imerabishvili, line 787: This group of 4 methods are misplaced in "Errors" section. Please check comments after dashes in this class:
image
Actually this class is already big and hard to manage. We can easily move new methods to corresponding classes but maybe this type of refactoring increases the scope a bit.

azizbekxm (approved):

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
Saba-Zedginidze-EPAM: That's a good point. I just refactored some irrelevant code and didn't go much into details. Will modify

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
Saba-Zedginidze-EPAM: Will move above the errors, thanks. I guess yes, the class is pretty big and can be refactored, unsure if in scope of this.

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -55,48 +65,82 @@ public Future<ReceivingResults> bindPieces(BindPiecesCollection bindPiecesCollec
...
+          var locationContext = RequestContextUtil.createContextWithNewTenantId(requestContext, entry.getKey());
+          return inventoryItemRequestService.getItemsWithActiveRequests(entry.getValue(), locationContext)
+            .compose(items -> {
+              if (!items.isEmpty()) {
+                // requestsAction is required to handle outstanding requests
+                if (Objects.isNull(bindPiecesCollection.getRequestsAction())) {
+                  logger.warn("checkRequestsForPieceItems:: Found outstanding requests on items with ids: {}", items);
+                  throw new HttpException(RestConstants.VALIDATION_ERROR, ErrorCodes.REQUESTS_ACTION_REQUIRED);
+                }
+                else if (bindPiecesCollection.getRequestsAction().equals(TRANSFER)) {
```
imerabishvili, line 102: Why do you need `else if` if you throw an error otherwise?
It's better to handle edge case first:
```
                if (items.isEmpty()) {
                }
```
and then continue with general case.
I'd write:
```
              if (!items.isEmpty() && Objects.isNull(bindPiecesCollection.getRequestsAction())) {
                logger.warn("checkRequestsForPieceItems:: Found outstanding requests on items with ids: {}", items);
                throw new HttpException(RestConstants.VALIDATION_ERROR, ErrorCodes.REQUESTS_ACTION_REQUIRED);
              }
              if (!items.isEmpty() && bindPiecesCollection.getRequestsAction().equals(TRANSFER) && !tenantToItem.keySet().isEmpty()) {
                logger.warn("checkRequestsForPieceItems:: All pieces do not have the same receivingTenantId: {}", tenantToItem.keySet());
                throw new HttpException(RestConstants.VALIDATION_ERROR, ErrorCodes.PIECES_HAVE_DIFFERENT_RECEIVING_TENANT_IDS);
              }
            });
```
This way I see a 2 cases where you throw an error but it's up to you. Main thing is to not keep many lines of code indented as this increases a complexity.

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -783,6 +784,30 @@ protected void addErrorForUpdatingItem(Piece piece, Throwable e) {
...
+    resultCounts.put(ProcessingStatus.Type.SUCCESS.toString(), 0);
+    resultCounts.put(ProcessingStatus.Type.FAILURE.toString(), 0);
+    return resultCounts;
+  }
+
+  protected StreamEx<Piece> extractAllPieces(Map<String, List<Piece>> piecesGroupedByPoLine) {
+    return StreamEx.ofValues(piecesGroupedByPoLine).flatMap(List::stream);
+  }
+
+  protected Map<String, List<String>> mapTenantIdsToItemIds(Map<String, List<Piece>> piecesGroupedByPoLine, RequestContext requestContext) {
```
imerabishvili, line 804: This is much like to what I did in `getItemRecrods`: lines 544-551. Maybe that place can reuse this?

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: Correct, better to get rid of nested conditions.

imerabishvili (approved):

PR #939, opened 2024-05-10
[MODORDERS-1115] - Extend bind pieces endpoint with functionality whe…

azizbekxm:
https://folio-org.atlassian.net/browse/MODORDERS-1115

- Created method to update previous title item ids
- and updated the bind item ids updating process

image

### DIFF
--- a/src/main/java/org/folio/helper/BindHelper.java
+++ b/src/main/java/org/folio/helper/BindHelper.java
@@ -8,9 +8,9 @@ import org.folio.models.ItemFields;
 import org.folio.orders.utils.PoLineCommonUtil;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.BindPiecesCollection;
-import org.folio.rest.jaxrs.model.CheckInPiece;
 import org.folio.rest.jaxrs.model.Piece;
 import org.folio.rest.jaxrs.model.ProcessingStatus;
+import org.folio.rest.jaxrs.model.ReceivedItem;
 import org.folio.rest.jaxrs.model.ReceivingResult;
 import org.folio.rest.jaxrs.model.ReceivingResults;
 import org.folio.rest.jaxrs.model.Title;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -73,6 +79,20 @@ private Map<String, List<Piece>> updatePieceRecords(Map<String, List<Piece>> pie
...
   }
 
+  private Map<String, List<Piece>> updateItemStatus(Map<String, List<Piece>> piecesGroupedByPoLine,
+                                                            RequestContext requestContext) {
+    List<String> itemIds = piecesGroupedByPoLine.values().stream().flatMap(List::stream)
+        .map(Piece::getItemId).toList();
+    inventoryItemManager.getItemRecordsByIds(itemIds, requestContext)
+      .compose(items -> {
+        items.forEach(item ->
+          item.put(ITEM_STATUS, new JsonObject().put("date", new Date()).put("name", "Unavailable"))
```
SerhiiNosko, line 89: we have an enum with item statuses that should be used

azizbekxm: Thanks, used

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -93,19 +116,23 @@ private Map<String, List<Piece>> updateTitleWithBindItems(Map<String, List<Piece
...
-      titlesService.getTitlesByPoLineIds(List.of(poLineId), requestContext)
-        .map(titles -> {
-            if (titles.containsKey(poLineId) && !titles.get(poLineId).isEmpty()) {
-              Title title = titles.get(poLineId).get(0).withBindItemIds(itemIds);
-              titlesService.saveTitle(title, requestContext);
-            }
-            return titles;
-          }
-        );
+      titlesService.getTitlesByQuery(String.format(TITLE_BY_POLINE_QUERY, poLineId), requestContext)
```
SerhiiNosko, line 119: current query will return titles by title id, because of id==%s
If you want to return by poLineId query should be like this: poLineId==%s

azizbekxm: good catch, thanks, fixed

SerhiiNosko (approved):

imerabishvili (approved):

PR #942, opened 2024-05-17
[MODORDERS-1108] Update PiecesAPI CRUD methods  to process inventory in desired tenant

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1108] Update PiecesAPI CRUD methods  to process inventory in desired tenant](https://folio-org.atlassian.net/browse/MODORDERS-1108)

## Approach
* Refactor existing classes
* Check for piece receivingTenantId, wherever present, use it to work on inventory items located in a different tenant.
* Keep both tenant contexts in order to be able to work for titles, pieces in the original tenant and also inventory objects in a different one

### DIFF
--- a/src/main/java/org/folio/orders/utils/RequestContextUtil.java
+++ b/src/main/java/org/folio/orders/utils/RequestContextUtil.java
@@ -6,6 +6,8 @@ import org.apache.logging.log4j.Logger;
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.rest.core.models.RequestContext;
 
+import java.util.Optional;
+
 public class RequestContextUtil {
   private static final Logger logger = LogManager.getLogger(RequestContextUtil.class);
 
--- a/src/main/java/org/folio/orders/utils/RequestContextUtil.java
+++ b/src/main/java/org/folio/orders/utils/RequestContextUtil.java
@@ -6,8 +6,6 @@ import org.apache.logging.log4j.Logger;
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.rest.core.models.RequestContext;
 
-import java.util.Optional;
-
 public class RequestContextUtil {
   private static final Logger logger = LogManager.getLogger(RequestContextUtil.class);
 
--- a/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
+++ b/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
@@ -24,6 +24,7 @@ import org.folio.rest.core.exceptions.InventoryException;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.CompositePoLine;
 import org.folio.rest.jaxrs.model.Piece;
+import org.folio.rest.jaxrs.model.Title;
 import org.folio.service.ProtectionService;
 import org.folio.service.inventory.InventoryHoldingManager;
 import org.folio.service.inventory.InventoryItemManager;
@@ -168,16 +169,24 @@ public class OpenCompositeOrderPieceService {
     var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
     return titlesService.getTitleById(piece.getTitleId(), requestContext)
       .compose(title -> titlesService.updateTitleWithInstance(title, isInstanceMatchingDisabled, locationContext, requestContext).map(title::withInstanceId))
-      .compose(title -> piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)
-        ? Future.succeededFuture(piece.getHoldingId())
-        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), locationContext)
-          .map(holdingId -> piece.withLocationId(null).withHoldingId(holdingId).getHoldingId()))
-      .compose(holdingId -> PoLineCommonUtil.isItemsUpdateRequired(compPOL)
-        ? inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, locationContext)
-        : Future.succeededFuture()
-      )
+      .compose(title -> getOrCreateHolding(compPOL, piece, title, locationContext))
+      .compose(holdingId -> updateItemsIfNeeded(compPOL, holdingId, locationContext))
       .map(itemId -> Optional.ofNullable(itemId).map(piece::withItemId))
       .mapEmpty();
+  }
+
+  private Future<String> getOrCreateHolding(CompositePoLine compPOL, Piece piece, Title title, RequestContext locationContext) {
+    if (piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)) {
+      return Future.succeededFuture(piece.getHoldingId());
+    }
+    return inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), locationContext)
+      .map(holdingId -> piece.withLocationId(null).withHoldingId(holdingId).getHoldingId());
+  }
+
+  private Future<String> updateItemsIfNeeded(CompositePoLine compPOL, String holdingId, RequestContext locationContext) {
+    return PoLineCommonUtil.isItemsUpdateRequired(compPOL)
+      ? inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, locationContext)
+      : Future.succeededFuture();
   }
 
   private void validateItemsCreation(CompositePoLine compPOL, int itemsSize) {
--- a/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
+++ b/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
@@ -3,6 +3,7 @@ package org.folio.service.orders.flows.update.open;
 import static java.util.stream.Collectors.toList;
 import static org.folio.orders.utils.HelperUtils.calculateInventoryItemsQuantity;
 import static org.folio.orders.utils.HelperUtils.collectResultsOnSuccess;
+import static org.folio.orders.utils.RequestContextUtil.createContextWithNewTenantId;
 
 import java.util.ArrayList;
 import java.util.Date;
@@ -13,14 +14,12 @@ import java.util.concurrent.CompletionException;
 import io.vertx.core.Future;
 import io.vertx.core.json.JsonObject;
 import org.apache.commons.collections4.CollectionUtils;
-import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.models.pieces.OpenOrderPieceHolder;
 import org.folio.orders.events.handlers.MessageAddress;
 import org.folio.orders.utils.PoLineCommonUtil;
 import org.folio.orders.utils.ProtectedOperationType;
-import org.folio.orders.utils.RequestContextUtil;
 import org.folio.rest.core.exceptions.InventoryException;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.CompositePoLine;
@@ -163,21 +162,21 @@ public class OpenCompositeOrderPieceService {
    * @return CompletableFuture with void.
    */
   public Future<Void> openOrderUpdateInventory(CompositePoLine compPOL, Piece piece, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
-    if (BooleanUtils.isNotTrue(compPOL.getIsPackage())) {
+    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
-    var locationContext = RequestContextUtil.createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
+    var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
     return titlesService.getTitleById(piece.getTitleId(), requestContext)
       .compose(title -> titlesService.updateTitleWithInstance(title, isInstanceMatchingDisabled, locationContext, requestContext).map(title::withInstanceId))
       .compose(title -> piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)
         ? Future.succeededFuture(piece.getHoldingId())
-        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), requestContext)
+        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), locationContext)
           .map(holdingId -> piece.withLocationId(null).withHoldingId(holdingId).getHoldingId()))
       .compose(holdingId -> PoLineCommonUtil.isItemsUpdateRequired(compPOL)
-        ? inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, requestContext)
+        ? inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, locationContext)
         : Future.succeededFuture()
       )
-      .onSuccess(itemId -> Optional.ofNullable(itemId).ifPresent(piece::setItemId))
+      .map(itemId -> Optional.ofNullable(itemId).map(piece::withItemId))
       .mapEmpty();
   }
 
--- a/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
+++ b/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
@@ -163,7 +163,7 @@ public class OpenCompositeOrderPieceService {
    * @return CompletableFuture with void.
    */
   public Future<Void> openOrderUpdateInventory(CompositePoLine compPOL, Piece piece, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
-    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
+    if (Boolean.FALSE.equals(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
     var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
--- a/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
+++ b/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
@@ -163,7 +163,7 @@ public class OpenCompositeOrderPieceService {
    * @return CompletableFuture with void.
    */
   public Future<Void> openOrderUpdateInventory(CompositePoLine compPOL, Piece piece, boolean isInstanceMatchingDisabled, RequestContext requestContext) {
-    if (Boolean.FALSE.equals(compPOL.getIsPackage())) {
+    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
     var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
--- a/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
+++ b/src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
@@ -13,12 +13,14 @@ import java.util.concurrent.CompletionException;
 import io.vertx.core.Future;
 import io.vertx.core.json.JsonObject;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.models.pieces.OpenOrderPieceHolder;
 import org.folio.orders.events.handlers.MessageAddress;
 import org.folio.orders.utils.PoLineCommonUtil;
 import org.folio.orders.utils.ProtectedOperationType;
+import org.folio.orders.utils.RequestContextUtil;
 import org.folio.rest.core.exceptions.InventoryException;
 import org.folio.rest.core.models.RequestContext;
 import org.folio.rest.jaxrs.model.CompositePoLine;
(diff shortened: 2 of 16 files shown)


src/main/java/org/folio/orders/utils/RequestContextUtil.java
```diff
@@ -6,6 +6,8 @@
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.rest.core.models.RequestContext;
 
+import java.util.Optional;
```
SerhiiNosko, line 9: seems it is redundant

Saba-Zedginidze-EPAM: Removed

src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowInventoryManager.java
```diff
@@ -29,80 +30,53 @@ public PieceCreateFlowInventoryManager(TitlesService titlesService,
...
-  public Future<Void> processInventory(CompositePoLine compPOL,  Piece piece,  boolean createItem, RequestContext requestContext) {
-    if (Boolean.TRUE.equals(compPOL.getIsPackage())) {
-      return packagePoLineUpdateInventory(compPOL, piece, createItem, requestContext);
-    }
-    else
-    {
-       return nonPackagePoLineUpdateInventory(compPOL, piece, createItem, requestContext);
-    }
+  public Future<Void> processInventory(CompositePoLine compPOL, Piece piece, boolean createItem, RequestContext requestContext) {
+    final var locationContext = RequestContextUtil.createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
```
imerabishvili, line 34: `final` can be removed

Saba-Zedginidze-EPAM: Removed

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -18,6 +18,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.folio.orders.utils.HelperUtils;
 import org.folio.orders.utils.ProtectedOperationType;
+import org.folio.orders.utils.RequestContextUtil;
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Removed all unused imports

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
-        })
-        .onSuccess(itemId -> Optional.ofNullable(itemId).ifPresent(piece::withItemId))
-        .mapEmpty();
-    }
-    else
-    {
+    if (BooleanUtils.isNotTrue(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
+    var locationContext = RequestContextUtil.createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
```
azizbekxm, line 169: can we remove RequestContextUtil class name and just use method itself

azizbekxm: if you are doing, please do in other changes in pr

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
-            return inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, requestContext);
-          }
-          return Future.succeededFuture();
-        })
-        .onSuccess(itemId -> Optional.ofNullable(itemId).ifPresent(piece::withItemId))
-        .mapEmpty();
-    }
-    else
-    {
+    if (BooleanUtils.isNotTrue(compPOL.getIsPackage())) {
```
SerhiiNosko, line 166: Boolean.TRUE.equals(compPOL.getIsPackage()) easier to read than BooleanUtils.isNotTrue(compPOL.getIsPackage()

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowInventoryManager.java
```diff
@@ -0,0 +1,61 @@
...
+      .compose(item -> Optional.ofNullable(item)
+        .map(mItem -> inventoryItemManager.deleteItem(piece.getItemId(), true, requestContext))
+        .orElse(Future.succeededFuture()));
+  }
+
+  private Future<JsonObject> getOnOrderItemForPiece(Piece piece, RequestContext requestContext) {
+    return StringUtils.isEmpty(piece.getItemId())
+      ? Future.succeededFuture()
+      : inventoryItemManager.getItemRecordById(piece.getItemId(), true, requestContext)
+          .map(item -> getItemWithStatus(item, ItemStatus.ON_ORDER.value()));
```
azizbekxm, line 51: ternary operators is good to use but sometimes, make it less readiable long chain of codes. we can do it like this
```
if (StringUtils.isEmpty(piece.getItemId())) {
      return Future.succeededFuture()
}
return inventoryItemManager.getItemRecordById(piece.getItemId(), true, requestContext)
    .map(item -> getItemWithStatus(item, ItemStatus.ON_ORDER.value()));
```

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
+    var locationContext = RequestContextUtil.createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
+    return titlesService.getTitleById(piece.getTitleId(), requestContext)
+      .compose(title -> titlesService.updateTitleWithInstance(title, isInstanceMatchingDisabled, locationContext, requestContext).map(title::withInstanceId))
+      .compose(title -> piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)
+        ? Future.succeededFuture(piece.getHoldingId())
+        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), requestContext)
```
SerhiiNosko, line 174: holdings the same as instances and items can be created in member tenant, so locationContext should be provided here

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
+          .map(holdingId -> piece.withLocationId(null).withHoldingId(holdingId).getHoldingId()))
+      .compose(holdingId -> PoLineCommonUtil.isItemsUpdateRequired(compPOL)
+        ? inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, requestContext)
```
SerhiiNosko, line 177: also location context should be provided to create item in any tenant

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
+        : Future.succeededFuture()
+      )
+      .onSuccess(itemId -> Optional.ofNullable(itemId).ifPresent(piece::setItemId))
```
SerhiiNosko, line 180: .onSuccess usage in previous implementation had drowbacks because it is terminal operation and can be executed even before last .mapEmpty(). Need to change to compose on map

src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowInventoryManager.java
```diff
@@ -29,80 +30,53 @@ public PieceCreateFlowInventoryManager(TitlesService titlesService,
...
-    return nonPackageUpdateTitleWithInstance(compPOL, piece.getTitleId(), requestContext)
-      .compose(instanceId -> handleHolding(compPOL, piece, instanceId, requestContext))
-      .compose(holdingId -> handleItem(compPOL, createItem, piece, requestContext))
-      .map(itemId -> {
-        Optional.ofNullable(itemId).ifPresent(piece::withItemId);
-        return null;
-      })
-      .mapEmpty();
+  private Future<String> updateInventoryForPoLine(CompositePoLine compPOL, Piece piece, RequestContext locationContext, RequestContext requestContext) {
+    if (BooleanUtils.isNotTrue(compPOL.getIsPackage())) {
```
SerhiiNosko, line 43: the same with method isNotTrue , isTrue is easier to understand quickly

src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowInventoryManager.java
```diff
@@ -29,80 +30,53 @@ public PieceCreateFlowInventoryManager(TitlesService titlesService,
...
-  private Future<Void> nonPackagePoLineUpdateInventory(CompositePoLine compPOL, Piece piece, boolean createItem, RequestContext requestContext) {
-    return nonPackageUpdateTitleWithInstance(compPOL, piece.getTitleId(), requestContext)
-      .compose(instanceId -> handleHolding(compPOL, piece, instanceId, requestContext))
-      .compose(holdingId -> handleItem(compPOL, createItem, piece, requestContext))
-      .map(itemId -> {
-        Optional.ofNullable(itemId).ifPresent(piece::withItemId);
-        return null;
-      })
-      .mapEmpty();
+  private Future<String> updateInventoryForPoLine(CompositePoLine compPOL, Piece piece, RequestContext locationContext, RequestContext requestContext) {
```
SerhiiNosko, line 42: Inventory it is very wide word, can be more speocifc updateInventoryInstanceForPoLine

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowInventoryManager.java
Saba-Zedginidze-EPAM: I think ternary is okay with 1 or 2 lines of code but if it was less readable for you, surely I will change this

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -1,59 +1,42 @@
...
 import org.folio.service.pieces.flows.BasePieceFlowHolderBuilder;
 
 import io.vertx.core.Future;
-import io.vertx.core.json.JsonObject;
 
 public class PieceDeleteFlowManager {
+
   private static final Logger logger = LogManager.getLogger(PieceDeleteFlowManager.class);
 
+  private final PieceDeleteFlowInventoryManager pieceDeleteFlowInventoryManager;
```
SerhiiNosko, line 24: It is a bit hard to understand when manager injects another manager. We can rename PieceDeleteFlowInventoryManager . We already have PieceUpdateInventoryService. Maybe it makes sense to rename to PieceDeleteInventoryService that will not have dependency to PieceUpdateInventoryService.
And PieceDeleteFlowManager will invoke both PieceUpdateInventoryService and PieceDeleteInventoryService

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Replaced BooleanUtils in several places

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Good catch, seems like I forgot to pass it here, fixed now

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Fixed this one too

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Good point, replaced it with map

src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowInventoryManager.java
Saba-Zedginidze-EPAM: Renamed

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
Saba-Zedginidze-EPAM: Introduced the new service

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Replaced everywhere in PR

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
SerhiiNosko: now contract is clear, thanks

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +162,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
+    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
+    var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
+    return titlesService.getTitleById(piece.getTitleId(), requestContext)
+      .compose(title -> titlesService.updateTitleWithInstance(title, isInstanceMatchingDisabled, locationContext, requestContext).map(title::withInstanceId))
+      .compose(title -> piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)
+        ? Future.succeededFuture(piece.getHoldingId())
+        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), locationContext)
+          .map(holdingId -> piece.withLocationId(null).withHoldingId(holdingId).getHoldingId()))
```
SerhiiNosko, line 174: a bit harder to read code now, it is necessary to apply more energy to understand it
But maybe it is just for me, resolving comment in any case because logically it is the same as before

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +162,22 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
-    {
+    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
       return inventoryItemManager.updateItemWithPieceFields(piece, requestContext);
     }
+    var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
+    return titlesService.getTitleById(piece.getTitleId(), requestContext)
+      .compose(title -> titlesService.updateTitleWithInstance(title, isInstanceMatchingDisabled, locationContext, requestContext).map(title::withInstanceId))
+      .compose(title -> piece.getHoldingId() != null || !PoLineCommonUtil.isHoldingsUpdateRequired(compPOL)
+        ? Future.succeededFuture(piece.getHoldingId())
+        : inventoryHoldingManager.createHoldingAndReturnId(title.getInstanceId(), piece.getLocationId(), locationContext)
```
BKadirkhodjaev, line 173: Perhaps you can extract this inner inventoryHoldingManager.createHoldingAndReturnId() into a separate private method, it does look complicated to read

src/main/java/org/folio/service/pieces/PieceUpdateInventoryService.java
```diff
@@ -63,13 +64,14 @@ public Future<String> manualPieceFlowCreateItemRecord(Piece piece, CompositePoLi
 
   public Future<Pair<String, String>> deleteHoldingConnectedToPiece(Piece piece, RequestContext requestContext) {
     if (piece != null && piece.getHoldingId() != null) {
+      var locationContext = createContextWithNewTenantId(requestContext, piece.getReceivingTenantId());
       String holdingId = piece.getHoldingId();
-      return inventoryHoldingManager.getHoldingById(holdingId, true, requestContext)
+      return inventoryHoldingManager.getHoldingById(holdingId, true, locationContext)
                   .compose(holding -> {
```
BKadirkhodjaev, line 70: I suggest we extract this block to a private method or perform smaller extraction work for each nested callback, if possible

SerhiiNosko: maybe lets do this refactoring in subsequent PRs because this PR already contain lot of refactoring that slows down code review and we need to finish it and merge

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
```diff
@@ -45,73 +47,62 @@ public PieceUpdateFlowInventoryManager(TitlesService titlesService,
...
+    var location = new Location().withLocationId(pieceToUpdate.getLocationId());
+    if (instanceId == null || !DefaultPieceFlowsValidator.isCreateHoldingForPiecePossible(pieceToUpdate, poLineToSave)) {
+      return Future.succeededFuture(location);
     }
-    return Future.succeededFuture(new Location().withLocationId(pieceToUpdate.getLocationId()));
+    return inventoryHoldingManager.createHoldingAndReturnId(instanceId, pieceToUpdate.getLocationId(), requestContext)
+      .map(holdingId -> {
+        if (holdingId != null) {
+          pieceToUpdate.withHoldingId(null).setHoldingId(holdingId);
+          location.withLocationId(null).setHoldingId(holdingId);
```
azizbekxm, line 102: is it supposed to be like this `pieceToUpdate.withLocationId(null).setHoldingId(holdingId)`

src/main/java/org/folio/service/pieces/PieceUpdateInventoryService.java
Saba-Zedginidze-EPAM: @SerhiiNosko As it is not quire readable, I can modify the two places Bob mentioned without much work and take care of the rest if there will be anything needed in the future

SerhiiNosko: okay got it, let's do

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowInventoryManager.java
Saba-Zedginidze-EPAM: That's right, thanks for pointing out

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: Good suggestion, modified code for both your comments

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -161,36 +163,30 @@ public Future<Void> updatePieceRecord(Piece piece, RequestContext requestContext
...
-            return inventoryItemManager.openOrderCreateItemRecord(compPOL, holdingId, requestContext);
-          }
-          return Future.succeededFuture();
-        })
-        .onSuccess(itemId -> Optional.ofNullable(itemId).ifPresent(piece::withItemId))
-        .mapEmpty();
-    }
-    else
-    {
+    if (!Boolean.TRUE.equals(compPOL.getIsPackage())) {
```
azizbekxm, line 166: Boolean.FALSE.equals(compPOL.getIsPackage() is more readable from my prespective.

azizbekxm: if you are chaning, please change in other places of pr

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -62,89 +45,39 @@ public PieceDeleteFlowManager(PieceStorageService pieceStorageService,
...
+      .mapEmpty();
   }
 
-  private boolean isItemWithStatus(JsonObject item, String status) {
-    return Optional.ofNullable(item).map(itemP -> item.getJsonObject(ITEM_STATUS))
-      .filter(itemStatus -> status.equalsIgnoreCase(itemStatus.getString(ITEM_STATUS_NAME)))
-      .isPresent();
+  protected Future<Void> updatePoLine(PieceDeletionHolder holder, RequestContext requestContext) {
+    var comPOL = holder.getOriginPoLine();
+    return BooleanUtils.isTrue(comPOL.getIsPackage()) || BooleanUtils.isTrue(comPOL.getCheckinItems())
```
azizbekxm, line 78: to be consistent, lets use Boolean.FALSE or TRUE equals

azizbekxm: If you are doing, please change in other places of pr

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
Saba-Zedginidze-EPAM: I think if we use compared to FALSE and the getIsPackage() is null, then it will bypass the check, so comparing to TRUE eliminates the need to check for nulls, as both false and null would give same result when using TRUE. What do you think about this?

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
Saba-Zedginidze-EPAM: Got rid of it in all places

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
SerhiiNosko: Basically it can be a dangerous swap because boolean flags can be null in our modules.
But for this module we are okay because we have default value declared as false. 

image

SerhiiNosko: I see that this field isPackage added in 2020.
Interesting what it will be for orders created before 2020. Will it be automatically populated as false
image

Saba-Zedginidze-EPAM: @SerhiiNosko Thanks for the heads up, changed to FALSE

SerhiiNosko: So if some onging order created in 2019 and will have  this check now: Boolean.FALSE.equals(compPOL.getIsPackage()) - it will return null and we will consider this order as non package that we extect, so I think we can use this check

azizbekxm: @Saba-Zedginidze-EPAM thanks for explanation, never thought about before, You can leave as it was.

Saba-Zedginidze-EPAM: @SerhiiNosko but would it not match with the current schema and populate with default values if it is missing?

SerhiiNosko: to be on safe side lets revert to initial state

Saba-Zedginidze-EPAM: Got it, reverted

SerhiiNosko (approved):

PR #947, opened 2024-05-23
[MODORDERS-1112] Sync bind pieces state with newly created item in correct tenant

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1112] Sync bind pieces state with newly created item in correct tenant](https://folio-org.atlassian.net/browse/MODORDERS-1112)

## Approach
After item is created during binding, do following:
* Clear these fields as piece can have these fields populated from previous item: displaySummary, enumeration, copyNumber, chronology, accessionNumber, discoverySuppress
* Populate these fields from bindItem: barcode, callNumber, locationId

### Pieces before binding:
Display Summary and Barcode are set
Pasted image 20240523142639

### Binding call
Pasted image 20240523142712

### Pieces after binding:
Display Summary is gone, and Barcode is updated
Pasted image 20240523143141

### DIFF
--- a/src/main/java/org/folio/helper/BindHelper.java
+++ b/src/main/java/org/folio/helper/BindHelper.java
@@ -75,11 +75,13 @@ public class BindHelper extends CheckinReceivePiecesHelper<BindPiecesCollection>
       .compose(piecesGroupedByPoLine -> updateItemStatus(piecesGroupedByPoLine, requestContext))
       // 5. Crate item for pieces with specific fields
       .compose(piecesGroupedByPoLine -> createItemForPiece(piecesGroupedByPoLine, bindPiecesCollection, requestContext))
-      // 6. Update received piece records in the storage
+      // 6. Sync pieces with new item fields and clear old ones
+      .map(piecesGroupedByPoLine -> syncPiecesWithNewItem(piecesGroupedByPoLine, bindPiecesCollection))
+      // 7. Update received piece records in the storage
       .compose(piecesGroupedByPoLine -> storeUpdatedPieceRecords(piecesGroupedByPoLine, requestContext))
-      // 7. Update Title with new bind items
+      // 8. Update Title with new bind items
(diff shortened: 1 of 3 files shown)


SerhiiNosko (approved):

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -179,6 +181,26 @@ private void validateHoldingIds(List<String> holdingIds, BindPiecesCollection bi
...
+        .withDisplaySummary(null)
+        .withEnumeration(null)
+        .withCopyNumber(null)
+        .withChronology(null)
+        .withAccessionNumber(null)
+        .withDiscoverySuppress(null)
+        // Populated fields contained in new item
+        .withBarcode(bindItem.getBarcode())
+        .withCallNumber(bindItem.getCallNumber())
+        .withLocationId(bindItem.getPermanentLocationId()));
```
SerhiiNosko, line 200: will discuss these fields on standup

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -216,15 +216,12 @@ private void checkIfAllPiecesFound(List<String> pieceIds, List<Piece> pieces) {
...
-      .map(piece -> piece.withStatusUpdatedDate(new Date()))
-      .map(piece -> storeUpdatedPieceRecord(piece, requestContext))
-      .toList();
-
-    return GenericCompositeFuture.join(futures)
+    return GenericCompositeFuture.join(
+      extractAllPieces(piecesGroupedByPoLine)
+        .filter(this::isSuccessfullyProcessedPiece)
+        .map(piece -> piece.withStatusUpdatedDate(new Date()))
+        .map(piece -> storeUpdatedPieceRecord(piece, requestContext))
```
azizbekxm, line 223: we have duplicate call to this method https://github.com/folio-org/mod-orders/pull/947/files#diff-3a7f736956a9ecfff0816c2d0a92c1261471143f91a3a5ab53b1dcd448892117R81

PR #948, opened 2024-05-23
MODORDERS-1122] Add logic to create item in any tenant for binding

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1122] Add logic to create item in any tenant for binding](https://folio-org.atlassian.net/browse/MODORDERS-1122)

## Approach
* Add tenantId to BindItem schema
* If tenantId is present while binding, create item in the specified tenant

### DIFF
--- a/src/main/java/org/folio/helper/BindHelper.java
+++ b/src/main/java/org/folio/helper/BindHelper.java
@@ -28,6 +28,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.folio.rest.jaxrs.model.BindPiecesCollection.RequestsAction.TRANSFER;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -28,6 +28,8 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
```
SerhiiNosko, line 32: please create separate PR in acq-models also with schema changes

SerhiiNosko (approved):

src/main/java/org/folio/helper/BindHelper.java
azizbekxm: then don't forget to point acq-models to master after acq-model pr merged

Saba-Zedginidze-EPAM: Sorry forgot that one, here's PR: https://github.com/folio-org/mod-orders/pull/948

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -88,34 +90,41 @@ private Future<Map<String, List<Piece>>> checkRequestsForPieceItems(Map<String,
...
+    }
+
+    // requestsAction is required to handle outstanding requests
+    if (Objects.isNull(bindPiecesCollection.getRequestsAction())) {
+      logger.warn("validateItemsForRequestTransfer:: Found outstanding requests on items with ids: {}", items);
+      throw new HttpException(RestConstants.VALIDATION_ERROR, ErrorCodes.REQUESTS_ACTION_REQUIRED);
+    }
+
+    var bindItemTenantId = bindPiecesCollection.getBindItem().getTenantId();
+    var itemsInSameTenant = Optional.ofNullable(bindItemTenantId)
```
azizbekxm, line 117: let's make it isItemInSameTenant to be more self-explanatory as it is boolean

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -88,34 +90,41 @@ private Future<Map<String, List<Piece>>> checkRequestsForPieceItems(Map<String,
...
   }
 
+  private Future<Void> validateItemsForRequestTransfer(Set<String> tenants,
+                                                       List<String> items,
+                                                       BindPiecesCollection bindPiecesCollection) {
+    if (items.isEmpty()) {
+      return Future.succeededFuture();
+    }
+
+    // requestsAction is required to handle outstanding requests
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Renamed all usages

src/main/java/org/folio/helper/BindHelper.java
Saba-Zedginidze-EPAM: areItemsInSameTenant would be more precise in this case, as the code check if all items will be kept in the same tenant, thus the boolean check

src/main/java/org/folio/helper/BindHelper.java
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Okay I see what you mean. I used outstanding just because the requests were described this way in the story, but essentially it means open requests. Will rename with "open"

azizbekxm (approved):

PR #952, opened 2024-05-28
[MODORDER-1087]-Delete received pieces in bulk

yuntianhu:
Description

Purpose: Libraries that use the Receiving app for receipt of print serials may wish to delete pieces that precede a specific date. Currently, FOLIO supports the deletion of pieces, but one at a time. When a library wishes to purge receiving records for multiple years on an ongoing workflow, the ability to delete pieces in bulk will streamline this workflow. 

Weeding old journal issues; receiving records no longer necessary since items have been withdrawn

User story statement(s):

As a serials librarian,
I want to delete multiple pieces at one time
so that the continuous addition of pieces does not eventually impact performance of the Receiving app.

Scenarios:

Bulk edit received/expected pieces:

Given user has received multiple piece records for a given receiving title

When expanding received/expected pieces accordion

Then user has option in action menu for Bulk edit

Select one or more pieces:

Given user clicks Bulk edit

When toggles appear in table

Then user can select one or more pieces

Edit pieces:

Given user has selected one or more pieces

When user clicks edit piece button

Then Bulk edit piece form is displayed

Delete all pieces

Given user clicks edit piece button

When Bulk edit piece form is displayed

Then "Delete all" button is active

Delete all pieces confirmation

Given user is viewing Bulk edit piece form

When user clicks "Delete all"

Then confirmation modal is shown

Message: "This will permanently remove all the pieces you have selected. Any related items with a status of "On order" will also be deleted. Would you like to proceed?

AND user can confirm or cancel

Library enhancement suggested by Lafayette College (May 2022 adopter)

Environment

None


src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -79,4 +85,26 @@ protected Future<Void> updatePoLine(PieceDeletionHolder holder, RequestContext r
       : pieceDeleteFlowPoLineService.updatePoLine(holder, requestContext);
   }
 
+  public Future<List<Void>> batchDeletePiece (PieceCollection entity, RequestContext requestContext) {
```
SerhiiNosko, line 88: here is some formatting issue, also we need to accept list of ids to delete instead of full PieceCollection object
```suggestion
  public Future<List<Void>> batchDeletePieces(List<String> pieceIds, RequestContext requestContext) {
```

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -79,4 +85,26 @@ protected Future<Void> updatePoLine(PieceDeletionHolder holder, RequestContext r
...
+    List <String> ids = new ArrayList<>();
```
SerhiiNosko, line 89: we should keep an existing logic that Delete by Id endpoint uses with many existing checks.
In the simplest way you can invoke this method in cycle:
Future<Void> deletePiece(String pieceId, boolean deleteHolding, RequestContext requestContext) 
It is not optimezed solution of course, but can be as starting point

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
```diff
@@ -79,4 +85,26 @@ protected Future<Void> updatePoLine(PieceDeletionHolder holder, RequestContext r
...
+    entity.getPieces().stream().forEach(v -> ids.add(v.getId()));
+    List<Future<PieceDeletionHolder>> deletionHolders = ids.stream()
+      .map(pieceId -> {
+        PieceDeletionHolder holder = new PieceDeletionHolder().withDeleteHolding(true);
```
SerhiiNosko, line 93: we could not hardcode deleteHolding to true, it should be fetched from request param as doing in Delete By Id endpoint

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
yuntianhu: Sure, thank you for pointing it out, will do! :)

src/main/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManager.java
yuntianhu: thank you for pointing it out, I will correct.

src/test/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManagerTest.java
```diff
@@ -368,6 +370,67 @@ void shouldUpdateLineQuantityIfPoLineIsNotPackageAndManualPieceCreateFalseAndInv
...
+      return succeededFuture(null);
+    }).when(basePieceFlowHolderBuilder).updateHolderWithOrderInformation(PieceDeletionHolderCapture.capture(), eq(requestContext));
+    doAnswer((Answer<Future<Void>>) invocation -> {
+      PieceDeletionHolder answerHolder = invocation.getArgument(0);
+      answerHolder.withTitleInformation(title);
+      return succeededFuture(null);
+    }).when(basePieceFlowHolderBuilder).updateHolderWithTitleInformation(PieceDeletionHolderCapture.capture(), eq(requestContext));
+
+    final ArgumentCaptor<PieceDeletionHolder> pieceDeletionHolderCapture = ArgumentCaptor.forClass(PieceDeletionHolder.class);
+    doReturn(succeededFuture(null)).when(pieceDeleteFlowPoLineService).updatePoLine(pieceDeletionHolderCapture.capture(), eq(requestContext));
```
azizbekxm, line 426: also separate declaration doReturn and action and verification part to be easy to read

src/test/java/org/folio/rest/impl/PieceApiTest.java
```diff
@@ -356,4 +362,95 @@ void deletePieceInternalErrorOnStorageTest() {
...
+    JsonArray jsonArrary = new JsonArray(ids);
+    JsonObject jsonObject = new JsonObject()
+      .put("ids", jsonArrary);
+    //  .put("deleteHolding", deleteHoldings);
+
+    List <Piece> pieces = new ArrayList<>();
+
+
+    // Mock response setup for deletion
+    //MockServer.addMockEntry(PIECES_STORAGE, JsonObject.mapFrom(jsonObject).encode()); // Simplified mock data
```
azizbekxm, line 382: also remove commentds, if it is not helpful

ramls/pieces.raml
```diff
@@ -47,6 +60,42 @@ resourceTypes:
...
+        application/json:
+          type: string
+          example: |
+            {
+              "ids": ["123", "456", "789"],
+              "deleteHoldings": false
+            }
+      responses:
+        204:
+          description: "Pieces records successfully deleted"
```
azizbekxm, line 76: `Piece records` or `Pieces ` more correct

azizbekxm: Also refactor Sentences in other places to be gramatically correct

src/test/java/org/folio/rest/impl/PieceApiTest.java
yuntianhu: thank you for pointing it out ! will do

ramls/pieces.raml
yuntianhu: thank you, sure will.

PR #959, opened 2024-06-06
Modify bind pieces endpoint to accept holdingId and locationId

Saba-Zedginidze-EPAM:
## Purpose
Modify bind pieces endpoint to accept holdingId and locationId

## Approach
* Add holdingId and locationId to bindPiecesCollection.json
* Remove existing holdings validation
* Add validation to include only one field in bindItem: holdingId or locationId
* Create holding if holdingId is not present

### DIFF
--- a/src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
+++ b/src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
@@ -1035,7 +1035,6 @@ public class CheckinReceivingApiTest {
   void testBindPiecesWithDifferentHoldingIdAndThrowError() {
     logger.info("=== Test POST Bind with different holdingId to Title With and throw error");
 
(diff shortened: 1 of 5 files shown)


SerhiiNosko (approved):

src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
```diff
@@ -1031,45 +1031,6 @@ void testBindPiecesToTitleWithItem() {
 
   }
 
-  @Test
-  void testBindPiecesWithDifferentHoldingIdAndThrowError() {
```
azizbekxm, line 1035: We can change this test to the case when location and holding exists or not exists

Saba-Zedginidze-EPAM: Replaced test with holding creation scenario

azizbekxm (approved):

PR #963, opened 2024-06-14
[MODORDERS-1130] ECS | Order can not be opened if contains three locations from different tenants

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1130

mod-consortia has endpoint that we can use to know in each tenant instanse was shared that should be used

### DIFF
--- a/src/main/java/org/folio/service/consortium/SharingInstanceService.java
+++ b/src/main/java/org/folio/service/consortium/SharingInstanceService.java
@@ -6,6 +6,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.models.consortium.ConsortiumConfiguration;
 import org.folio.models.consortium.SharingInstance;
+import org.folio.models.consortium.SharingInstanceCollection;
 import org.folio.models.consortium.SharingStatus;
 import org.folio.rest.core.RestClient;
 import org.folio.rest.core.exceptions.ConsortiumException;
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/service/consortium/SharingInstanceService.java
```diff
@@ -67,4 +71,23 @@ private Future<SharingInstance> shareInstance(String consortiumId, SharingInstan
...
+   *
+   * @param instanceId  the instance id
+   * @param consortiumConfiguration the consortium configuration
+   * @param requestContext the request context
+   * @return list with shared instances
+   */
+  public Future<SharingInstanceCollection> getSharingInstances(String instanceId,
+                                                               ConsortiumConfiguration consortiumConfiguration,
+                                                               RequestContext requestContext) {
+    RequestEntry requestEntry = new RequestEntry(SHARE_INSTANCE_ENDPOINT)
```
imerabishvili, line 85: **Question**
How is this implemented? Does it iterates over all tenants and tries to obtain instance with given id?

SerhiiNosko: We have table in mod-consortia central tenant sharing_instance that knows all about tenants where shared instance exist, so we use it here.
Here is detailed design diagram: https://folio-org.atlassian.net/wiki/spaces/DD/pages/1781565/Sharing+Instances+With+Consortium+Members

imerabishvili (approved):

azizbekxm (approved):

PR #971, opened 2024-06-20
MODORDERS-1120. Fix into populating searchLocationIds

SerhiiNosko:
## Purpose
Fix issue to not create new Po Line object for https://folio-org.atlassian.net/browse/MODORDERS-1120

### DIFF
--- a/src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
+++ b/src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
@@ -499,18 +499,19 @@ public class PurchaseOrderLineService {
       });
   }
 
-  public Future<Void> updateSearchLocations(CompositePoLine poLine, RequestContext requestContext) {
-    return updateSearchLocations(HelperUtils.convertToPoLine(poLine), requestContext);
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
```diff
@@ -499,18 +499,19 @@ public Future<Void> validateAndNormalizeISBNCommon(List<CompositePoLine> composi
       });
   }
 
-  public Future<Void> updateSearchLocations(CompositePoLine poLine, RequestContext requestContext) {
-    return updateSearchLocations(HelperUtils.convertToPoLine(poLine), requestContext);
```
SerhiiNosko, line 503: root cause of not populating searchLoctionIds was creating new object by HelperUtils.convertToPoLine(poLine), so searchLocationIds are set to this new object instead of original poLine

BKadirkhodjaev (approved):

imerabishvili (approved):

Saba-Zedginidze-EPAM (approved):

PR #972, opened 2024-06-20
added missing permission for changing status of circulation request and making improvements

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/FAT-14355
there is issue with permission and exception handling
```
ERROR FutureImpl$2         org.folio.rest.core.exceptions.HttpException: Access for user 'test-user' (00000000-1111-5555-9999-999999999992) requires permission: circulation.requests.item.put
```
## Approach
adding permission to change circulation request status for transfer action and adding exception handling to notify user about error. 

Removing unsed method. This will be stored in version control system, if it need in the future

### DIFF
--- a/src/main/java/org/folio/service/inventory/InventoryItemRequestService.java
+++ b/src/main/java/org/folio/service/inventory/InventoryItemRequestService.java
@@ -12,7 +12,6 @@ import org.folio.service.CirculationRequestsRetriever;
 
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/service/inventory/InventoryItemRequestService.java
```diff
@@ -51,16 +50,6 @@ public Future<List<String>> getItemIdsWithActiveRequests(List<String> itemIds, R
         .toList());
   }
 
-  public Future<Void> cancelItemRequests(List<String> itemIds, RequestContext requestContext) {
```
azizbekxm, line 54: since this method is not being used, removing it.

SerhiiNosko (approved):

Saba-Zedginidze-EPAM (approved):

PR #983, opened 2024-07-18
[MODORDERS-1141]. ECS | Item during receiving is not moved to new location, but duplicates in original location

BKadirkhodjaev:
## Purpose

- https://folio-org.atlassian.net/browse/MODORDERS-1141
- https://folio-org.atlassian.net/browse/MODORDERS-1158

## Approach

- For MODORDERS-1141:
  - Create a new service
  - Create item in member tenant 2
  - Delete item in member tenant 1
- For MODORDERS-1158:
  - Add tenant id to POL location on location update
- Test using MockServer-based integration tests


src/main/java/org/folio/orders/utils/ResourcePathResolver.java
```diff
@@ -96,6 +99,9 @@ private ResourcePathResolver() {
     apis.put(USERS, "/users");
     apis.put(ORDER_SETTINGS, "/orders-storage/settings");
     apis.put(ROUTING_LISTS, "/orders-storage/routing-lists");
+    apis.put(INSTANCE_STORAGE, "/inventory/instances");
+    apis.put(HOLDINGS_STORAGE, "/holdings-storage/holdings");
+    apis.put(ITEMS_STORAGE, "/inventory/items");
```
SerhiiNosko, line 104: for Inventory I see that we have InventoryUtils.INVENTORY_LOOKUP_ENDPOINTS
Better to have single declaration of api endpoint in the module

azizbekxm (approved):

src/main/java/org/folio/service/pieces/ItemRecreateInventoryService.java
```diff
@@ -0,0 +1,84 @@
...
+  public ItemRecreateInventoryService(InventoryItemManager inventoryItemManager) {
+    this.inventoryItemManager = inventoryItemManager;
+  }
+
+  // Return an Id of the recreated Item
+  public Future<String> recreateItemInDestinationTenant(PieceUpdateHolder holder, RequestContext srcLocCtx, RequestContext dstLocCtx) {
+    // Example Case: Member Tenant 1 (University) -> Member Tenant 2 (College)
+    // Create Item in Member Tenant 2 with the same Item Id
+    // Delete Item in Member Tenant 1 by Item Id
+    var itemFuture = Promise.<String>promise();
```
SerhiiNosko, line 36: this code can be rewritten without using promises. It simplifies code. We can use promises in complex scenarios with lot of diffrent conditions and input points.
So you can save result of inventoryItemManager in some future and make compose to delete an item

src/main/java/org/folio/service/pieces/ItemRecreateInventoryService.java
```diff
@@ -0,0 +1,84 @@
...
+    var compPol = holder.getPoLineToSave();
+
+    try {
+      logger.debug("Handling {} items for PO Line and holdings with id={}", ITEM_QUANTITY, piece.getHoldingId());
+
+      if (piece.getFormat() == Piece.Format.ELECTRONIC && DefaultPieceFlowsValidator.isCreateItemForElectronicPiecePossible(piece, compPol)) {
+        inventoryItemManager.recreateMissingElectronicItems(compPol, piece, ITEM_QUANTITY, dstLocCtx)
+          .onSuccess(idS -> itemFuture.complete(idS.get(0)))
+          .onFailure(itemFuture::fail);
+      } else if (DefaultPieceFlowsValidator.isCreateItemForNonElectronicPiecePossible(piece, compPol)) {
```
SerhiiNosko, line 47: similar if else statement as in PieceUpdateInventoryService, that can be potentially extracted in common class that will have 2 abstact method to override, but can be looked as overengineering. I am ok to leave as it, just thoughts

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -370,6 +392,29 @@ public Future<List<String>> createMissingPhysicalItems(CompositePoLine compPOL,
...
+                                                           RequestContext requestContext) {
+    if (quantity <= 0) {
+      return Future.succeededFuture(List.of());
+    }
+    String holdingId = piece.getHoldingId();
+    return buildPhysicalItemRecordJsonObject(compPOL, holdingId, requestContext)
+      .compose(item -> {
+        InventoryUtils.updateItemWithPieceFields(item, piece);
+        item.put(ID, piece.getItemId());
+        logger.debug("Posting {} physical item(s) for PO Line with '{}' id and '{}' item id", quantity, compPOL.getId(), piece.getItemId());
```
SerhiiNosko, line 413: let's change log level to info and for similar new places. This is useful information to see in logs in any case.

SerhiiNosko: by the way, we already have methods: createMissingPhysicalItems, createMissingElectronicItems.
So they differ only by this line: item.put(ID, piece.getItemId());
As one of potential solution you can add new parameter itemId to them and previous callers will send null or random uuids.

BKadirkhodjaev: > by the way, we already have methods: createMissingPhysicalItems, createMissingElectronicItems. So they differ only by this line: item.put(ID, piece.getItemId()); As one of potential solution you can add new parameter itemId to them and previous callers will send null or random uuids.

Yeah, I originally thought doing so, but that would be bit disruptive, I will make some changes now to reusable the existing 2 create methods

src/main/java/org/folio/service/pieces/ItemRecreateInventoryService.java
```diff
@@ -0,0 +1,81 @@
...
+      if (piece.getFormat() == Piece.Format.ELECTRONIC && DefaultPieceFlowsValidator.isCreateItemForElectronicPiecePossible(piece, compPol)) {
+        itemFuture = inventoryItemManager.createMissingElectronicItems(compPol, piece, ITEM_QUANTITY, dstLocCtx);
+      } else if (DefaultPieceFlowsValidator.isCreateItemForNonElectronicPiecePossible(piece, compPol)) {
+        itemFuture = inventoryItemManager.createMissingPhysicalItems(compPol, piece, ITEM_QUANTITY, dstLocCtx);
+      } else {
+        return Future.succeededFuture(null);
+      }
+
+      return itemFuture
+        .map(itemIds -> itemIds.stream().findFirst().orElseThrow())
```
SerhiiNosko, line 52: we have else for itemFuture that returns null so NPE is possible here

src/main/java/org/folio/service/pieces/ItemRecreateInventoryService.java
```diff
@@ -0,0 +1,81 @@
...
+        .compose(itemId -> deleteItem(piece, srcLocCtx).map(voidResult -> itemId));
+    } catch (Exception e) {
+      return Future.failedFuture(e);
```
SerhiiNosko, line 55: we can delete this catch, we will have just failed future in case of exception.

SerhiiNosko (approved):

PR #984, opened 2024-07-19
[MODORDERS-1154] 500 response when opening order

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1154] 500 response when opening order with 2 restricted funds and appropriate locations for each fund from Member tenant](https://folio-org.atlassian.net/browse/MODORDERS-1154)

## Approach
Check for existing sharing instances when opening order

## Verification
https://folio-org.atlassian.net/browse/MODORDERS-1154?focusedCommentId=221778


BKadirkhodjaev (approved):

src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
```diff
@@ -330,7 +334,7 @@ public Future<String> createInstanceRecord(Title title, RequestContext requestCo
 
     Future<Void> contributorNameTypeIdFuture = verifyContributorNameTypesExist(title.getContributors(), requestContext);
 
-    return CompositeFuture.join(instanceTypeFuture, statusFuture, contributorNameTypeIdFuture)
+    return GenericCompositeFuture.join(List.of(instanceTypeFuture, statusFuture, contributorNameTypeIdFuture))
```
SerhiiNosko, line 337: Just for my knowledge, it is just not changeі anything, just use the same GenericCompositeFuture as in other places?

SerhiiNosko: because we have 42 usages of 'CompositeFuture.join' and if it causes some issues we need to think about refactoring them as well

Saba-Zedginidze-EPAM: We use GenericCompositeFuture, the CompositeFuture is deprecated, just modified it as the IDE was giving warnings, should not have any different effect

src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
```diff
@@ -342,27 +346,19 @@ private Future<String> createInstance(JsonObject instanceRecJson, RequestContext
...
+      .map(compPOL::withInstanceId);
   }
 
   private Future<String> shareInstanceAmongTenantsIfNeeded(String instanceId, ConsortiumConfiguration consortiumConfiguration,
                                                            List<Location> locations, RequestContext requestContext) {
     return getTenantIdsWithoutShadowInstances(instanceId, consortiumConfiguration, locations, requestContext)
       .map(tenantIds -> tenantIds.stream()
-        .map(targetTenantId -> sharingInstanceService.createShadowInstance(instanceId, consortiumConfiguration,
-            RequestContextUtil.createContextWithNewTenantId(requestContext, targetTenantId)))
+        .map(tenantId -> createShadowInstanceIfNeeded(instanceId, consortiumConfiguration, createContextWithNewTenantId(requestContext, tenantId)))
```
Saba-Zedginidze-EPAM, line 361: Directly calling `createShadowInstance` caused the issue, as it would not check for existing instances and attempt to create one directly. Call should be `createShadowInstanceIfNeeded` that does all the necessary checks and finally calls `createShadowInstance` if needed

azizbekxm (approved):

SerhiiNosko (approved):

PR #985, opened 2024-07-22
MODORDERS-1149. 500 response when changing the instance connection with "Create new" option

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1149


src/main/java/org/folio/rest/impl/AcquisitionsUnitsApi.java
```diff
@@ -63,8 +63,8 @@ public void getAcquisitionsUnitsUnits(String query, String totalRecords, int off
 
     acquisitionsUnitsService.getAcquisitionsUnits(query, offset, limit, new RequestContext(vertxContext, okapiHeaders))
       .onSuccess(units -> {
-        if (logger.isInfoEnabled()) {
-          logger.info("Successfully retrieved acquisitions units: {}", JsonObject.mapFrom(units).encodePrettily());
+        if (logger.isDebugEnabled()) {
+          logger.debug("Successfully retrieved acquisitions units: {}", JsonObject.mapFrom(units).encodePrettily());
```
SerhiiNosko, line 67: remained only CREATE, UPDATE, DELETE operations with into level to not have noise in logs

PR #986, opened 2024-07-23
[MODORDERS-1154] Remove duplicate tenantId's while sharing

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1154] 500 response when opening order with 2 restricted funds and appropriate locations for each fund from Member tenant](https://folio-org.atlassian.net/browse/MODORDERS-1154)

## Approach
Remove duplicate tenantIds while sharing

### DIFF
--- a/src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
@@ -6,6 +6,7 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import one.util.streamex.StreamEx;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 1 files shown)


azizbekxm (approved):

src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
```diff
@@ -28,6 +29,7 @@
 import org.folio.service.consortium.SharingInstanceService;
 
 import java.util.Collection;
+import java.util.HashSet;
```
SerhiiNosko, line 32: 2 unused imports

Saba-Zedginidze-EPAM: Thanks, removed

SerhiiNosko (approved):

PR #990, opened 2024-08-05
[MODORDERS-1150-2]. Apply minor test refactoring and cleaning

BKadirkhodjaev:
## Purpose

- Apply minor test refactoring and cleaning (for better long-term maintenance)

## Approach

- Update comments, rename & refactor methods

### DIFF
--- a/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
+++ b/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
@@ -56,8 +56,10 @@ public class ManualCurrencyConversion extends AbstractCurrencyConversion {
   @Override
   public MonetaryAmount apply(MonetaryAmount amount) {
     // Add support for reverse conversion for cases when just a single manual exchange rate is set in the POL
-    // i.e. ECB dynamic rates would be: AUD -> USD will use ~0.66, USD -> AUD will use ~1.55
-    // but in case of manual exchange rate this second exchange rate is not provided
+    // for example ECB/IMF exchange rate for encumbrance transaction creation (AUD->USD) will be 0.651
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
```diff
@@ -56,8 +56,10 @@ public CurrencyConversion with(ConversionContext conversionContext) {
   @Override
   public MonetaryAmount apply(MonetaryAmount amount) {
     // Add support for reverse conversion for cases when just a single manual exchange rate is set in the POL
-    // i.e. ECB dynamic rates would be: AUD -> USD will use ~0.66, USD -> AUD will use ~1.55
-    // but in case of manual exchange rate this second exchange rate is not provided
+    // for example ECB/IMF exchange rate for encumbrance transaction creation (AUD->USD) will be 0.651
+    // while during rollover POL total amount (USD->AUD) will use 1.536 from the sum of all encumbrance transactions
+    // but in case of manual exchange rate set on the POL, this second exchange rate is not provided and during
+    // rollover total amount will be calculated using this same exchange rate
```
azizbekxm, line 62: seems it is a lot of info, probably need to move to java doc. if it needs for if statment, can be kepts as it was

azizbekxm (approved):

src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
BKadirkhodjaev: I agree, a Javadoc comment on the method would be more appropriate here

Saba-Zedginidze-EPAM (approved):

PR #991, opened 2024-08-13
[MODORDERS-1159]. ECS | Fix Fullscreen Receive Item Recreation In Destination Tenant

BKadirkhodjaev:
## Purpose

<https://folio-org.atlassian.net/browse/MODORDERS-1159>

## Approach

- Add item recreation during checkIn for Fullscreen Receive ECS workflow. This item recreation is independent from Quick Receive, but shares some of its methods. The reason for it to be independent is that Quick Receive performs item recreation during Piece update (`PUT /orders/pieces/{id}`) followed by CheckIn (`POST /orders/check-in`), while Fullscreen Receive does both actions from a single endpoint  (`POST /orders/check-in`) on multiple pieces. Item recreation should not be removed from Quick Receive in cases when the Piece is to be updated but not immediately received
- Tested by MockServer-based tests and mod-orders and cross-module Karate tests for potential regression


src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -82,29 +83,49 @@ private Future<ReceivingResults> processCheckInPieces(CheckinCollection checkinC
...
             if (checkInPiece.getId().equals(piece.getId()) && Boolean.TRUE.equals(checkInPiece.getCreateItem())) {
-              futures.add(purchaseOrderLineService.getOrderLineById(poLineId, requestContext)
+              pieceFutures.add(purchaseOrderLineService.getOrderLineById(poLineId, requestContext)
                 .map(PoLineCommonUtil::convertToCompositePoLine)
-                .compose(compPOL -> pieceCreateFlowInventoryManager.processInventory(compPOL, piece,
-                  checkInPiece.getCreateItem(), requestContext))
-                .map(aVoid -> Pair.of(poLineId, piece)));
+                .compose(compPol -> pieceCreateFlowInventoryManager.processInventory(compPol, piece, checkInPiece.getCreateItem(), requestContext))
+                .map(voidResult -> new PiecesHolder.PiecePoLineDto(poLineId, piece)));
+            } else if (checkInPiece.getId().equals(piece.getId()) && Objects.nonNull(srcTenantId) && Objects.nonNull(dstTenantId) && !srcTenantId.equals(dstTenantId)) {
```
SerhiiNosko, line 100: InventoryUtils.allowItemRecreate can be uased instead of Objects.nonNull(srcTenantId) && Objects.nonNull(dstTenantId) && !srcTenantId.equals(dstTenantId)

SerhiiNosko (approved):

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -82,29 +83,49 @@ private Future<ReceivingResults> processCheckInPieces(CheckinCollection checkinC
...
+    piecePoLineDtoList.stream()
+      .filter(PiecesHolder.PiecePoLineDto::isRecreateItem)
+      .forEach(dto -> itemRecreateDtoMap.put(dto.getPoLineId(), dto));
+    holder.withItemsToRecreate(itemRecreateDtoMap);
+  }
+
+  private static Map<String, List<Piece>> createPoLineIdPiecePair(List<PiecesHolder.PiecePoLineDto> piecePoLineDtoList) {
+    return StreamEx.of(piecePoLineDtoList)
+      .map(v2 -> Pair.of(v2.getPoLineId(), v2.getPieceFromStorage())).distinct()
+      .groupingBy(Pair::getKey, mapping(Pair::getValue, collectingAndThen(toList(), lists -> StreamEx.of(lists).collect(toList()))));
```
Saba-Zedginidze-EPAM, line 128: Could you clarify why the second parameter is used for `mapping()` method as `collectingAndThen(toList(), lists -> StreamEx.of(lists).collect(toList()))` and not `toList()` directly?

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -82,29 +83,49 @@ private Future<ReceivingResults> processCheckInPieces(CheckinCollection checkinC
...
+    var itemRecreateDtoMap = new HashMap<String, PiecesHolder.PiecePoLineDto>();
+    piecePoLineDtoList.stream()
+      .filter(PiecesHolder.PiecePoLineDto::isRecreateItem)
+      .forEach(dto -> itemRecreateDtoMap.put(dto.getPoLineId(), dto));
+    holder.withItemsToRecreate(itemRecreateDtoMap);
+  }
+
+  private static Map<String, List<Piece>> createPoLineIdPiecePair(List<PiecesHolder.PiecePoLineDto> piecePoLineDtoList) {
+    return StreamEx.of(piecePoLineDtoList)
+      .map(v2 -> Pair.of(v2.getPoLineId(), v2.getPieceFromStorage())).distinct()
```
Saba-Zedginidze-EPAM, line 127: What does `v2` stand for? Maybe `dto` could be a better name for this case

src/main/java/org/folio/helper/CheckinHelper.java
BKadirkhodjaev: > Could you clarify why the second parameter is used for `mapping()` method as `collectingAndThen(toList(), lists -> StreamEx.of(lists).collect(toList()))` and not `toList()` directly?

Yeah it's redundant, was there before, but I will improve it

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -483,11 +489,39 @@ protected Future<Map<String, List<Piece>>> updateInventoryItemsAndHoldings(Map<S
...
 
+  protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupedByPoLine,
+                                             PiecesHolder holder,
+                                             RequestContext requestContext) {
+    if (Objects.isNull(holder.getItemsToRecreate()) || holder.getItemsToRecreate().isEmpty()) {
+      return Future.succeededFuture();
+    }
+    var futures = new ArrayList<Future<String>>();
+    piecesGroupedByPoLine.keySet().stream()
+      .map(compPolId -> holder.getItemsToRecreate().get(compPolId))
```
Saba-Zedginidze-EPAM, line 506: Better to rename `comPolId` to `poLineId` as in other methods of the class I think, and lowercase L and 1 together could be confusing to read a bit sometimes

src/main/java/org/folio/service/inventory/InventoryUtils.java
```diff
@@ -272,4 +274,17 @@ public static void updateItemWithReceivedItemFields(JsonObject item, ReceivedIte
...
+    }
+    return new ItemRecreateConfig(receivingTenantId, createContextWithNewTenantId(requestContext, receivingTenantId));
+  }
+
+  public static boolean allowItemRecreate(ItemRecreateConfig srcConfig, ItemRecreateConfig dstConfig) {
+    return Objects.nonNull(srcConfig.tenantId()) && Objects.nonNull(dstConfig.tenantId())
+      && !srcConfig.tenantId().equals(dstConfig.tenantId());
+  }
+
+  public record ItemRecreateConfig(String tenantId, RequestContext context) {}
```
Saba-Zedginidze-EPAM, line 289: Correct me if I am wrong, but I think RequestContext already contains the tenant id which can be extracted. In other places different contexts are used to manage cross-tenant actions. I see these config records are only used for `allowItemRecreate` so maybe this could be cleaned up and this method could be modified to accept request contexts. What do you think?

src/main/java/org/folio/helper/CheckinHelper.java
```diff
@@ -82,29 +83,49 @@ private Future<ReceivingResults> processCheckInPieces(CheckinCollection checkinC
...
+              pieceFutures.add(Future.succeededFuture(new PiecesHolder.PiecePoLineDto(poLineId, piece)));
             }
           })));
         return null;
       })
-      .compose(v -> collectResultsOnSuccess(futures).map(poLineIdVsPieceList -> StreamEx.of(poLineIdVsPieceList)
-        .distinct()
-        .groupingBy(Pair::getKey, mapping(Pair::getValue, collectingAndThen(toList(), lists -> StreamEx.of(lists)
-          .collect(toList()))))));
+      .compose(v1 -> collectResultsOnSuccess(pieceFutures)
```
Saba-Zedginidze-EPAM, line 110: Here I guess this represents void, `v` would be better

src/main/java/org/folio/service/inventory/InventoryUtils.java
BKadirkhodjaev: > Correct me if I am wrong, but I think RequestContext already contains the tenant id which can be extracted. In other places different contexts are used to manage cross-tenant actions. I see these config records are only used for `allowItemRecreate` so maybe this could be cleaned up and this method could be modified to accept request contexts. What do you think?

No, this Record and its RequestContext is used to move item(s) from one tenant to the other, and not just for validation. It is  used downstream in context switching in both Quick and Fullscreen Receive.

Saba-Zedginidze-EPAM (approved):

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -483,11 +489,39 @@ protected Future<Map<String, List<Piece>>> updateInventoryItemsAndHoldings(Map<S
...
+          logger.info("recreateItemRecords:: recreating item by id '{}', srcTenantId: '{}', dstTenantId: '{}'", piece.getItemId(), srcConfig.tenantId(), dstConfig.tenantId());
+          futures.add(itemRecreateInventoryService.recreateItemInDestinationTenant(itemToRecreate.getCompositePoLine(), piece, srcConfig.context(), dstConfig.context()));
+        }
+      });
+    return collectResultsOnSuccess(futures)
+      .map(itemIdsRecreated -> {
+        itemIdsRecreated.forEach(itemIdRecreated -> logger.info("recreateItemRecords:: recreated item by id '{}'", itemIdRecreated));
+        return null;
+      })
+      .compose(v -> Future.succeededFuture());
```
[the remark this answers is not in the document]
BKadirkhodjaev: Seems like `.mapEmpty()` does not accept any args, and I don't want to use `.peek()` for this use case

PR #998, opened 2024-08-19
add extra composite purchase order param to use status for item

azizbekxm:
## Purpose
- when fetching purchase order there is issue with not found exception in case of member tenant context

## Approach
- added extra composite purchase order param

### DIFF
--- a/src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
+++ b/src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
@@ -506,39 +506,50 @@ public abstract class CheckinReceivePiecesHelper<T> extends BaseHelper {
     if (Objects.isNull(holder.getItemsToRecreate()) || holder.getItemsToRecreate().isEmpty()) {
       return Future.succeededFuture();
     }
-    var futures = new ArrayList<Future<String>>();
-    piecesGroupedByPoLine.keySet().stream()
-      .map(poLineId -> holder.getItemsToRecreate().get(poLineId))
-      .forEach(itemToRecreateList ->
-        itemToRecreateList.forEach(itemToRecreate -> {
-          var piece = itemToRecreate.getPieceFromStorage();
-          var checkInPiece  = itemToRecreate.getCheckInPiece();
-          var srcConfig = InventoryUtils.constructItemRecreateConfig(piece.getReceivingTenantId(), requestContext, true);
-          var dstConfig = InventoryUtils.constructItemRecreateConfig(checkInPiece.getReceivingTenantId(), requestContext, false);
-          if (InventoryUtils.allowItemRecreate(srcConfig.tenantId(), dstConfig.tenantId())) {
-            logger.info("recreateItemRecords:: recreating item by id '{}', srcTenantId: '{}', dstTenantId: '{}'", piece.getItemId(), srcConfig.tenantId(), dstConfig.tenantId());
 
-            // Fetch purchase order asynchronously
-            Future<CompositePurchaseOrder> compositePurchaseOrderFuture = purchaseOrderStorageService
-              .getCompositeOrderByPoLineId(itemToRecreate.getPoLineId(), requestContext)
-              .recover(throwable -> {
-                logger.warn("recreateItemRecords:: Composite Purchase order can't be found for poLine: {}", itemToRecreate.getPoLineId());
-                return Future.succeededFuture(null);
-              });
+    Map<String, Future<CompositePurchaseOrder>> purchaseOrderFutures = piecesGroupedByPoLine.keySet().stream()
+      .collect(Collectors.toMap(
+        poLineId -> poLineId,
+        poLineId -> purchaseOrderStorageService.getCompositeOrderByPoLineId(poLineId, requestContext)
+          .recover(throwable -> {
+            logger.warn("recreateItemRecords:: Composite Purchase order can't be found for poLine: {}", poLineId);
+            return Future.succeededFuture(null);
+          })
+      ));
 
-            Future<String> itemIdFuture = compositePurchaseOrderFuture
-              .compose(purchaseOrder -> itemRecreateInventoryService
-                .recreateItemInDestinationTenant(purchaseOrder, itemToRecreate.getCompositePoLine(),
-                  piece, srcConfig.context(), dstConfig.context()));
-            futures.add(itemIdFuture);
-          }
-        }));
-    return collectResultsOnSuccess(futures)
-      .map(itemIdsRecreated -> {
-        itemIdsRecreated.forEach(itemIdRecreated -> logger.info("recreateItemRecords:: recreated item by id '{}'", itemIdRecreated));
-        return null;
-      })
-      .compose(v -> Future.succeededFuture());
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -512,7 +517,17 @@ protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupe
           var dstConfig = InventoryUtils.constructItemRecreateConfig(checkInPiece.getReceivingTenantId(), requestContext, false);
           if (InventoryUtils.allowItemRecreate(srcConfig.tenantId(), dstConfig.tenantId())) {
             logger.info("recreateItemRecords:: recreating item by id '{}', srcTenantId: '{}', dstTenantId: '{}'", piece.getItemId(), srcConfig.tenantId(), dstConfig.tenantId());
-            futures.add(itemRecreateInventoryService.recreateItemInDestinationTenant(itemToRecreate.getCompositePoLine(), piece, srcConfig.context(), dstConfig.context()));
+
+            // Fetch purchase order asynchronously
+            Future<CompositePurchaseOrder> poFuture = purchaseOrderStorageService.getPurchaseOrderByIdAsJson(itemToRecreate.getPoLineId(), requestContext)
```
BKadirkhodjaev, line 522: I suggest you change the Reference type and variable name to `var purchaseOrderFuture = ...`

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -512,7 +517,17 @@ protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupe
...
+              .map(HelperUtils::convertToCompositePurchaseOrder)
+              .recover(throwable -> {
+                return Future.succeededFuture(null); // Return null to continue processing other requests
+              });
+
+            Future<String> itemFuture = poFuture.compose(purchaseOrder ->
```
BKadirkhodjaev, line 528: Similar renaming can apply

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -512,7 +517,17 @@ protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupe
...
+              itemRecreateInventoryService.recreateItemInDestinationTenant(purchaseOrder, itemToRecreate.getCompositePoLine(), piece, srcConfig.context(), dstConfig.context()));
+            futures.add(itemFuture);
```
BKadirkhodjaev, line 530: We can maybe name this as `itemIdFutures`

src/main/java/org/folio/orders/utils/PoLineCommonUtil.java
```diff
@@ -290,4 +290,8 @@ private static void checkFieldsChanged(JsonPathParser oldObject, JsonPathParser
       throw new HttpException(400, error, parameters);
     }
   }
+
+  public static CompositePoLine getPoLine(CompositePurchaseOrder compositePurchaseOrder) {
```
SerhiiNosko, line 294: need to delete this method from utils. It should not be commonly usagable. If in some places we need to get first po line - it is exceptional case, but not normal situation

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
SerhiiNosko, line 522: purchaseOrderStorageService already has method getCompositeOrderByPoLineId() that returns CompositePurchaseOrder

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -512,7 +517,17 @@ protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupe
           var dstConfig = InventoryUtils.constructItemRecreateConfig(checkInPiece.getReceivingTenantId(), requestContext, false);
           if (InventoryUtils.allowItemRecreate(srcConfig.tenantId(), dstConfig.tenantId())) {
             logger.info("recreateItemRecords:: recreating item by id '{}', srcTenantId: '{}', dstTenantId: '{}'", piece.getItemId(), srcConfig.tenantId(), dstConfig.tenantId());
-            futures.add(itemRecreateInventoryService.recreateItemInDestinationTenant(itemToRecreate.getCompositePoLine(), piece, srcConfig.context(), dstConfig.context()));
+
+            // Fetch purchase order asynchronously
+            Future<CompositePurchaseOrder> poFuture = purchaseOrderStorageService.getPurchaseOrderByIdAsJson(itemToRecreate.getPoLineId(), requestContext)
+              .map(HelperUtils::convertToCompositePurchaseOrder)
+              .recover(throwable -> {
+                return Future.succeededFuture(null); // Return null to continue processing other requests
```
SerhiiNosko, line 525: we don't have logs here, so it is not possible to find a reaons here. Alos this change does not relate to current story, we can make it if confident in this

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -512,7 +517,20 @@ protected Future<Void> recreateItemRecords(Map<String, List<Piece>> piecesGroupe
...
           if (InventoryUtils.allowItemRecreate(srcConfig.tenantId(), dstConfig.tenantId())) {
             logger.info("recreateItemRecords:: recreating item by id '{}', srcTenantId: '{}', dstTenantId: '{}'", piece.getItemId(), srcConfig.tenantId(), dstConfig.tenantId());
-            futures.add(itemRecreateInventoryService.recreateItemInDestinationTenant(itemToRecreate.getCompositePoLine(), piece, srcConfig.context(), dstConfig.context()));
+
+            // Fetch purchase order asynchronously
+            Future<CompositePurchaseOrder> compositePurchaseOrderFuture = purchaseOrderStorageService
+              .getCompositeOrderByPoLineId(itemToRecreate.getPoLineId(), requestContext)
+              .recover(throwable -> {
+                logger.warn("recreateItemRecords:: Composite Purchase order can't be found for poLine: {}", itemToRecreate.getPoLineId());
+                return Future.succeededFuture(null);
```
SerhiiNosko, line 526: formatting with 2 lines should be used( these 2 lines are inside of .recover block)

src/test/java/org/folio/rest/impl/CheckinReceivingApiTest.java
```diff
@@ -695,14 +696,15 @@ void testPostCheckInPhysicalFullyReceivedWithChangedLocation() {
 
   private static Stream<Arguments> testPostCheckInPhysicalFullyReceivedEcsArgs() {
     return Stream.of(
-      Arguments.of(12, "checkin-fully-receive-physical-resource-ecs-single.json", 1, 4),
-      Arguments.of(13, "checkin-fully-receive-physical-resource-ecs-multiple.json", 2, 7)
+      Arguments.of(12, "checkin-fully-receive-physical-resource-ecs-single.json", 1, 4, 1),
```
BKadirkhodjaev, line 699: Nice optimization

BKadirkhodjaev (approved):

PR #1005, opened 2024-08-23
[MODORDERS-1167]. POL statuses remain "Ongoing" after unopen ongoing order

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1167>

## Approach

- Add Ongoing to each status
- Apply additional minor refactoring
- Tested with existing mod-orders and cross-module Karate tests for conflict
- Will add a dedicated Karate tests for Unopening retrospectively


src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
```diff
@@ -107,10 +108,10 @@ private Future<CompositePurchaseOrder> updateAndGetOrderWithLines(CompositePurch
   }
 
   private void makePoLinePending(CompositePoLine poLine) {
-    if (poLine.getPaymentStatus() == CompositePoLine.PaymentStatus.AWAITING_PAYMENT) {
+    if (EnumSet.of(CompositePoLine.PaymentStatus.AWAITING_PAYMENT, CompositePoLine.PaymentStatus.ONGOING).contains(poLine.getPaymentStatus())) {
```
SerhiiNosko, line 111: is it possible to write some test here that both conditions are covered?

SerhiiNosko (approved):

src/main/java/org/folio/service/orders/flows/update/unopen/UnOpenCompositeOrderManager.java
BKadirkhodjaev: > is it possible to write some test here that both conditions are covered?

I'll add a Karate test next week to cover the whole Unopening scenario

Saba-Zedginidze-EPAM (approved):

PR #1007, opened 2024-08-26
[MODORDERS-1169]. Fix piece update on null processedHoldings.get(holdingKey)

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1169>

## Approach

- Add a condition to prevent setting a null `holdingId` in the map, which is used to update the processed pieces
- Remove item status validation during item recreation on receive with an affiliation change
- Make holding creation apply for every piece that has an incoming `locationId`, which requires a new holding to be made
- Simplify item recreation method to fetch an item from the source tenant before other operations
- Tested with mod-orders and cross-module Karate tests:
  - image
  - image

### DIFF
--- a/src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
+++ b/src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
@@ -399,10 +399,14 @@ public abstract class CheckinReceivePiecesHelper<T> extends BaseHelper {
     logger.info("calculatePoLineReceiptStatus:: Processed pieces: {}, Pieces from storage: {}", piecesSuccessfullyProcessed.size(), piecesByPoLine.size());
     piecesSuccessfullyProcessed.forEach(v -> logger.info("calculatePoLineReceiptStatus:: Processed Piece, id: {}, status: {}", v.getId(), v.getReceivingStatus()));
     piecesByPoLine.forEach(v -> logger.info("calculatePoLineReceiptStatus:: Piece from storage, id: {}, status: {}", v.getId(), v.getReceivingStatus()));
-    long expectedPiecesQuantity = piecesByPoLine.stream().filter(piece -> EXPECTED_STATUSES.contains(piece.getReceivingStatus())).count();
-    long receivedQty = piecesByPoLine.stream().filter(piece -> RECEIVED_STATUSES.contains(piece.getReceivingStatus())).count();
+    long expectedPiecesQuantity = piecesByPoLine.stream()
+      .filter(piece -> EXPECTED_STATUSES.contains(piece.getReceivingStatus()))
+      .count();
+    long receivedPiecesQuantity = piecesByPoLine.stream()
+      .filter(piece -> RECEIVED_STATUSES.contains(piece.getReceivingStatus()))
+      .count();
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -777,28 +767,15 @@ private Future<Map<String, List<Piece>>> processItemsUpdate(Map<String, List<Pie
...
       isHoldingUpdateRequired = PoLineCommonUtil.isHoldingUpdateRequiredForEresource(poLine);
     } else {
       isHoldingUpdateRequired = PoLineCommonUtil.isHoldingUpdateRequiredForPhysical(poLine);
     }
-    String locationId = getLocationId(piece);
-    String holdingId = getHoldingId(piece);
-    return isHoldingUpdateRequired && (StringUtils.isNotEmpty(locationId) || StringUtils.isNotEmpty(holdingId));
-  }
-
-  private String buildProcessedHoldingKey(Piece piece, String instanceId) {
```
BKadirkhodjaev, line 791: This method prevented holding creation if a new holding is requested by the user, so it was removed altogether

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -749,13 +738,14 @@ private Future<Map<String, List<Piece>>> processItemsUpdate(Map<String, List<Pie
...
 
-      if (holdingUpdateOnCheckinReceiveRequired(piece, poLine) && !isRevertToOnOrder(piece) && !processedHoldings.isEmpty()) {
-        String holdingKey = buildProcessedHoldingKey(piece, title.getInstanceId());
-        String holdingId = processedHoldings.get(holdingKey);
-        item.put(ITEM_HOLDINGS_RECORD_ID, holdingId);
-        logger.info("processItemsUpdate:: Item was processed, itemId: {}", piece.getItemId());
-      } else {
-        logger.info("processItemsUpdate:: Item processing is not required, deferred, itemId: {}", piece.getItemId());
+      if (holdingUpdateOnCheckinReceiveRequired(piece, poLine) && !isRevertToOnOrder(piece)) {
+        if (processedHoldings.containsKey(piece.getId())) {
```
BKadirkhodjaev, line 742: Here we check if the `holdingId` is not null/exists before setting it

src/main/java/org/folio/orders/utils/RequestContextUtil.java
```diff
@@ -18,7 +18,7 @@ public static RequestContext createContextWithNewTenantId(RequestContext request
     }
     var modifiedHeaders = new CaseInsensitiveMap<>(requestContext.getHeaders());
     modifiedHeaders.put(XOkapiHeaders.TENANT, tenantId);
-    logger.info("Request context has been changed with new tenant: {}", tenantId);
+    logger.debug("Request context has been changed with new tenant: {}", tenantId);
```
SerhiiNosko, line 21: as for me better to remain this log as info, because it shows at lest that me make requests in different tenants

SerhiiNosko (approved):

PR #1009, opened 2024-09-05
[MODORDERS-1177] A user cannot add/edit piece when related order has another pieces with affiliations in tenants user does not belong to

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1177

## Approach
Update only those locations that is related to the piece, but not all locations. It allows to reduce lot of http calls to fetch holdings for other not modified locations. Also it allowed to fix root cause of the issue because call to tenant where user does not have an affiliation is not made

### DIFF
--- a/src/main/java/org/folio/config/ApplicationConfig.java
+++ b/src/main/java/org/folio/config/ApplicationConfig.java
@@ -483,10 +483,8 @@ public class ApplicationConfig {
   InventoryItemManager inventoryItemManager(RestClient restClient,
                                             ConfigurationEntriesCache configurationEntriesCache,
                                             InventoryCache inventoryCache,
-                                            ConsortiumConfigurationService consortiumConfigurationService,
-                                            PurchaseOrderStorageService purchaseOrderStorageService) {
-    return new InventoryItemManager(restClient, configurationEntriesCache, inventoryCache,
-      consortiumConfigurationService, purchaseOrderStorageService);
+                                            ConsortiumConfigurationService consortiumConfigurationService) {
+    return new InventoryItemManager(restClient, configurationEntriesCache, inventoryCache, consortiumConfigurationService);
   }
 
   @Bean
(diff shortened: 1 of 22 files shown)


src/main/java/org/folio/service/pieces/PieceUpdateInventoryService.java
```diff
@@ -43,28 +42,19 @@ public PieceUpdateInventoryService(InventoryItemManager inventoryItemManager,
...
-          .onFailure(itemFuture::fail);
-      } else {
-        logger.warn("manualPieceFlowCreateItemRecord:: Creating Item is not possible for piece: {}, poLine: {}",
-          piece.getId(), compPOL.getId());
-        itemFuture.complete(null);
-      }
-    } catch (Exception e) {
-      logger.error("Error while creating item for piece:{} and comPOL: {}", piece.getId(), compPOL.getId(), e);
-      itemFuture.fail(e);
+    logger.info("manualPieceFlowCreateItemRecord:: Handling {} items for PO Line and holdings with id={}, receivingTenantId={}",
```
SerhiiNosko, line 45: it is just refactoring not releted to an issue

src/main/java/org/folio/config/ApplicationConfig.java
```diff
@@ -483,10 +483,8 @@ ProtectionService protectionHelper(AcquisitionsUnitsService acquisitionsUnitsSer
   InventoryItemManager inventoryItemManager(RestClient restClient,
                                             ConfigurationEntriesCache configurationEntriesCache,
                                             InventoryCache inventoryCache,
-                                            ConsortiumConfigurationService consortiumConfigurationService,
-                                            PurchaseOrderStorageService purchaseOrderStorageService) {
```
SerhiiNosko, line 487: purchaseOrderStorageService was not used in InventoryItemManager

BKadirkhodjaev (approved):

PR #1017, opened 2024-09-12
[MODORDERS-1174] Filter pieces based on user affiliations

Saba-Zedginidze-EPAM:
## Purpose

## Approach
- Fetch user tenants in case of ECS mode
- Filter pieces with user tenants if needed
- Update tests

#### Orders Karate Tests
image

### DIFF
--- a/src/main/java/org/folio/helper/BindHelper.java
+++ b/src/main/java/org/folio/helper/BindHelper.java
@@ -93,7 +93,7 @@ public class BindHelper extends CheckinReceivePiecesHelper<BindPiecesCollection>
 
   private Future<Void> clearTitleBindItemsIfNeeded(String titleId, String bindItemId, RequestContext requestContext) {
     String query = String.format("titleId==%s and bindItemId==%s and isBound==true", titleId, bindItemId);
-    return pieceStorageService.getAllPieces(query, requestContext)
+    return pieceStorageService.getAllPieces(0, 0, query, requestContext)
       .compose(pieceCollection -> {
         var totalRecords = pieceCollection.getTotalRecords();
         if (totalRecords != 0) {
--- a/src/main/java/org/folio/helper/BindHelper.java
+++ b/src/main/java/org/folio/helper/BindHelper.java
@@ -93,7 +93,7 @@ public class BindHelper extends CheckinReceivePiecesHelper<BindPiecesCollection>
 
   private Future<Void> clearTitleBindItemsIfNeeded(String titleId, String bindItemId, RequestContext requestContext) {
     String query = String.format("titleId==%s and bindItemId==%s and isBound==true", titleId, bindItemId);
-    return pieceStorageService.getPieces(0, 0, query, requestContext)
+    return pieceStorageService.getAllPieces(query, requestContext)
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/service/pieces/PieceStorageService.java
```diff
@@ -110,10 +119,26 @@ public Future<List<Piece>> getPiecesByHoldingId(String holdingId, RequestContext
...
 
   public Future<PieceCollection> getPieces(int limit, int offset, String query, RequestContext requestContext) {
-    RequestEntry requestEntry = new RequestEntry(PIECE_STORAGE_ENDPOINT).withQuery(query)
-      .withOffset(offset)
-      .withLimit(limit);
-    return restClient.get(requestEntry, PieceCollection.class, requestContext);
+    var requestEntry = new RequestEntry(PIECE_STORAGE_ENDPOINT).withQuery(query).withOffset(offset).withLimit(limit);
+    return restClient.get(requestEntry, PieceCollection.class, requestContext)
+      .compose(piecesCollection -> filterPiecesByUserTenantsIfNecessary(piecesCollection.getPieces(), requestContext))
+      .map(pieces -> new PieceCollection().withPieces(pieces).withTotalRecords(pieces.size()));
```
SerhiiNosko, line 125: should be piecesCollection.getTotal() instead of pieces.size(). It breaks pagination but at least for non ECS envs pagination will work

Saba-Zedginidze-EPAM: Good point. Pagination can be still kept valid in case we implement a mechanism to retry GET requests until we get desired number of pieces with correct offset parameters. Could be some work, or should we keep this as is?

SerhiiNosko: Probably let leave as is to make solution simple, we do not have pagination for pieces for now, and in case it would be some integration in ECS mode to retrieve pieces - they just need to have an affiliation in each tenant to bypass this, because there is no other clear way here.
Just need to change pieces.size() to total

src/main/java/org/folio/helper/BindHelper.java
```diff
@@ -93,7 +93,7 @@ private Future<Void> removeForbiddenEntities(Piece piece, RequestContext request
 
   private Future<Void> clearTitleBindItemsIfNeeded(String titleId, String bindItemId, RequestContext requestContext) {
     String query = String.format("titleId==%s and bindItemId==%s and isBound==true", titleId, bindItemId);
-    return pieceStorageService.getPieces(0, 0, query, requestContext)
+    return pieceStorageService.getAllPieces(query, requestContext)
```
SerhiiNosko, line 96: we just need this call to get totalRecords, but we do not need pieces, so in order to not load DB can use
getPieces(0, 0, query, requestContext)

Saba-Zedginidze-EPAM: Good catch, fixed

SerhiiNosko (approved):

azizbekxm (approved):

PR #1022, opened 2024-09-18
[MODORDERS-1162]. Order line encumbrance link not updated when re-encumber is false and order type not rolled over

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1162>

## Approach

- Modify `buildOpenOrClosedOrderQueryByFundIdsAndTypes` to build a query without including any encumbrance settings
- Perform slight `OrderRolloverService` refactoring 

## Related issues

- <https://github.com/folio-org/mod-finance-storage/pull/424>
- <https://github.com/folio-org/folio-integration-tests/pull/1536>

### DIFF
--- a/src/main/java/org/folio/service/orders/OrderRolloverService.java
+++ b/src/main/java/org/folio/service/orders/OrderRolloverService.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
@@ -114,7 +115,6 @@ public class OrderRolloverService {
         .compose(progress -> startRollover(ledgerFYRollover, progress, requestContext)));
   }
 
-
   public Future<Void> prepareRollover(LedgerFiscalYearRollover ledgerFYRollover, RequestContext requestContext) {
     return ledgerRolloverProgressService.getRolloversProgressByRolloverId(ledgerFYRollover.getId(), requestContext)
       .map(progress -> progress.withOrdersRolloverStatus(RolloverStatus.IN_PROGRESS))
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/service/orders/OrderRolloverService.java
```diff
@@ -144,15 +141,17 @@ private Future<Void> handleOrderRolloverError(Throwable t, LedgerFiscalYearRollo
...
-        List<Future<Void>> futures = new ArrayList<>();
+    var ctx =  requestContext.getContext();
+    return ctx
+      .<List<Future<Void>>>executeBlocking(promise -> {
+        var futures = new ArrayList<Future<Void>>();
         // process fundId chunks one by one
-        Semaphore semaphore = new Semaphore(1, requestContext.getContext().owner());
-        for (List<String> fundIds : fundIdChunks) {
+        var semaphore = new Semaphore(1, requestContext.getContext().owner());
+        for (var fundIds : fundIdChunks) {
```
SerhiiNosko, line 154: do you think that this formatting style is better then previous? for me it is the same, but let it be

BKadirkhodjaev: I just saw that there are 2 styles here in this method i.e. reference types & var together, I think everything inside should be either one or the other to make it more readable

src/main/java/org/folio/service/orders/OrderRolloverService.java
```diff
@@ -181,33 +180,29 @@ private void throwExceptionIfOneOfFailed(AsyncResult<Void> openOrdersResult, Asy
...
+    var query = buildOpenOrClosedOrderQueryByFundIdsAndTypes(chunkFundIds, workflowStatus, ledgerFYRollover);
     var totalRecordsFuture = purchaseOrderLineService.getOrderLineCollection(query, 0, 0, requestContext)
       .map(PoLineCollection::getTotalRecords);
-
     var ctx = requestContext.getContext();
-    return totalRecordsFuture.compose(totalRecords -> ctx.<List<Future<Void>>>executeBlocking(promise -> {
-        List<Future<Void>> futures = new ArrayList<>();
+    return totalRecordsFuture.compose(totalRecords ->
+      ctx.<List<Future<Void>>>executeBlocking(promise -> {
+        var futures = new ArrayList<Future<Void>>();
```
SerhiiNosko, line 190: if you think that previous code was not so redable we have util HelperUtils::executeWithSemaphores that encupsulate complexity, but we can remain as is

src/main/java/org/folio/service/orders/OrderRolloverService.java
```diff
@@ -431,37 +423,38 @@ private String buildQueryEncumbrancesForRollover(List<String> polineIds, LedgerF
...
+    resultQuery.append(" sortBy metadata.createdDate");
+    logger.info("buildOpenOrClosedOrderQueryByFundIdsAndTypes:: Resulting PO line query: {}", resultQuery);
+    return resultQuery.toString();
+  }
 
-    resultQuery = resultQuery + sortByCreatedDate;
-    logger.debug("buildOpenOrClosedOrderQueryByFundIdsAndTypes :: resulting PO line query: {}", resultQuery);
-
-    return resultQuery;
+  private static String buildFundIdQuery(List<String> fundIds) {
```
SerhiiNosko, line 445: can be non static, static better to use to utility classes only

azizbekxm (approved):

PR #1024, opened 2024-09-19
[MODORDERS-1174] Fix piece filtering for member tenants

Saba-Zedginidze-EPAM:
## Purpose

## Approach
- Fix piece filtering for member tenants
- Update tests

### DIFF
--- a/src/main/java/org/folio/service/pieces/PieceStorageService.java
+++ b/src/main/java/org/folio/service/pieces/PieceStorageService.java
@@ -25,6 +25,7 @@ import org.folio.rest.jaxrs.model.Piece;
 import org.folio.rest.jaxrs.model.PieceCollection;
 
 import io.vertx.core.Future;
+import org.folio.rest.tools.utils.TenantTool;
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/service/pieces/PieceStorageService.java
```diff
@@ -185,4 +187,9 @@ private Future<List<Piece>> getPieceChunkByLineIds(Collection<String> poLineIds,
       .map(PieceCollection::getPieces);
   }
 
+  private static boolean shouldFilterPiecesForTenant(String centralTenantId, RequestContext requestContext) {
+    var requestTenantId = TenantTool.tenantId(requestContext.getHeaders());
+    return requestTenantId.equals(centralTenantId);
```
BKadirkhodjaev, line 192: Is this to make pieces viewable only by the central tenant?

Saba-Zedginidze-EPAM: To match condition when to filter fetched pieces while calling GET all pieces endpoint

SerhiiNosko (approved):

PR #1025, opened 2024-09-19
[MODORDERS-1143] Paying an invoice reopens the order

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1143] Paying an invoice reopens the order](https://folio-org.atlassian.net/browse/MODORDERS-1143)

## Approach
- Minor refactoring for exising helper code
- Prevent order reopen when playing invoice

### OrdersApiTest
image
### ConsortiaOrdersApiTest
image
### CrossModulesApiTest
image


src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -204,48 +204,44 @@ public Future<CompositePoLine> createPoLine(CompositePoLine compPOL, JsonObject
...
-        .map(jsonConfig -> {
-          updateCreateInventory(compPOL, jsonConfig);
-          return null;
-        });
+    Promise<JsonObject> promise = Promise.promise();
+    if (tenantConfiguration != null && !tenantConfiguration.isEmpty() && !StringUtils.isEmpty(tenantConfiguration.getString(CREATE_INVENTORY))) {
+      promise.complete(new JsonObject(tenantConfiguration.getString(CREATE_INVENTORY)));
     } else {
-      return Future.succeededFuture();
+      promise.complete(new JsonObject());
```
SerhiiNosko, line 238: if you started refacroting here, we can improve even more by deleting promise usage

src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -204,48 +204,44 @@ public Future<CompositePoLine> createPoLine(CompositePoLine compPOL, JsonObject
...
-      } else {
-        promise.complete(new JsonObject());
-      }
-      return promise.future()
-        .map(jsonConfig -> {
-          updateCreateInventory(compPOL, jsonConfig);
-          return null;
-        });
+    Promise<JsonObject> promise = Promise.promise();
+    if (tenantConfiguration != null && !tenantConfiguration.isEmpty() && !StringUtils.isEmpty(tenantConfiguration.getString(CREATE_INVENTORY))) {
```
SerhiiNosko, line 235: StringUtils.isNotEmpty() easier to read then !StringUtils.isEmpty()

src/main/java/org/folio/orders/utils/HelperUtils.java
```diff
@@ -393,59 +386,16 @@ public static CompositePurchaseOrder convertToCompositePurchaseOrder(JsonObject
...
-    return purchaseOrder.getWorkflowStatus() == PurchaseOrder.WorkflowStatus.CLOSED
-      && poLines.stream().anyMatch(line -> !isCompletedPoLine(line));
-  }
-
-  private static boolean isCompletedPoLine(PoLine line) {
-    PoLine.PaymentStatus paymentStatus = line.getPaymentStatus();
-    PoLine.ReceiptStatus receiptStatus = line.getReceiptStatus();
-    return (paymentStatus == PAYMENT_NOT_REQUIRED || paymentStatus == FULLY_PAID || paymentStatus == PaymentStatus.CANCELLED)
-      && (receiptStatus == FULLY_RECEIVED || receiptStatus == RECEIPT_NOT_REQUIRED || receiptStatus == ReceiptStatus.CANCELLED);
+  public static CompositePoLine convertToCompositePoLine(PoLine poLine) {
```
azizbekxm, line 394: we already have this method you can use it and modify if need. https://github.com/folio-org/mod-orders/blob/7bd52b48dbab62b64e1742eabb5609dd95c46b8d/src/main/java/org/folio/orders/utils/PoLineCommonUtil.java#L253

src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -672,47 +626,43 @@ private Future<String> generateLineNumber(CompositePurchaseOrder compOrder, Requ
...
+      });
   }
 
-  private CompositePurchaseOrder addLineToCompOrder(CompositePurchaseOrder compOrder, JsonObject lineFromStorage) {
-    PoLine poLine = lineFromStorage.mapTo(PoLine.class);
-    poLine.setAlerts(null);
-    poLine.setReportingCodes(null);
-    CompositePoLine compPoLine = JsonObject.mapFrom(poLine).mapTo(CompositePoLine.class);
+  private CompositePurchaseOrder addLineToCompOrder(CompositePurchaseOrder compOrder, PoLine lineFromStorage) {
+    var compPoLine = convertToCompositePoLine(lineFromStorage);
```
SerhiiNosko, line 648: do you think it is not needed?
poLine.setAlerts(null);
poLine.setReportingCodes(null);
Maybe it was specially set to null?

SerhiiNosko: I see, they are deleted in convertToCompositePoLine()

src/main/java/org/folio/orders/utils/HelperUtils.java
Saba-Zedginidze-EPAM: I see this one modifies the poLine itself and used in many places, probably having alternate one that does not modify original object could be sensible, how do you think?

src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -724,36 +674,28 @@ private Future<Void> processPoLineEncumbrances(CompositePurchaseOrder compOrder,
...
-          List<PoLine> notCanceledPoLines = poLines.stream().filter(Predicate.not(this::isDbPoLineStatusCancelled)).toList();
-          if (CollectionUtils.isEmpty(notCanceledPoLines)) {
-            logger.info("updateInventoryItemStatus:: not canceled po lines not found, returning...");
-            return Future.succeededFuture(null);
-          }
-          return itemStatusSyncService.updateInventoryItemStatus(compOrderLine.getId(), compOrderLine.getLocations(),
-            ItemStatus.ORDER_CLOSED, requestContext);
-        });
+  private Future<Void> updateInventoryItemStatus(CompositePoLine compOrderLine, PoLine lineFromStorage, RequestContext requestContext) {
+    if (!isStatusChanged(compOrderLine, lineFromStorage) || !isStatusCanceledCompositePoLine(compOrderLine)) {
```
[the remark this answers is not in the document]
Saba-Zedginidze-EPAM: Wrote this way to prevent indentation. Will make it more clear

src/main/java/org/folio/orders/utils/HelperUtils.java
azizbekxm: Didn't notice, It is strange behaviour if one method will modify actual object and other don't. I recommend to keep only one and lets not modify actual object, so move your logic to existing one and remove current one.

SerhiiNosko (approved):

src/main/java/org/folio/orders/utils/HelperUtils.java
Saba-Zedginidze-EPAM: Could there be some dependency for removing those values? We could keep the old one and use new one in the future

azizbekxm: not sure, but I assume, changing old one to new one should affect to anywhere, since it is converting method. if it is, there is big problem old method. from my opinion, this thing only two of us know about situation but no one know when using these methods, leading to using old one because of located in PoLineCommonUtil. However, if you still want to keep both, please add some java doc to explain about method (not recommended from my side).

Saba-Zedginidze-EPAM: Sure I see your point. Best would be to remove it and I will check if doing it will affect anything

PR #1026, opened 2024-09-23
[MODORDERS-1179]. Take in consideration only pieces where user has an affilition during POL receipt status calculation

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1179

This PR contains changes to speedup rollover process also


src/main/java/org/folio/service/orders/OrderRolloverService.java
```diff
@@ -217,7 +217,7 @@ private Future<Void> rolloverOrdersByFundIds(List<String> chunkFundIds, LedgerFi
   private Future<Void> saveOrderLines(List<PoLine> orderLines, LedgerFiscalYearRollover ledgerFYRollover,
                                       PurchaseOrder.WorkflowStatus workflowStatus, RequestContext requestContext) {
     logger.info("saveOrderLines:: Saving POLs after rollover processing, size: {}", orderLines.size());
-    return purchaseOrderLineService.saveOrderLines(orderLines, requestContext)
+    return purchaseOrderLineService.saveOrderLinesWithoutSearchLocationsUpdate(orderLines, requestContext)
```
SerhiiNosko, line 220: performed some optimization to avoid so many requests to fetch holdings to update searchLocationId, because rollover does not update locations in Po Line, so we don't need so many new calls.
Related ticket https://folio-org.atlassian.net/browse/MODORDSTOR-419

src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
```diff
@@ -130,22 +131,54 @@ public Future<Void> saveOrderLine(CompositePoLine compositePoLine, List<Location
...
-    return GenericCompositeFuture.join(poLineCollections.stream()
-      .map(poLineCollection -> saveOrderLines(poLineCollection, requestContext))
-      .toList())
+  private Future<Void> saveOrderLinesCollections(List<PoLineCollection> poLineCollections, RequestContext requestContext) {
+    return GenericCompositeFuture.join(poLineCollections
+      .stream()
+      .map(poLineCollection -> {
+        logger.info("saveOrderLines:: start saving {} po lines in batch", poLineCollection.getTotalRecords());
+        RequestEntry requestEntry = new RequestEntry(BATCH_ENDPOINT);
+        return restClient.put(requestEntry, poLineCollection, requestContext);
```
azizbekxm, line 180: I think we can put poLineCollections to one poLineCollections to avoid multiple requests by creating some method before sending request
```
       List<PoLine> combinedPoLines = poLIneCollections.stream().flatMap(collections -> collections.getPolines.stream())
       .toList() 
        int totalRecords = combinedPoLines.size();
        return new PoLineCollections(combinedPoLines, totalRecords);
```

src/main/java/org/folio/helper/CheckinReceivePiecesHelper.java
```diff
@@ -415,7 +423,8 @@ private PoLine updateRelatedPoLineDetails(PoLine poLine,
       updatedLocations++;
     }
 
-    return updatedLocations > 0 ? PoLineCommonUtil.convertToPoLine(compositePoLine) : poLine;
+    return updatedLocations > 0 ? PoLineLocationsPair.of(PoLineCommonUtil.convertToPoLine(compositePoLine), locations.stream().toList()) :
```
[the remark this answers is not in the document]
SerhiiNosko: improved

src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
SerhiiNosko: it was an intent to make partitioning here to update each 100 po lines at a time, see method getPartitionedPoLines()

BKadirkhodjaev (approved):

src/main/java/org/folio/service/orders/PurchaseOrderLineService.java
azizbekxm: Ok, I see

azizbekxm (approved):

PR #1028, opened 2024-09-30
[MODORDERS-1174] Use CQL query to filter pieces based on user affiliations

Saba-Zedginidze-EPAM:
## Purpose

## Approach
- Replace filtering with CQL query after fetching user tenants first
- Add SettingsRetriever to filter pieces only when central ordering setting is enabled
- Update tests

### CrossModulesApiTest
image

### OrdersApiTest
image

### ConsortiaOrdersApiTest
image


src/main/java/org/folio/service/pieces/PieceStorageService.java
```diff
@@ -187,6 +179,15 @@ private Future<List<Piece>> getPieceChunkByLineIds(Collection<String> poLineIds,
       .map(PieceCollection::getPieces);
   }
 
+  static String getQueryForUserTenants(List<String> userTenants, String query) {
+    if (userTenants == null) {
+      return query;
+    }
+    userTenants.add(null);
+    var cql = convertIdsToCqlQuery(userTenants, "receivingTenantId");
```
SerhiiNosko, line 187: If we have tenantId length for example 20 symbols, the calculation suggests that approximately 103 strings receivingTenantId=<<tenantId>>, each with a length of 20 characters, can be included in GET request without exceeding the 4KB URL limit.
For the largest consortia we have 61 tenant, so we should have some capacity for the original 'query'

SerhiiNosko: but to be on more safe side we can add checking setting into settings table(because not all consortia can use central ordering): ALLOW_ORDERING_WITH_AFFILIATED_LOCATIONS
Here is how this setting is set on UI: https://folio-dev-thunderjet-consortium.ci.folio.org/settings/consortia-settings/central-ordering
This also can speedup the sql query because this query would be more simple without checking receivingTenantId
This setting should use cache, it can not be disabled when it was enabled

Saba-Zedginidze-EPAM: Thanks, now filters based on the setting

src/main/java/org/folio/service/pieces/PieceStorageService.java
```diff
@@ -134,21 +137,19 @@ public Future<PieceCollection> getAllPieces(int limit, int offset, String query,
...
-      .filter(piece -> Optional.ofNullable(piece.getReceivingTenantId())
-        .map(userTenants::contains)
-        .orElse(true))
-      .toList();
+      .compose(consortiumConfiguration -> {
+        if (consortiumConfiguration.isEmpty()) {
+          return Future.succeededFuture();
+        }
+        var configuration = consortiumConfiguration.get();
+        return settingsRetriever.getSettingByKey(SettingKey.CENTRAL_ORDERING_ENABLED, requestContext)
```
SerhiiNosko, line 147: Module descriptors of endpoints that use getPieces should be updated with the new module permission

SerhiiNosko (approved):

src/main/java/org/folio/service/pieces/PieceStorageService.java
```diff
@@ -187,9 +190,19 @@ private Future<List<Piece>> getPieceChunkByLineIds(Collection<String> poLineIds,
...
-  private static boolean shouldFilterPiecesForTenant(String centralTenantId, RequestContext requestContext) {
+  static String getQueryForUserTenants(List<String> userTenants, String query) {
+    if (CollectionUtils.isEmpty(userTenants)) {
+      return query;
+    }
+    var cql = convertIdsToCqlQuery(userTenants, "receivingTenantId");
+    var matchNullTenantExpr = getCqlExpressionForFieldNullValue("receivingTenantId");
+    cql = combineCqlExpressions("or", cql, matchNullTenantExpr);
+    return combineCqlExpressions("and", query, cql);
+  }
```
azizbekxm, line 201: this part is a little bit hard to follow. Suggest:
```
static String getQueryForUserTenants(List<String> userTenants, String query) {
    if (CollectionUtils.isEmpty(userTenants)) {
        return query;
    }

    String tenantCql = convertIdsToCqlQuery(userTenants, "receivingTenantId");
    String nullTenantCql = getCqlExpressionForFieldNullValue("receivingTenantId");
    String combinedTenantCql = combineCqlExpressions("or", tenantCql, nullTenantCql);

    return combineCqlExpressions("and", query, combinedTenantCql);
}
```

src/main/java/org/folio/service/settings/util/SettingFields.java
```diff
@@ -0,0 +1,17 @@
...
+
+  private final String value;
+
+  SettingFields(String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
+    return value;
+  }
```
azizbekxm, line 15: we can use Lombok in here if possible
@Getter
@RequiredArgsConstructor

pom.xml
```diff
@@ -588,6 +588,10 @@
           <showWeaveInfo>true</showWeaveInfo>
           <forceAjcCompile>true</forceAjcCompile>
           <aspectDirectory>src/main/java/org/folio/rest/annotations</aspectDirectory>
+          <includes>
+            <include>**/impl/*.java</include>
+            <include>**/*.aj</include>
+          </includes>
```
Saba-Zedginidze-EPAM, line 594: Copied from mod-orders-storage as building with lombok annotations was failing

src/main/java/org/folio/service/pieces/PieceStorageService.java
Saba-Zedginidze-EPAM: Both comments addressed

azizbekxm (approved):

PR #1029, opened 2024-10-02
[MODORDERS-1170] Order not auto-closing as complete despite not receiving/payment requirements

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1170

Karate test: https://github.com/folio-org/folio-integration-tests/pull/1549


src/test/java/org/folio/orders/utils/StatusUtilsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Test for StatusUtils.
+ * Generated with Copilot.
```
SerhiiNosko, line 18: can use it for our unit tests to gather statistic.

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -273,25 +273,17 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
...
-
-    if (isEmpty(compPO.getCompositePoLines())) {
-      purchaseOrderLineService.getPoLinesByOrderId(compPO.getId(), requestContext)
-        .onSuccess(promise::complete)
-        .onFailure(promise::fail);
-    } else {
-      Future.succeededFuture()
-        .map(v -> {
-          List<PoLine> poLines = PoLineCommonUtil.convertToPoLines(compPO.getCompositePoLines());
-          if (initialOrdersStatus.equals(compPO.getWorkflowStatus().value()))
```
SerhiiNosko, line 286: this if statement was a root cause, it prevented to this transitions from Open to Closed status

src/main/java/org/folio/service/orders/utils/StatusUtils.java
```diff
@@ -28,16 +28,22 @@ public static boolean areAllPoLinesCanceled(List<PoLine> poLines) {
     return poLines.stream().allMatch(StatusUtils::isStatusCanceledPoLine);
   }
 
-  public static boolean changeOrderStatus(PurchaseOrder purchaseOrder, List<PoLine> poLines) {
+  public static boolean changeOrderStatusForOrderUpdate(PurchaseOrder purchaseOrder, List<PoLine> poLines) {
```
SerhiiNosko, line 31: we should not transit from Closed to Open workflowStatus in case when using API even if PO Line has some uncomplete status, related karate test: https://github.com/folio-org/folio-integration-tests/blob/e0ef0143cd7c72ab2b61fcbb933615169a7470b7/acquisitions/src/main/resources/thunderjet/mod-orders/features/close-order-including-lines.feature#L4

SerhiiNosko: UI for closing the order never includes composite PO Lines, so deleting toBeReopened() has not influence in UI, but API after this change become more transperent

src/main/java/org/folio/service/orders/utils/StatusUtils.java
```diff
@@ -28,16 +28,22 @@ public static boolean areAllPoLinesCanceled(List<PoLine> poLines) {
...
+  }
+
+  public static boolean changeOrderStatusForPoLineUpdate(PurchaseOrder purchaseOrder, List<PoLine> poLines) {
+    if (toBeCancelled(purchaseOrder, poLines)) {
+      return closeOrder(purchaseOrder, REASON_CANCELLED);
+    }
+    if (toBeClosed(purchaseOrder, poLines)) {
+      return closeOrder(purchaseOrder, REASON_COMPLETE);
     }
     if (toBeReopened(purchaseOrder, poLines)) {
```
SerhiiNosko, line 48: we need to remain toBeReopened() from PO Line update, in this case for example when some PO Line paymentStatus changed from FULLY_PAID to AWAITING_PAYMENT we should reopen the order. also related ticket: https://folio-org.atlassian.net/browse/MODORDERS-1143

Saba-Zedginidze-EPAM (approved):

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -273,25 +273,17 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
...
 
-    return promise.future()
+    return Future.succeededFuture()
+      .compose(v -> {
+        if (isEmpty(compPO.getCompositePoLines())) {
+          return purchaseOrderLineService.getPoLinesByOrderId(compPO.getId(), requestContext);
+        } else {
+          List<PoLine> poLines = PoLineCommonUtil.convertToPoLines(compPO.getCompositePoLines());
+          changeOrderStatusForOrderUpdate(purchaseOrder, poLines);
+          return Future.succeededFuture(poLines);
```
azizbekxm, line 284: a little confusion based on method name. if poLines exists in comPO we are chaning status but if not we are skipping. is it possible that fetched poLines have undesired status

azizbekxm (approved):

src/main/java/org/folio/helper/PurchaseOrderHelper.java
SerhiiNosko: agree, it is not very obvious written initially, for example we use API to  GET composite order with lines and update workflowStatus for example to Closed with the same payload as we fetched from GET. In this case both order and lines will be included into PUT request.
But when UI is making PUT wiht Closed status it does not send po lines.
So decided to stop refactoring here for now to not introduce new issues.

PR #1034, opened 2024-10-10
[MODORDERS-1190] Fund distribution should not be changed in invoice line (invoice is in "Open" or "Reviewed" status) when fund was changed in related open POL

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1190

## Approach
Delete code that that changes fund distributions in invoice lines after editing po line's fund distributions that is not needed any more. We just need to delete encumbrance link in invoice line fund distribution.

## Karate tests
Related karate PR: https://github.com/folio-org/folio-integration-tests/pull/1560

### OrdersKarate
There is 1 test that failes always locally because of rounding issue
orders_karate
### CrossModulesKarate
cross_,odules


src/main/java/org/folio/service/invoice/InvoiceLineService.java
```diff
@@ -93,7 +98,12 @@ public Future<Void> removeEncumbranceLinks(List<InvoiceLine> invoiceLines, List<
         }
       }
     }
-    return saveInvoiceLines(invoiceLinesToUpdate, requestContext);
+    if (CollectionUtils.isNotEmpty(invoiceLinesToUpdate)) {
+      logger.info("removeEncumbranceLinks:: updating invoice lines: {} with removed encumbrance links",
+        invoiceLinesToUpdate.stream().map(InvoiceLine::getId).collect(joining(", ")));
+      return saveInvoiceLines(invoiceLinesToUpdate, requestContext);
+    }
+    return Future.succeededFuture();
```
azizbekxm, line 106: very small suggestion. 
    if (CollectionUtils.isEmpty(invoiceLinesToUpdate)) {
      logger.info("removeEncumbranceLinks:: no invoice lines to update");
    }
    logger.info("removeEncumbranceLinks:: updating invoice lines: {} with removed encumbrance links",
        invoiceLinesToUpdate.stream().map(InvoiceLine::getId).collect(joining(", ")));
    return saveInvoiceLines(invoiceLinesToUpdate, requestContext);

azizbekxm (approved):

Saba-Zedginidze-EPAM (approved):

src/main/java/org/folio/service/invoice/InvoiceLineService.java
SerhiiNosko: thanks, applied

PR #1055, opened 2024-12-02
[MODORDERS-1210] Improve validation for piece batch status update

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1210] Implement API to change piece statuses in batch](https://folio-org.atlassian.net/browse/MODORDERS-1210)

## Approach
- First validate operation restriction to avoid unnecessary API call
- Add fetched piece quantity validation
- Changed error response in case piece is not found
```
{
  "errors": [
    {
      "message": "The piece record is not found",
      "code": "pieceNotFound",
      "parameters": [
        {
          "key": "pieceIds",
          "value": "[479b2177-3dc8-46aa-8cbc-81d0a1456242]"
        }
      ]
    }
  ],
  "total_records": 1
}
```

### DIFF
--- a/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
+++ b/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import lombok.extern.log4j.Log4j2;
+import org.folio.HttpStatus;
 import org.folio.models.pieces.BasePieceFlowHolder;
 import org.folio.models.pieces.PieceBatchStatusUpdateHolder;
 import org.folio.models.pieces.PieceUpdateHolder;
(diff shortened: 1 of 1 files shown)


BKadirkhodjaev (approved):

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
```diff
@@ -159,14 +163,19 @@ private List<Location> getPieceLocations(List<Piece> pieces, CompositePoLine poL
...
         .toList())
       .map(GenericCompositeFuture::all)
-      .map(pieces);
+      .mapEmpty();
+  }
+
+  private Future<List<Piece>> validateFetchedPiecesQuantity(List<Piece> pieces, List<String> pieceIds) {
+    return pieces.size() == pieceIds.size()
+      ? Future.succeededFuture(pieces)
+      : Future.failedFuture(new HttpException(HttpStatus.HTTP_BAD_REQUEST.toInt(), ErrorCodes.PIECE_NOT_FOUND));
```
azizbekxm, line 178: better to add which pieces not found or just return all pieces  as a param

azizbekxm (approved):

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
Saba-Zedginidze-EPAM: Good suggestion, included

PR #1057, opened 2024-12-02
[EUREKA-561-2]. (ECS) Can open/unopen order with PO line locations from tenant in which user does not have affiliation

BKadirkhodjaev:
## Purpose

## Approach

- Add more logging for better observability and debugging

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
@@ -742,14 +742,17 @@ public class PurchaseOrderLineHelper {
         }
         var storageUnaffiliatedLocations = extractUnaffiliatedLocations(storedPoLine.getLocations(), userTenants);
         var updatedUnaffiliatedLocations = extractUnaffiliatedLocations(updatedPoLine.getLocations(), userTenants);
-        logger.info("validateUserUnaffiliatedLocationUpdates:: Found unaffiliated location tenant ids: stored: '{}', updated: '{}'",
-          storageUnaffiliatedLocations.stream().map(Location::getTenantId).toList(),
-          updatedUnaffiliatedLocations.stream().map(Location::getTenantId).toList());
+        logger.info("validateUserUnaffiliatedLocationUpdates:: Found unaffiliated POL location tenant ids, poLineId: '{}', stored: '{}', updated: '{}'",
+          updatedPoLine.getId(),
+          storageUnaffiliatedLocations.stream().map(Location::getTenantId).distinct().toList(),
+          updatedUnaffiliatedLocations.stream().map(Location::getTenantId).distinct().toList());
         if (!SetUtils.isEqualSet(storageUnaffiliatedLocations, updatedUnaffiliatedLocations)) {
-          logger.info("validateUserUnaffiliatedLocationUpdates:: User is not affiliated with all locations on the POL");
+          logger.info("validateUserUnaffiliatedLocationUpdates:: User is not affiliated with all locations on the POL, poLineId: '{}'",
+            updatedPoLine.getId());
           return Future.failedFuture(new HttpException(422, ErrorCodes.LOCATION_UPDATE_WITHOUT_AFFILIATION));
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -736,13 +737,19 @@ private Future<Void> validateUserUnaffiliatedLocationUpdates(CompositePoLine upd
...
+          logger.info("validateUserUnaffiliatedLocationUpdates:: User tenants is empty");
           return Future.succeededFuture();
         }
         var storageUnaffiliatedLocations = extractUnaffiliatedLocations(storedPoLine.getLocations(), userTenants);
         var updatedUnaffiliatedLocations = extractUnaffiliatedLocations(updatedPoLine.getLocations(), userTenants);
+        logger.info("validateUserUnaffiliatedLocationUpdates:: Found unaffiliated location tenant ids: stored: '{}', updated: '{}'",
+          storageUnaffiliatedLocations.stream().map(Location::getTenantId).toList(),
+          updatedUnaffiliatedLocations.stream().map(Location::getTenantId).toList());
         if (!SetUtils.isEqualSet(storageUnaffiliatedLocations, updatedUnaffiliatedLocations)) {
+          logger.info("validateUserUnaffiliatedLocationUpdates:: User is not affiliated with all locations on the POL");
```
SerhiiNosko, line 749: better to add some identifier because for some cases it can be problematic to find correct logs. Propose to add updatedPoLine.getId().
In this case for example QA just can get us po line id, and we can use it in search to find all information what is going on with that po line id.

src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
```diff
@@ -829,6 +836,7 @@ private Future<Void> verifyDeleteAllowed(PoLine line, RequestContext requestCont
   }
 
   private Future<Void> createShadowInstanceIfNeeded(CompositePoLine compositePoLine, RequestContext requestContext) {
+    logger.info("createShadowInstanceIfNeeded:: Creating shadow instance if needed");
```
SerhiiNosko, line 839: we have full logs into inventoryInstanceManager.createShadowInstanceIfNeeded that is invoked under the hood.
We can add that log into if statement instead.

src/main/java/org/folio/service/ProtectionService.java
```diff
@@ -69,6 +69,7 @@ public Future<Void> isOperationRestricted(List<String> unitIds, ProtectedOperati
    */
   public Future<Void> isOperationRestricted(List<String> unitIds, Set<ProtectedOperationType> operations, RequestContext requestContext) {
     if (userHasDesiredPermission(BYPASS_ACQ_UNITS, requestContext)) {
+      log.info("isOperationRestricted:: No desired permissions found");
```
SerhiiNosko, line 72: propose to rename log to "bypass checking acq units"

src/main/java/org/folio/service/consortium/ConsortiumConfigurationService.java
```diff
@@ -63,12 +63,12 @@ private CompletableFuture<Optional<ConsortiumConfiguration>> getConsortiumConfig
       .map(jsonObject -> jsonObject.getJsonArray(USER_TENANTS_ARRAY_IDENTIFIER))
       .map(userTenants -> {
         if (userTenants.isEmpty()) {
-          logger.debug("Central tenant and consortium id not found");
+          logger.info("getConsortiumConfigurationFromRemote:: Central tenant and consortium id not found");
```
SerhiiNosko, line 66: for non ecs setup this log would be everywhere and support during investigating of RRT issues can think that something bad is happening by reading this message

BKadirkhodjaev: > for non ecs setup this log would be everywhere and support during investigating of RRT issues can think that something bad is happening by reading this message

I will change it back to debug mode

SerhiiNosko (approved):

PR #1060, opened 2024-12-05
MODORDERS-1215 Adjust permissions for system user

dmytrokrutii:
* [MODORDERS-1215](https://issues.folio.org/browse/MODORDERS-1215) Adjust permissions for system user

To support DI operation in Eureka, we should implement support for performing asynchronous operations on behalf of a system user. 

## Approach
Handle data-import events with null or empty token


JavokhirAbdullayev (approved):

KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/service/dataimport/utils/DataImportUtils.java
```diff
@@ -33,4 +39,36 @@ public static Map<String, String> extractOkapiHeaders(DataImportEventPayload eve
     }
     return headers;
   }
+
+  public static Map<String, String> kafkaHeadersToMap(KafkaConsumerRecord<String, String> kafkaRecord) {
+    var headerStream = kafkaRecord.headers().stream();
+    if (!isSystemUserEnabled()) {
```
RuslanLavrov, line 45: if I’m not confused then here we should use condition (isSystemUserEnabled()) (without `!` operator), correct? 
Because otherwise we will get headers without token when system user is disabled.

descriptors/ModuleDescriptor-template.json
```diff
@@ -2243,6 +2243,32 @@
       "description" : "Permission should be provided to reopen order"
     }
   ],
+  "metadata": {
+    "user": {
+      "type": "system",
+      "permissions": [
```
RuslanLavrov, line 2249: please add also these permissions:
"organizations.organizations.collection.get"
"change-manager.jobExecutions.item.get"

these permissions are not required:
"invoice-storage.invoice-line-number.get",
"invoice-storage.invoice-lines.item.post",
"invoice-storage.invoice-lines.item.put",
"invoice-storage.invoice-lines.collection.get",
"invoice-storage.invoice-number.get",
"invoice-storage.invoices.item.get",
"invoice-storage.invoices.item.post",
"invoice-storage.invoices.item.put",

VRohach: @RuslanLavrov  Done! can you double-check please?

RuslanLavrov (approved):

src/main/java/org/folio/service/dataimport/utils/DataImportUtils.java
RuslanLavrov: thanks!

PR #1061, opened 2024-12-06
[EUREKA-561-4]. Add validateUserUnaffiliatedLocationUpdates method to Order Open/unopen

BKadirkhodjaev:
## Purpose

## Approach

- Add `validateUserUnaffiliatedLocationUpdates` to `updateOrder` method to guard against unaffiliated order opening/unopening
- Tested with Karate tests against a mod-orders modules: 

image

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderHelper.java
@@ -238,8 +238,8 @@ public class PurchaseOrderHelper {
             if (!compPO.getCompositePoLines().isEmpty()) {
               compPO.getCompositePoLines().forEach(poLine -> {
                 var compPoLineFromStorage = clonedPoFromStorage.getCompositePoLines().stream()
-                  .filter(Objects::nonNull).filter(entry-> Objects.nonNull(entry.getId()))
-                  .filter(entry -> entry.getId().equals(poLine.getId())).findFirst().orElse(null);
+                  .filter(entry -> StringUtils.equals(entry.getId(), poLine.getId()))
+                  .findFirst().orElse(null);
                 if (Objects.nonNull(compPoLineFromStorage)) {
                   var updatedLocations = poLine.getLocations();
                   var storedLocations = compPoLineFromStorage.getLocations();
--- a/src/main/java/org/folio/helper/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderHelper.java
@@ -239,11 +239,11 @@ public class PurchaseOrderHelper {
               compPO.getCompositePoLines().forEach(poLine -> {
                 var compPoLineFromStorage = clonedPoFromStorage.getCompositePoLines().stream()
                   .filter(Objects::nonNull).filter(entry-> Objects.nonNull(entry.getId()))
-                  .filter(entry -> entry.getId().equals(poLine.getId()))
-                  .findFirst().orElse(null);
+                  .filter(entry -> entry.getId().equals(poLine.getId())).findFirst().orElse(null);
                 if (Objects.nonNull(compPoLineFromStorage)) {
-                  var poLineFromStorage = PoLineCommonUtil.convertToPoLine(compPoLineFromStorage);
-                  poLineFutures.add(compositePoLineValidationService.validateUserUnaffiliatedLocationUpdates(poLine, poLineFromStorage, requestContext));
+                  var updatedLocations = poLine.getLocations();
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -224,9 +230,26 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
       .map(HelperUtils::convertToCompositePurchaseOrder)
       .compose(lines -> purchaseOrderLineService.populateOrderLines(lines, requestContext))
       .compose(poFromStorage -> {
+        CompositePurchaseOrder clonedPoFromStorage = JsonObject.mapFrom(poFromStorage).mapTo(CompositePurchaseOrder.class);
         boolean isTransitionToOpen = isTransitionToOpen(poFromStorage, compPO);
         return orderValidationService.validateOrderForUpdate(compPO, poFromStorage, deleteHoldings, requestContext)
-          .compose(ok -> {
+          .compose(v -> {
+            var poLineFutures = new ArrayList<Future<Void>>();
+            if (!compPO.getCompositePoLines().isEmpty()) {
```
SerhiiNosko, line 238: CollectionUtils.isNotEmpty is prefferable and easy to read as for me

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -224,9 +230,26 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
...
+              compPO.getCompositePoLines().forEach(poLine -> {
+                var compPoLineFromStorage = clonedPoFromStorage.getCompositePoLines().stream()
+                  .filter(Objects::nonNull).filter(entry-> Objects.nonNull(entry.getId()))
+                  .filter(entry -> entry.getId().equals(poLine.getId()))
+                  .findFirst().orElse(null);
+                if (Objects.nonNull(compPoLineFromStorage)) {
+                  var poLineFromStorage = PoLineCommonUtil.convertToPoLine(compPoLineFromStorage);
+                  poLineFutures.add(compositePoLineValidationService.validateUserUnaffiliatedLocationUpdates(poLine, poLineFromStorage, requestContext));
```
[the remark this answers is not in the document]
SerhiiNosko: also it is for bug fix, we need to avoid risky conversion operations, if needed refactoring we can do it for Sunflower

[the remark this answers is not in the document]
BKadirkhodjaev: I will try changing the method

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -224,9 +230,26 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
...
         boolean isTransitionToOpen = isTransitionToOpen(poFromStorage, compPO);
         return orderValidationService.validateOrderForUpdate(compPO, poFromStorage, deleteHoldings, requestContext)
-          .compose(ok -> {
+          .compose(v -> {
+            var poLineFutures = new ArrayList<Future<Void>>();
+            if (!compPO.getCompositePoLines().isEmpty()) {
+              compPO.getCompositePoLines().forEach(poLine -> {
+                var compPoLineFromStorage = clonedPoFromStorage.getCompositePoLines().stream()
+                  .filter(Objects::nonNull).filter(entry-> Objects.nonNull(entry.getId()))
+                  .filter(entry -> entry.getId().equals(poLine.getId())).findFirst().orElse(null);
```
SerhiiNosko, line 242: StringUtils.equals can be used to find correct po lnie from storage, so in this case 1 filter will be used and this method also avoids NPE

src/main/java/org/folio/service/orders/CompositePoLineValidationService.java
```diff
@@ -329,4 +344,38 @@ private void validateReceivingWorkflowForBindary(CompositePoLine poLine, List<Er
       errors.add(RECEIVING_WORKFLOW_INCORRECT_FOR_BINDARY_ACTIVE.toError());
     }
   }
+
+  public Future<Void> validateUserUnaffiliatedLocationUpdates(String poLineId, List<Location> updatedPoLineLoc,
```
SerhiiNosko, line 348: names should be in plural

SerhiiNosko (approved):

PR #1062, opened 2024-12-10
[EUREKA-561-5]. Simplify location validation, update tests

BKadirkhodjaev:
## Purpose

## Approach

- Add validation to Update Order action
- Update method arguments for Update Order Line action
- Simplify location validation
- Update unit tests

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderHelper.java
@@ -33,6 +33,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import io.vertx.core.json.JsonArray;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -292,6 +279,16 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
       });
   }
 
+  private Future<List<Void>> validateUserUnaffiliatedPoLineLocations(CompositePurchaseOrder purchaseOrder, RequestContext requestContext) {
+    var poLineFutures = new ArrayList<Future<Void>>();
+    if (!purchaseOrder.getCompositePoLines().isEmpty()) {
```
azizbekxm, line 284: I think check for empty is not need 
1) we are invoking after empty check line if (isTransitionToOpen && CollectionUtils.isEmpty(compPO.getCompositePoLines())) { and 
2) we are using filter

However, if it is better to use incase of invoking in other places in future lets use CollectionUtils.isNotEmpty(compPO.getCompositePoLines()  to be easier to read

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -292,6 +279,16 @@ public Future<Void> updateOrder(CompositePurchaseOrder compPO, boolean deleteHol
...
+      purchaseOrder.getCompositePoLines().stream()
+        .filter(Objects::nonNull)
+        .forEach(poLine -> poLineFutures.add(compositePoLineValidationService.validateUserUnaffiliatedLocationUpdates(poLine.getId(), poLine.getLocations(), requestContext)));
```
azizbekxm, line 287: also seems this line is long

src/main/java/org/folio/service/orders/CompositePoLineValidationService.java
```diff
@@ -345,27 +345,20 @@ private void validateReceivingWorkflowForBindary(CompositePoLine poLine, List<Er
...
-          updatedUnaffiliatedLocations.stream().map(Location::getTenantId).distinct().toList());
-        if (!SetUtils.isEqualSet(storageUnaffiliatedLocations, updatedUnaffiliatedLocations)) {
-          logger.info("validateUserUnaffiliatedLocationUpdates:: User is not affiliated with all locations on the POL, poLineId: '{}'",
-            poLineId);
-          return Future.failedFuture(new HttpException(422, ErrorCodes.LOCATION_UPDATE_WITHOUT_AFFILIATION));
+        var uniqueUnAffiliatedLocations = extractUnaffiliatedLocations(locations, userTenants).stream()
+          .map(Location::getTenantId).distinct().toList();
+        if (CollectionUtils.isNotEmpty(uniqueUnAffiliatedLocations)) {
+          logger.info("validateUserUnaffiliatedLocationUpdates:: User has unaffiliated locations on the POL, poLineId: {}, unique locations: {}", poLineId, uniqueUnAffiliatedLocations);
+          return Future.failedFuture(new HttpException(HttpStatus.SC_UNPROCESSABLE_ENTITY, ErrorCodes.LOCATION_UPDATE_WITHOUT_AFFILIATION));
```
azizbekxm, line 359: also lets add as a param of error poLine Id and uniqueAfflicatedLocations 
var params = List.of(new Parameter().withKey("poLineId......
errors.withParams..

src/main/java/org/folio/helper/PurchaseOrderHelper.java
BKadirkhodjaev: Better to leave this slightly excessive null checks because the method is also used for Unopen and also can be called from the Postman where there may not be any PoLines in the request to validate

src/main/java/org/folio/service/orders/CompositePoLineValidationService.java
BKadirkhodjaev: This PR is a bugfix, probably it's not safe to change the message on the fly

azizbekxm (approved):

PR #1063, opened 2024-12-10
[MODORDERS-1222] - Error when trying to Unopen an ongoing order to add a purchase order line with multiple funds

azizbekxm:
## Purpose

https://folio-org.atlassian.net/browse/MODORDERS-1222
## Approach

- Filtered duplicate poLines in side of retrieveMapFiscalYearsWithCompPOLines method

### DIFF
--- a/src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
+++ b/src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
@@ -1,15 +1,12 @@
 package org.folio.service.finance.transaction;
 
-import static java.util.stream.Collectors.groupingBy;
-import static java.util.stream.Collectors.mapping;
-import static java.util.stream.Collectors.toList;
-
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
```diff
@@ -167,23 +162,28 @@ private void mapHoldersToTransactions(List<EncumbranceRelationsHolder> encumbran
...
-          ))
+      .map(erhList -> erhList.stream()
+        .filter(erh -> erh.getCurrentFiscalYearId() != null)
+        .collect(Collectors.groupingBy(
+          EncumbranceRelationsHolder::getCurrentFiscalYearId,
+          Collectors.mapping(
+            EncumbranceRelationsHolder::getPoLine,
+            Collectors.collectingAndThen(
+              Collectors.toList(),
+              list -> list.stream().distinct().toList()
```
SerhiiNosko, line 183: so in this stream we jsut filter and prevent having duplicates in list by using distinct.
Probably without steams here would be easier to read but as you prefer.
Approving and also please add karate test

SerhiiNosko (approved):

BKadirkhodjaev (approved):

PR #1070, opened 2024-12-18
[MODORDERS-1209-2]. Update Send Claims and Piece Batch Update models with new properties

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1209>

## Approach

- Add extra properties to Send Claims (claiming interval, internalNote & externalNote) for saving
- Add extra property (claiming interval) to Piece Batch Update for saving
- Additional changes:
  - Change validation responses to throw an HTTP exception with code, message and parameters populated
  - Move all claiming-related error messages into ErrorCodes enum 
  - Switch places between calls, making jobs creation happen before pieces update
  - Add tenant validation on Integration Details retrieval


azizbekxm (approved):

src/main/java/org/folio/models/claiming/IntegrationDetailField.java
```diff
@@ -6,6 +6,7 @@
 @Getter
 @AllArgsConstructor
 public enum IntegrationDetailField {
+  TENANT("tenant"),
```
SerhiiNosko, line 9: Do you think this field is needed?

SerhiiNosko (approved):

src/main/java/org/folio/models/claiming/IntegrationDetailField.java
BKadirkhodjaev: > Do you think this field is needed?

Yeah I think so, just in case there may be some other integrations details with different tenants getting used to create jobs: 

image

Need some end-to-end testing to verify this behaviour though

PR #1072, opened 2024-12-19
[MODORDSTOR-435] Use batch piece update endpoint for batch status update

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDSTOR-435] Introduce new batch update pieces endpoint](https://folio-org.atlassian.net/browse/MODORDSTOR-435)

## Approach
- Use new batch update endpoint when updating piece statuses


src/main/java/org/folio/orders/utils/ResourcePathResolver.java
```diff
@@ -79,6 +80,7 @@ private ResourcePathResolver() {
     apis.put(PO_NUMBER, "/orders-storage/po-number");
     apis.put(PURCHASE_ORDER_STORAGE, "/orders-storage/purchase-orders");
     apis.put(PIECES_STORAGE, "/orders-storage/pieces");
+    apis.put(PIECES_STORAGE_BATCH, "/orders-storage/pieces-batch");
```
SerhiiNosko, line 83: need to add module permission to call this new endpoint

azizbekxm: Added permission

azizbekxm:
image

BKadirkhodjaev (approved):

PR #1074, opened 2024-12-25
[MODORDERS-1209-3]. Detach update pieces from job creations

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1209>
 
## Approach

- TODO

### DIFF
--- a/src/main/java/org/folio/service/ProtectionService.java
+++ b/src/main/java/org/folio/service/ProtectionService.java
@@ -69,7 +69,7 @@ public class ProtectionService {
    */
   public Future<Void> isOperationRestricted(List<String> unitIds, Set<ProtectedOperationType> operations, RequestContext requestContext) {
     if (userHasDesiredPermission(BYPASS_ACQ_UNITS, requestContext)) {
-      log.info("isOperationRestricted:: Bypassing checking acq units");
+      log.debug("isOperationRestricted:: Bypassing checking acq units");
       return Future.succeededFuture();
     }
     if (CollectionUtils.isNotEmpty(unitIds)) {
@@ -82,7 +82,7 @@ public class ProtectionService {
               .collect(Collectors.toList());
 
             if (!activeUnits.isEmpty() && applyMergingStrategy(activeUnits, operations)) {
-              log.info("isOperationRestricted:: Acq unit ids are merged");
+              log.debug("isOperationRestricted:: Acq unit ids are merged");
               return verifyUserIsMemberOfOrdersUnits(extractUnitIds(activeUnits), requestContext);
             }
             return Future.succeededFuture();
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/service/ProtectionService.java
```diff
@@ -69,7 +69,7 @@ public Future<Void> isOperationRestricted(List<String> unitIds, ProtectedOperati
    */
   public Future<Void> isOperationRestricted(List<String> unitIds, Set<ProtectedOperationType> operations, RequestContext requestContext) {
     if (userHasDesiredPermission(BYPASS_ACQ_UNITS, requestContext)) {
-      log.info("isOperationRestricted:: Bypassing checking acq units");
+      log.debug("isOperationRestricted:: Bypassing checking acq units");
```
BKadirkhodjaev, line 72: Change log level to make this log entry less abundant

src/main/java/org/folio/service/pieces/PiecesClaimingService.java
```diff
@@ -196,39 +195,39 @@ private Future<ClaimingResults> createJobsByVendor(ClaimingCollection claimingCo
...
-                                                Map.Entry<String, Object> configEntry, RequestContext requestContext) {
-    log.info("updatePiecesAndCreateJob:: Updating pieces and creating a job, job key: {}, count: {}", configEntry.getKey(), pieceIds.size());
-    return createJob(configEntry.getKey(), configEntry.getValue(), pieceIds, requestContext).map(pieceIds)
-      .compose(v -> pieceUpdateFlowManager.updatePiecesStatuses(pieceIds, PieceBatchStatusCollection.ReceivingStatus.CLAIM_SENT,
-        claimingCollection.getClaimingInterval(), claimingCollection.getInternalNote(), claimingCollection.getExternalNote(), requestContext));
+  private Future<Void> updatePieces(ClaimingCollection claimingCollection, String vendorId, List<String> pieceIds, RequestContext requestContext) {
+    log.info("updatePieces:: Updating pieces for vendor, vendorId: {}, pieces: {}", vendorId, pieceIds.size());
+    return pieceUpdateFlowManager.updatePiecesStatuses(pieceIds, PieceBatchStatusCollection.ReceivingStatus.CLAIM_SENT,
+      claimingCollection.getClaimingInterval(), claimingCollection.getInternalNote(), claimingCollection.getExternalNote(), requestContext);
```
azizbekxm, line 219: if we can simplify this in scope of this pr, it would be good. we can create another using overload and method will take claimingCollection and extract to this method

BKadirkhodjaev: Maybe better to just extract 3 variables for visibility in this current `updatePiecesStatuses()` method, because `PieceUpdateFlowManager.updatePiecesStatuses()` doesn't really need to take ClaimingCollection as a parameter

SerhiiNosko (approved):

azizbekxm (approved):

PR #1088, opened 2025-01-23
MODORDERS-1207 The order with 'Pending' status is created after importing file with mapping profile what contains 'Open' purchase order status

JavokhirAbdullayev:
## Purpose
[MODORDERS-1207](https://folio-org.atlassian.net/browse/MODORDERS-1207) The order with 'Pending' status is created after importing file with mapping profile what contains 'Open' purchase order status
an observed race condition is happening during update item during the creation of the piece 
put request to mod-inventory is not strictly happening after getting the request, as a result, subsequent put requests to mod-inv return us 409 conflict error.

## Approach
* sequentially update the item using concurrent map only if the get endpoint is returning the item

image
image

### DIFF
--- a/src/main/java/org/folio/service/inventory/InventoryItemManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryItemManager.java
@@ -165,10 +165,9 @@ public class InventoryItemManager {
 
   private void fetchAndUpdate(Piece piece, RequestContext requestContext, Promise<Void> promise) {
     String itemId = piece.getItemId();
-    String poLineId = piece.getPoLineId();
     getItemRecordById(itemId, true, requestContext)
       .compose(item -> {
-        if (poLineId == null || item == null || item.isEmpty()) {
+        if (item == null || item.isEmpty()) {
           return Future.succeededFuture();
         }
         InventoryUtils.updateItemWithPieceFields(item, piece);
--- a/src/main/java/org/folio/service/inventory/InventoryItemManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryItemManager.java
@@ -35,7 +35,6 @@ import java.util.Map;
 import java.util.Optional;
 
 import static java.util.stream.Collectors.toList;
-import static org.folio.orders.utils.HelperUtils.chainCall;
 import static org.folio.orders.utils.HelperUtils.collectResultsOnSuccess;
 import static org.folio.orders.utils.QueryUtils.convertIdsToCqlQuery;
 import static org.folio.rest.RestConstants.MAX_IDS_FOR_GET_RQ_15;
@@ -147,14 +146,15 @@ public class InventoryItemManager {
     String itemId = piece.getItemId();
 
     return getItemRecordById(itemId, true, requestContext)
-      .compose(item -> {
-        if (item == null || item.isEmpty()) {
-          return Future.succeededFuture(); // Nothing to update
-        }
-        InventoryUtils.updateItemWithPieceFields(item, piece);
-        return chainCall(Collections.singletonList(item), i -> updateItem(i, requestContext));
-      });
+        .compose(item -> {
+          if (item == null || item.isEmpty()) {
+            return Future.succeededFuture();
+          }
+          InventoryUtils.updateItemWithPieceFields(item, piece);
+          return updateItem(item, requestContext);
+        });
   }
+
 
   public Future<Void> deleteItem(String id, boolean skipNotFoundException, RequestContext requestContext) {
     RequestEntry requestEntry = new RequestEntry(INVENTORY_LOOKUP_ENDPOINTS.get(ITEM_BY_ID_ENDPOINT)).withId(id);
--- a/src/main/java/org/folio/service/inventory/InventoryItemManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryItemManager.java
@@ -98,6 +98,7 @@ public class InventoryItemManager {
 
   public Future<JsonObject> getItemRecordById(String itemId, boolean skipThrowNorFoundException, RequestContext requestContext) {
     RequestEntry requestEntry = new RequestEntry(INVENTORY_LOOKUP_ENDPOINTS.get(ITEM_BY_ID_ENDPOINT)).withId(itemId);
+    logger.info("getItemRecordById:: {}", requestContext.toString());
     return restClient.getAsJsonObject(requestEntry, skipThrowNorFoundException, requestContext);
   }
 
--- a/src/main/java/org/folio/service/inventory/InventoryItemManager.java
+++ b/src/main/java/org/folio/service/inventory/InventoryItemManager.java
@@ -143,18 +143,32 @@ public class InventoryItemManager {
     if (piece.getItemId() == null || piece.getPoLineId() == null || piece.getIsBound()) {
       return Future.succeededFuture();
     }
-    logger.info("updateItemWithPieceFields:: piece {} item {}", piece.getId(), piece.getItemId());
     String itemId = piece.getItemId();
     String poLineId = piece.getPoLineId();
+    return fetchAndUpdateItem(itemId, poLineId, piece, requestContext, 2);
+  }
+
+  private Future<Void> fetchAndUpdateItem(String itemId, String poLineId, Piece piece, RequestContext requestContext, int retries) {
     return getItemRecordById(itemId, true, requestContext)
       .compose(item -> {
         if (poLineId == null || item == null || item.isEmpty()) {
           return Future.succeededFuture();
         }
-        logger.info("updateItemWithPieceFields:: {}", item);
+        logger.info("fetchAndUpdateItem:: Current item: {}", item);
         InventoryUtils.updateItemWithPieceFields(item, piece);
-        return updateItem(item, requestContext);
+        return updateItem(item, requestContext)
+          .recover(error -> {
+            if (isConflictError(error) && retries > 0) {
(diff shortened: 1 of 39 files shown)


Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -142,15 +144,46 @@ public Future<Void> updateItemWithPieceFields(Piece piece, RequestContext reques
     if (piece.getItemId() == null || piece.getPoLineId() == null || piece.getIsBound()) {
       return Future.succeededFuture();
     }
+
+    // Create a promise for this operation
+    Promise<Void> promise = Promise.promise();
```
SerhiiNosko, line 149: can we change logic in pieces іукмшсу maybe to do sequentual processing(we have utility for this) or some other technic instead having this workaround?

JavokhirAbdullayev: Hi @SerhiiNosko you have typo іукмшсу  what does it mean?

SerhiiNosko: I mean, is it possible to fix the cause of the race condition instead of addressing the consequences? I don't know where this race condition is occurring, but one potential solution could be to use this chainCall utility method, which will execute features sequentially.
https://github.com/folio-org/mod-orders/blob/master/src/main/java/org/folio/orders/utils/HelperUtils.java#L292

JavokhirAbdullayev: thank you Serhii!
 will update the logic, this utility comes handy
the race condition is occurring because of non-blocking nature of vertx 
we should make a blocking operation on get request

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -142,18 +143,22 @@ public Future<Void> updateItemWithPieceFields(Piece piece, RequestContext reques
...
-        if (poLineId == null || item == null || item.isEmpty()) {
-          return Future.succeededFuture();
-        }
-        InventoryUtils.updateItemWithPieceFields(item, piece);
-        return updateItem(item, requestContext);
-      });
+
+    return chainCall(Collections.singletonList(itemId), id ->
+      getItemRecordById(id, true, requestContext)
+        .compose(item -> {
```
SerhiiNosko, line 151: Can you please explain difference, in previous implementation we used compose and in function chainCall() we also use compose between 2 futures.

src/main/java/org/folio/service/inventory/InventoryItemManager.java
```diff
@@ -142,15 +143,16 @@ public Future<Void> updateItemWithPieceFields(Piece piece, RequestContext reques
...
     return getItemRecordById(itemId, true, requestContext)
       .compose(item -> {
-        if (poLineId == null || item == null || item.isEmpty()) {
-          return Future.succeededFuture();
+        if (item == null || item.isEmpty()) {
+          return Future.succeededFuture(); // Nothing to update
         }
         InventoryUtils.updateItemWithPieceFields(item, piece);
-        return updateItem(item, requestContext);
+        return chainCall(Collections.singletonList(item), i -> updateItem(i, requestContext));
```
SerhiiNosko, line 155: If it is only single call it is not make sense to use chainCall(). This function is needed in case we have at least 2 calls and need to make them one after another using compose() method

JavokhirAbdullayev: in fact this fucntion its not guarantee ordering of the requests 
```
invoking getItemStorageItemsByItemId
[diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  LogUtil              10.0.127.188:57534 GET /item-storage/items/abc5e279-a55f-4ff8-9584-4d03da0712cb null HTTP_1_1 200 1030 3 tid=diku OK 
 [diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  RestRouting          invoking getItemStorageItemsByItemId
[diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  LogUtil              10.0.127.188:57534 GET /item-storage/items/abc5e279-a55f-4ff8-9584-4d03da0712cb null HTTP_1_1 200 1030 3 tid=diku OK 
 [diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  RestRouting          invoking putItemStorageItemsByItemId
[diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  LogUtil              10.0.127.188:57534 PUT /item-storage/items/abc5e279-a55f-4ff8-9584-4d03da0712cb null HTTP_1_1 204 0 19 tid=diku No Content 
[diku] [56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] INFO  RestRouting          invoking putItemStorageItemsByItemId
[56786e52-5c62-4210-8afd-db1fdc0b7a69] [mod_inventory_storage] WARN  dpointFailureHandler An error occurred
2025-01-31T10:45:02.670311602Z org.folio.rest.tools.client.exceptions.ResponseException: 409 Conflict
```
still makes 2 consequent get request and after 2 put requests

JavokhirAbdullayev: we need to execute this within single one future, to ensure atomicity and correct ordering

SerhiiNosko: In this case you need to use this function on more higher level. For example we have 2 items, for each item we are getting it and updating after this, so for each item we should have own chainCall()

JavokhirAbdullayev: i see, let me try

JavokhirAbdullayev: it didnt work

JavokhirAbdullayev: @SerhiiNosko a bunch of refactoring is needed in piece creation logic 
if we want to use this utility method. 
I see that the batch endpoint was added to mod-orders-storage recently

suggestion:
we can close this issue with sequential processing by using one future as I previously suggested and
I will create a separate ticket for addressing piece batch creation logic in mod-orders

JavokhirAbdullayev: i see even the ticket exist https://folio-org.atlassian.net/browse/MODORDERS-1109
if you don't mind we will merge this pr as it is

src/main/resources/log4j2-json.properties
```diff
@@ -33,6 +33,6 @@ appender.console.layout.moduleId.type = KeyValuePair
 appender.console.layout.moduleId.key = moduleId
 appender.console.layout.moduleId.value = $${FolioLoggingContext:moduleid}
 
-rootLogger.level = info
+rootLogger.level = ${sys:LOG_LEVEL:-INFO}
```
BKadirkhodjaev, line 36: I don't think we need this changed at the moment

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -95,31 +96,37 @@ private Future<List<Piece>> updatePieces(OpenOrderPieceHolder holder, RequestCon
...
 
-    logger.info("createPieces:: Passed acq unit validation and open order '{}' inventory update", order.getId());
-    return collectResultsOnSuccess(validationFutures)
-      .compose(validatedPieces ->
-        pieceStorageService.insertPiecesBatch(validatedPieces, requestContext)
-          .map(PieceCollection::getPieces)
+    // Use chainCallCollect to sequentially process each piece with openOrderUpdateInventory.
+    return chainCall(preparedPieces, piece ->
+      openOrderUpdateInventory(piece, order, isInstanceMatchingDisabled, requestContext)
+        .onSuccess(validatedPiece -> validationFutures.add(Future.succeededFuture(validatedPiece)))
```
SerhiiNosko, line 114: better to use map here instead of onSuccess. Using map() instead of onSuccess() would ensure that the execution of the previous step is completed before proceeding

SerhiiNosko (approved):

azizbekxm (approved):

src/main/java/org/folio/service/orders/flows/update/open/OpenCompositeOrderPieceService.java
```diff
@@ -95,31 +96,37 @@ private Future<List<Piece>> updatePieces(OpenOrderPieceHolder holder, RequestCon
...
+      openOrderUpdateInventory(piece, order, isInstanceMatchingDisabled, requestContext)
+        .map(validatedPiece -> validationFutures.add(Future.succeededFuture(validatedPiece)))
       )
+      .compose(ignored -> collectResultsOnSuccess(validationFutures))
+      .compose(validatedPieces -> {
+        logger.info("createPieces:: Passed acq unit validation and open order '{}' inventory update", order.getId());
+        // Once all pieces have been updated, insert them in batch.
+        return pieceStorageService.insertPiecesBatch(validatedPieces, requestContext)
+          .map(PieceCollection::getPieces);
+      })
```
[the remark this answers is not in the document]
JavokhirAbdullayev: sorry i didn't get your point, we need to sequentially update pieces collect them
then batch insert

PR #1099, opened 2025-03-18
[MODORDERS - 1260] - Implement endpoint to unlink Title from Package

azizbekxm:
## Purpose

https://folio-org.atlassian.net/browse/MODORDERS-1260
## Pre-Merge Checklist:

### DIFF
--- a/ramls/titles.raml
+++ b/ramls/titles.raml
@@ -53,10 +53,10 @@ resourceTypes:
     delete:
       description: Unlink a title from an instance
       queryParameters:
-        deleteHolding:
-          displayName: The associated holding should be removed
+        deleteHoldings:
+          displayName: The associated holdings should be removed
           type: string
-          description: The associated holding should be removed
+          description: The associated holdings should be removed
           example: "true"
           required: false
       responses:
--- a/ramls/titles.raml
+++ b/ramls/titles.raml
@@ -44,14 +44,8 @@ resourceTypes:
       collection-item:
         exampleItem: !include acq-models/mod-orders-storage/examples/title_get.sample
         schema: title
-
-  /unlink/{id}:
-    uriParameters:
-      id:
-        description: The UUID of a Title
-        type: UUID
     delete:
-      description: Unlink a title from an instance
+      description: Delete a title record
       queryParameters:
         deleteHoldings:
           displayName: The associated holdings should be removed
@@ -60,10 +54,8 @@ resourceTypes:
           example: "true"
           required: false
       responses:
-        201:
-          description: holdings are existing
         204:
-          description: Title unlinked
+          description: Title deleted
         404:
           description: Title not found
         500:
--- a/ramls/titles.raml
+++ b/ramls/titles.raml
@@ -50,8 +50,16 @@ resourceTypes:
       id:
         description: The UUID of a Title
         type: UUID
-    post:
+    put:
       description: Unlink a title from an instance
+      queryParameters:
+        deleteHolding:
+          displayName: The associated holding should be removed
+          type: boolean
+          description: The associated holding should be removed
+          example: true
+          required: false
+          default: false
       responses:
         204:
           description: Title unlinked
--- a/ramls/titles.raml
+++ b/ramls/titles.raml
@@ -44,3 +44,18 @@ resourceTypes:
       collection-item:
         exampleItem: !include acq-models/mod-orders-storage/examples/title_get.sample
         schema: title
+
+  /unlink/{id}:
+    uriParameters:
+      id:
+        description: The UUID of a Title
+        type: UUID
+    post:
+      description: Unlink a title from an instance
+      responses:
+        204:
+          description: Title unlinked
+        404:
+          description: Title not found
+        500:
+          description: Internal server error
--- a/ramls/titles.raml
+++ b/ramls/titles.raml
@@ -50,17 +50,18 @@ resourceTypes:
       id:
         description: The UUID of a Title
         type: UUID
-    put:
+    delete:
       description: Unlink a title from an instance
       queryParameters:
         deleteHolding:
           displayName: The associated holding should be removed
-          type: boolean
+          type: string
           description: The associated holding should be removed
-          example: true
+          example: "true"
           required: false
-          default: false
       responses:
+        201:
+          description: holdings are existing
         204:
           description: Title unlinked
         404:
--- a/src/main/java/org/folio/rest/impl/TitlesApi.java
+++ b/src/main/java/org/folio/rest/impl/TitlesApi.java
@@ -110,7 +110,13 @@ public class TitlesApi extends BaseApi implements OrdersTitles {
   @Override
   public void deleteOrdersTitlesUnlinkById(String id, String deleteHolding, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     titlesService.unlinkTitleFromPackage(id, deleteHolding, new RequestContext(vertxContext, okapiHeaders))
-      .onSuccess(response -> asyncResultHandler.handle(succeededFuture(buildOkResponse(response))))
+      .onSuccess(response -> {
+        if (response == null || response.isEmpty()) {
+          asyncResultHandler.handle(succeededFuture(buildNoContentResponse()));
+        } else {
+          asyncResultHandler.handle(succeededFuture(buildOkResponse(response)));
+        }
+      })
       .onFailure(fail -> handleErrorResponse(asyncResultHandler, fail));
   }
 }
--- a/src/main/java/org/folio/rest/impl/TitlesApi.java
+++ b/src/main/java/org/folio/rest/impl/TitlesApi.java
@@ -108,11 +108,9 @@ public class TitlesApi extends BaseApi implements OrdersTitles {
   }
 
   @Override
-  public void putOrdersTitlesUnlinkById(String id, boolean deleteHolding, Map<String, String> okapiHeaders,
-                                        Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+  public void deleteOrdersTitlesUnlinkById(String id, String deleteHolding, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     titlesService.unlinkTitleFromPackage(id, deleteHolding, new RequestContext(vertxContext, okapiHeaders))
-      .onSuccess(v -> asyncResultHandler.handle(succeededFuture(buildNoContentResponse())))
+      .onSuccess(response -> asyncResultHandler.handle(succeededFuture(buildOkResponse(response))))
       .onFailure(fail -> handleErrorResponse(asyncResultHandler, fail));
   }
-
 }
--- a/src/main/java/org/folio/rest/impl/TitlesApi.java
+++ b/src/main/java/org/folio/rest/impl/TitlesApi.java
@@ -74,10 +74,10 @@ public class TitlesApi extends BaseApi implements OrdersTitles {
 
   @Override
   @Validate
-  public void deleteOrdersTitlesById(String id, Map<String, String> okapiHeaders,
+  public void deleteOrdersTitlesById(String id, String deleteHoldings, Map<String, String> okapiHeaders,
                                      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    titlesService.deleteTitle(id, new RequestContext(vertxContext, okapiHeaders))
-      .onSuccess(v -> asyncResultHandler.handle(succeededFuture(buildNoContentResponse())))
+    titlesService.deleteTitle(id, deleteHoldings, new RequestContext(vertxContext, okapiHeaders))
+      .onSuccess(response -> asyncResultHandler.handle(succeededFuture(buildNoContentResponse())))
       .onFailure(fail -> handleErrorResponse(asyncResultHandler, fail));
   }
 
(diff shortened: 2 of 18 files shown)


src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -142,4 +167,115 @@ public Future<String> updateTitleWithInstance(Title title, boolean isInstanceMat
...
+        return Future.succeededFuture(CollectionUtils.isNotEmpty(holdingIds) ? holdingIds : List.of());
+      });
+  }
+
+  private Future<Void> deleteHoldingItemPieces(PoLine poLine, RequestContext requestContext) {
+    // TODO: add correct context for deletion of holdings reference change instance connection
+    return getHoldings(poLine, requestContext)
+      .compose(holdingIds -> deleteItemsForHolding(holdingIds, requestContext)
+        .compose(v -> deletePiecesForPoLine(poLine.getId(), requestContext))
+        .compose(v -> deleteHoldings(holdingIds, requestContext)));
```
azizbekxm, line 230: need some review on putting context correctly

BKadirkhodjaev: Hmm there is some potential for this to fail with Central Ordering enabled, because some pieces can be in a different tenant, you need to switch from Central tenant to the member one to delete stuff correctly.

You can potentially group pieces by `receivingTenantId` on a Piece and run piece-item-holding deletion in a forEach collecting into a list of Futures

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -142,4 +170,162 @@ public Future<String> updateTitleWithInstance(Title title, boolean isInstanceMat
...
+        }
+
+        if (StringUtils.isEmpty(deleteHolding)) {
+          log.info("processHoldings:: Holdings in poLine '{}' will not be deleted", poLine.getId());
+          return getHoldingsWithTenants(poLine, requestContext)
+            .map(holdingsByTenantId ->
+              holdingsByTenantId.values().stream().flatMap(Collection::stream).distinct().toList());
+        }
+
+        if (!"true".equalsIgnoreCase(deleteHolding) && !"false".equalsIgnoreCase(deleteHolding)) {
```
BKadirkhodjaev, line 207: Can be simplified with something like: 

```java
if (!Set.of("true", "false").equals(value.toLowerCase())) {
...
}
```

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -142,4 +170,162 @@ public Future<String> updateTitleWithInstance(Title title, boolean isInstanceMat
...
+
+  private Future<Map<String, List<String>>> getHoldingsWithTenants(PoLine poLine, RequestContext requestContext) {
+    return pieceStorageService.getPiecesByLineId(poLine.getId(), requestContext)
+      .compose(pieces -> {
+
+        Map<String, List<String>> holdingIdsByTenant = pieces.stream()
+          .filter(piece -> StringUtils.isNotEmpty(piece.getHoldingId()))
+          .collect(groupingBy(Piece::getReceivingTenantId, mapping(Piece::getHoldingId, toList())));
+
+        poLine.getLocations().forEach(location -> {
```
BKadirkhodjaev, line 228: Is this really necessary or in some cases there is a mismatch between Piece holdings and POL location holdings?

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -142,4 +170,162 @@ public Future<String> updateTitleWithInstance(Title title, boolean isInstanceMat
...
+          if (StringUtils.isNotEmpty(location.getHoldingId())) {
+            holdingIdsByTenant.computeIfAbsent(location.getTenantId(),
+              k -> new ArrayList<>()).add(location.getHoldingId());
+          }
+        });
+
+        holdingIdsByTenant.forEach((tenantId, holdings) ->
+          holdingIdsByTenant.put(tenantId, holdings.stream().distinct().toList()));
```
BKadirkhodjaev, line 236: It seems like needs to be collected into a new map because it seems like it can cause Concurrent modification exceptions

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -142,4 +170,162 @@ public Future<String> updateTitleWithInstance(Title title, boolean isInstanceMat
...
+  }
+
+  private Future<Void> deleteHoldingItemPieces(PoLine poLine, RequestContext centralContext) {
+    // TODO: add correct context for deletion of holdings reference change instance connection
+    return getHoldingsWithTenants(poLine, centralContext)
+      .compose(holdingIdsByTenant -> {
+
+        var tenantOperationFutures = new ArrayList<Future<Void>>();
+        holdingIdsByTenant.forEach((tenantId, holdingIds) -> {
+          if (CollectionUtils.isEmpty(holdingIds)) {
```
BKadirkhodjaev, line 249: Maybe best filter these null holdings before had in a Stream

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +156,154 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
+        var holdingIdsByTenant = pieces.stream()
+          .filter(piece -> StringUtils.isNotEmpty(piece.getHoldingId()))
+          .collect(groupingBy(Piece::getReceivingTenantId, mapping(Piece::getHoldingId, toList())));
+
+        return Future.succeededFuture(holdingIdsByTenant);
+      });
+  }
+
+  private Future<Void> deleteHoldingItemPieces(PoLine poLine, Title title, RequestContext centralContext) {
+    return getHoldingsWithTenants(poLine, title, centralContext)
```
BKadirkhodjaev, line 228: We can call this only in ECS if Central Ordering is enabled, and if the env is single tenant we can bypass all tenant grouping and delete all with the current context

BKadirkhodjaev (approved):

ramls/titles.raml
```diff
@@ -44,3 +44,27 @@ resourceTypes:
       collection-item:
         exampleItem: !include acq-models/mod-orders-storage/examples/title_get.sample
         schema: title
+
+  /unlink/{id}:
+    uriParameters:
+      id:
+        description: The UUID of a Title
+        type: UUID
+    delete:
```
damien-git, line 53: The ticket description mentions a POST method. Description should be updated if DELETE is used instead. I feel like POST makes more sense than DELETE, as we are not deleting the title or an "unlink" resource. I think DELETE should only be used to delete a resource by id.

src/main/java/org/folio/rest/impl/TitlesApi.java
```diff
@@ -107,4 +107,16 @@ public void putOrdersTitlesById(String id, Title entity, Map<String, String> oka
       .onFailure(fail -> handleErrorResponse(asyncResultHandler, fail));
   }
 
+  @Override
+  public void deleteOrdersTitlesUnlinkById(String id, String deleteHolding, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
damien-git, line 111: This line is too long.

ramls/titles.raml
```diff
@@ -44,3 +44,27 @@ resourceTypes:
...
+      description: Unlink a title from an instance
+      queryParameters:
+        deleteHolding:
+          displayName: The associated holding should be removed
```
damien-git, line 57: I think this should be "`deleteHoldings`" and "holdings". Or to be clear, I think this parameter should be "`forceDeletePiecesHoldingsAndItems`" (defaulting to `false`), because they are deleted by default, unless connected to another title.

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +160,175 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
     return titleInstanceService.getOrCreateInstance(title, isInstanceMatchingDisabled, locationContext)
       .map(title::withInstanceId)
       .compose(entity -> saveTitle(entity, requestContext)
         .map(v -> entity.getInstanceId()));
   }
 
+  /**
+   * Unlinks a title from a package. This involves checking the existence of holdings and items for the title.
+   * If the holdings have references to other titles, the unlinking process skips them.
+   * If the holdings do not have references, it asks for confirmation to delete related pieces, items, and holdings.
```
damien-git, line 180: From what I read in [UIOR-968](https://folio-org.atlassian.net/browse/UIOR-968), if there are no references to other titles, pieces holdings and titles should automatically be deleted by backend (confirmation is done beforehand by UI).

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +160,175 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
+   * @return a Future representing the result of the unlinking operation
+   */
+  public Future<List<String>> unlinkTitleFromPackage(String titleId, String deleteHolding,
+                                                     RequestContext requestContext) {
+    log.debug("Trying to unlink title with id: {} and deleteHolding: {}", titleId, deleteHolding);
+
+    return getTitleById(titleId, requestContext)
+      .compose(title -> purchaseOrderLineService.getOrderLineById(title.getPoLineId(), requestContext)
+        .compose(poLine -> processHoldingItemPieces(poLine, title, deleteHolding, requestContext))
+        .compose(holdings -> deleteTitle(title, holdings, requestContext)));
```
damien-git, line 194: I think we don't want to delete the title here, just unlink it. There is a different endpoint (`DELETE /orders/titles/{id}`) to delete titles.

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +160,175 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
+  }
+
+  private Future<List<String>> processHoldingItemPieces(PoLine poLine, Title title, String deleteHolding,
+                                                        RequestContext requestContext) {
+    if (StringUtils.isEmpty(deleteHolding)) {
```
damien-git, line 199: I think by default (without the parameter), holdings should be deleted, unless connected to another title. If connected, an error should be returned with a specific code that UI can check to display a modal and try again with the parameter set to `true`.

damien-git (changes requested):

src/main/java/org/folio/rest/impl/TitlesApi.java
azizbekxm: fixed it

ramls/titles.raml
azizbekxm: Updated

src/main/java/org/folio/service/titles/TitlesService.java
azizbekxm: 5th AC is saying that if holdings don't have reference and then confirmation model should be displayed, so currently logic will throw error to use deleteHoldings flag before proceed
```
    When user clicks confirm and their are related Holdings that do not reference any other pieces or items
    Then Delete Holdings confirmation is shown
    AND user can choose to Keep Holdings OR Delete Holding(s)
```

damien-git: I did not realize there was going to be two confirmation modals. OK.

src/main/java/org/folio/service/titles/TitlesService.java
SerhiiNosko: If we want to unlink a title without deleting it, is to add a new boolean flag called ‘linked’ to the Title schema. This flag can be used to filter out titles from the Package PO line when linked=false. In this case, we would be able to re-link the title again when needed.
We asked about this in https://folio-org.atlassian.net/browse/MODORDERS-1260

ramls/titles.raml
SerhiiNosko: POST is more suitable for Unlinking operations for me as well, DELETE is for deleting title

azizbekxm: After discussion, we agreed to move DELETE orders/titles/{id} endpoint

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +158,212 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
 
+  /**
+   * Delete title. This involves checking the existence of holdings and items for the title.
+   * If the holdings have references to other titles, the unlinking process skips them.
+   * If the holdings do not have references, it asks for confirmation to delete related pieces, items, and holdings.
+   *
+   * @param titleId        the ID of the title to unlink
+   * @param deleteHoldings  flag indicating whether to delete the holding if it has no other references
+   * @param requestContext the request context
+   * @return a Future representing the result of the unlinking operation
```
SerhiiNosko, line 183: word 'unlinking' in this javadoc can be replaced

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +158,212 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
+    List<Future<AbstractMap.SimpleEntry<String, String>>> deletableHoldingsFuture = new ArrayList<>();
+    holdingIdsGroupByTenant.forEach((tenantId, holdingIds) -> {
+      var tenantContext = holder.isCentralEnabled()
+        ? createContextWithNewTenantId(requestContext, tenantId)
+        : requestContext;
+
+      holdingIds.forEach(holdingId -> deletableHoldingsFuture.add(
+        inventoryItemManager.getItemsByHoldingId(holdingId, tenantContext)
+          .map(items -> items.stream()
+            .anyMatch(item -> holder.getPoLineId().equals(item.getString("purchaseOrderLineIdentifier")))
```
SerhiiNosko, line 256: There is constant in InventoryItemManager ITEM_PURCHASE_ORDER_LINE_IDENTIFIER that we can use here

damien-git (approved):

src/main/java/org/folio/service/titles/TitlesService.java
```diff
@@ -130,16 +158,212 @@ private List<String> getNonPackageLineIds(List<CompositePoLine> compositePoLines
...
+  private Future<Map<String, List<String>>> getHoldingsToDeleteWithTenants(TitleHolder holder, RequestContext requestContext) {
+    return getHoldingIdsGroupedByTenantId(holder.getTitle(), holder.getPoLine(), requestContext)
+      .compose(holdingIdsGroupedByTenant -> extractDeletableHoldings(holder, holdingIdsGroupedByTenant, requestContext));
+  }
+
+  private Future<Map<String, List<String>>> getHoldingIdsGroupedByTenantId(Title title, PoLine poLine, RequestContext requestContext) {
+    return pieceStorageService.getPiecesByLineIdAndTitleId(poLine.getId(), title.getId(), requestContext)
+      .map(pieces -> pieces.stream()
+        .filter(piece -> StringUtils.isNotEmpty(piece.getHoldingId()))
+        .collect(groupingBy(Piece::getReceivingTenantId, mapping(Piece::getHoldingId, toList())))
```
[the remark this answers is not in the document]
azizbekxm: Fixed it, using tenantId from request in case of null

SerhiiNosko (approved):

PR #1111, opened 2025-04-16
[MODORDERS-1252]. Switch to using the Finance Exchange Rate Api

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1252>

## Approach

- Add `CacheableExchangeRateService` that will be called before `CentralExchangeRateProvider` is initialized
- `CentralExchangeRateProvider` will handle both dynamic API-based and custom exchange rates
- Fully disable Java Money Moneta's inbuilt external providers by setting `urls` to empty string and `type` set to **NEVER**
- Classes affected by new Exchange Rate logic: 
  - `OrderRolloverService`
  - `FinanceHoldersBuilder`
  - `ReEncumbranceHoldersBuilder`
  - `OrderLinesSummaryPopulateService`
- Remove the need for the blocking code (`Vertx.executeBlocking(() -> ..)` which will speed up the UI of the Orders App
- Update and clean test classes, remove some deprecated methods and add more unit tests


src/main/java/org/folio/service/orders/OrderLinesSummaryPopulateService.java
```diff
@@ -56,31 +61,48 @@ public Future<CompositeOrderRetrieveHolder> populate(CompositeOrderRetrieveHolde
...
       RequestContext requestContext) {
     return configurationEntriesCache.getSystemCurrency(requestContext)
-      .compose(toCurrency -> getCollect(compositePoLines, requestContext, toCurrency)
-        .map(amounts -> amounts.stream()
-        .reduce(Money.of(0, toCurrency), Money::add)
-        .getNumber()
-        .doubleValue()));
+      .compose(toCurrency -> getExchangeRatesPerPoLine(compositePoLines, toCurrency, requestContext)
+      .compose(poLineExchangeRate -> Future.succeededFuture(Pair.of(toCurrency, poLineExchangeRate)))
+      .compose(toCurrencyPolExcRates -> Future.succeededFuture(convertEstimatedPrice(compositePoLines, toCurrencyPolExcRates))));
```
SerhiiNosko, line 66: with .map would be simpler here
```suggestion
.map(poLineExchangeRate -> Pair.of(toCurrency, poLineExchangeRate))
.map(toCurrencyPolExcRates -> convertEstimatedPrice(compositePoLines, toCurrencyPolExcRates)));
```

src/main/java/org/folio/service/orders/OrderLinesSummaryPopulateService.java
```diff
@@ -56,31 +61,48 @@ public Future<CompositeOrderRetrieveHolder> populate(CompositeOrderRetrieveHolde
...
+  }
+
+  private Future<Map<String, ExchangeRate>> getExchangeRatesPerPoLine(List<CompositePoLine> poLines, String toCurrency, RequestContext requestContext) {
+    var poLineExchangeRateFutures = new ArrayList<Future<Map<String, ExchangeRate>>>();
+    poLines.forEach(poLine -> {
+      var cost = poLine.getCost();
+      poLineExchangeRateFutures.add(cacheableExchangeRateService.getExchangeRate(cost.getCurrency(), toCurrency, cost.getExchangeRate(), requestContext)
+        .compose(exchangeRate -> Future.succeededFuture(Map.of(poLine.getId(), exchangeRate))));
```
SerhiiNosko, line 74: the same, map can be used here

SerhiiNosko (approved):

BKadirkhodjaev:
The necessary permission `finance.exchange-rate.item.get` is already present for: 

- GET /orders/composite-orders/{id}
- POST /orders/composite-orders/{id}/re-encumber
- POST /orders/rollover

PR #1140, opened 2025-07-01
[MODORDERS-1321]. Add 'Suppress from discovery' toggle for instance interaction

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1321>

## Approach

- Add mapping of a new `suppressInstanceFromDiscovery` field during instance creation
- Bump `order-lines` interface minor version to **4.1** in the `ModuleDescriptor-template.json`
- Update deprecated methods, set @SuppressWarnings where necessary, and update tests

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderLineHelper.java
@@ -380,7 +380,7 @@ public class PurchaseOrderLineHelper {
       return;
     }
     compPO.getPoLines().forEach(poLine -> {
-      var fields = POLineProtectedFieldsUtil.getFieldNames(poLine.getOrderFormat().value());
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/service/inventory/InventoryInstanceManager.java
```diff
@@ -457,9 +458,8 @@ private Future<SharingInstance> createShadowInstanceIfNeeded(String instanceId,
       });
   }
 
-  private Future<JsonObject> getInstanceById(String instanceId, boolean skipNotFoundException, RequestContext requestContext) {
+  private Future<JsonObject> getInstanceById(String instanceId, RequestContext requestContext) {
```
BKadirkhodjaev, line 461: Sonar recommended to set `skipNotFoundException` to `true` because no other method calls it with `false`

PR #1151, opened 2025-07-17
[MODFIN-434]. Update Treasury exchange rate provider handler to allow using the same exchange rate in both directions

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODFIN-434>

## Approach

- Update CacheableExchangeRateService to support a new operationMode field
- Update Order Totals, Order Opening, Rollover and Re-encumber Encumbrances by setting the new operationMode field

### DIFF
--- a/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
+++ b/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
@@ -24,10 +24,7 @@ public class ManualCurrencyConversion extends AbstractCurrencyConversion {
 
   public ManualCurrencyConversion(ConversionQuery conversionQuery, ExchangeRateProvider rateProvider,
                                   ConversionContext conversionContext) {
-    super(conversionQuery.getCurrency(), conversionContext);
-    this.conversionQuery = conversionQuery;
-    this.rateProvider = rateProvider;
-    this.operationMode = OperationMode.MULTIPLY;
+    this(conversionQuery, rateProvider, conversionContext, OperationMode.MULTIPLY);
   }
 
   public ManualCurrencyConversion(ConversionQuery conversionQuery, ExchangeRateProvider rateProvider,
--- a/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
+++ b/src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
@@ -19,27 +19,22 @@ public class ManualCurrencyConversion extends AbstractCurrencyConversion {
 
   private final ExchangeRateProvider rateProvider;
   private final ConversionQuery conversionQuery;
-  private final OperationMode operationMode;
+  private final org.folio.rest.acq.model.finance.ExchangeRate.OperationMode operationMode;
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/service/exchange/ManualCurrencyConversion.java
```diff
@@ -19,29 +19,24 @@ public class ManualCurrencyConversion extends AbstractCurrencyConversion {
...
-  private final OperationMode operationMode;
+  private final org.folio.rest.acq.model.finance.ExchangeRate.OperationMode operationMode;
 
   public ManualCurrencyConversion(ConversionQuery conversionQuery, ExchangeRateProvider rateProvider,
                                   ConversionContext conversionContext) {
     super(conversionQuery.getCurrency(), conversionContext);
     this.conversionQuery = conversionQuery;
     this.rateProvider = rateProvider;
-    this.operationMode = OperationMode.MULTIPLY;
+    this.operationMode = org.folio.rest.acq.model.finance.ExchangeRate.OperationMode.MULTIPLY;
```
Saba-Zedginidze-EPAM, line 29: Maybe we can import ExchangeRate.OperationMode directly here instead using full package names

BKadirkhodjaev: > Maybe we can import ExchangeRate.OperationMode directly here instead using full package names

There is a problem with duplicates, it does not allow to shorten it up unfortunately

src/main/java/org/folio/service/orders/OrderRolloverService.java
```diff
@@ -373,9 +374,11 @@ protected BigDecimal calculateTotalInitialAmountEncumbered(PoLineEncumbrancesHol
     return BigDecimal.valueOf(totalAmountAfterConversion.doubleValue());
   }
 
-  protected CurrencyConversion retrieveCurrencyConversion(String fromCurrency, String toCurrency, Number exchangeRate, boolean isCustomExchangeRate) {
-    var query = buildConversionQuery(fromCurrency, toCurrency, exchangeRate);
-    var provider = new CustomExchangeRateProvider(Boolean.TRUE.equals(isCustomExchangeRate) ? DIVIDE : MULTIPLY);
+  protected CurrencyConversion retrieveCurrencyConversion(ExchangeRate exchangeRate, boolean isCustomExchangeRate) {
+    var query = buildConversionQuery(exchangeRate.getFrom(), exchangeRate.getTo(), exchangeRate.getExchangeRate());
+    var operationMode = isCustomExchangeRate || exchangeRate.getOperationMode() == DIVIDE ? DIVIDE : MULTIPLY;
```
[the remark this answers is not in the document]
BKadirkhodjaev: > I see the logic is different here and we have 2 conditions. We can either leave it, or have it simplified as: `isCustomExchangeRate ? DIVIDE : exchangeRate.getOperationMode();`.

Yeah, this looks good, will update now

SerhiiNosko (approved):

PR #1163, opened 2025-08-08
MODORDERS-1341 - Configure consumer groupInstanceId

RuslanLavrov:
## Purpose
to configure groupInstanceId parameter during consumer wrapper creation

## Learning
[MODORDERS-1341](https://issues.folio.org/browse/MODORDERS-1341)

### DIFF
--- a/src/main/java/org/folio/verticle/DataImportConsumerVerticle.java
+++ b/src/main/java/org/folio/verticle/DataImportConsumerVerticle.java
@@ -24,6 +24,7 @@ import org.springframework.stereotype.Component;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
 
 import static org.folio.DataImportEventTypes.DI_INCOMING_MARC_BIB_FOR_ORDER_PARSED;
 import static org.folio.DataImportEventTypes.DI_INVENTORY_INSTANCE_MATCHED;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/verticle/DataImportConsumerVerticle.java
```diff
@@ -83,6 +84,7 @@ public void start(Promise<Void> startPromise) {
         .globalLoadSensor(globalLoadSensor)
         .subscriptionDefinition(subscriptionDefinition)
         .processRecordErrorHandler(getErrorHandler())
+        .groupInstanceId(getClass().getSimpleName() + "-" + UUID.randomUUID())
```
SerhiiNosko, line 87: It can help to resolve consumer rebalancing issue that appears in logs?

SerhiiNosko (approved):

src/main/java/org/folio/verticle/DataImportConsumerVerticle.java
RuslanLavrov: this change sets "group.instance.id" parameter for underlying consumer that may prevent the occurence of MemberIdRequiredException. Also, it can reduce time of rebalancing process because Kafka will retain previous consumer assingment

PR #1173, opened 2025-09-04
MODORDERS-1040 - Subscribe to JobExecution cancellation and add cache

RuslanLavrov:
## Purpose
to subscribe the module to the event about import job cancellation (DI_JOB_CANCELLED) to skip event processing for cancelled import jobs in the future.

## Approach
* create cache to store cancelled jobs IDs
* subscribe module to DI_JOB_CANCELLED event about import job cancellation and save received cancelled jobs ids to the cache. Consumer is created within a group to pick up topics created after consumer creation. The module will consume all messages from the topic each time the module runs.
* add tests

## Learning
[MODORDERS-1040](https://issues.folio.org/browse/MODORDERS-1040)

## Misc
Fixed expiration time configuration for JobProfileSnapshotCache and MappingParametersCache caches

### DIFF
--- a/src/test/java/org/folio/verticle/CancelledJobExecutionConsumerVerticleTest.java
+++ b/src/test/java/org/folio/verticle/CancelledJobExecutionConsumerVerticleTest.java
@@ -10,7 +10,6 @@ import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
-import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.folio.TestConfig;
 import org.folio.kafka.KafkaConfig;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/service/caches/CancelledJobsIdsCache.java
```diff
@@ -0,0 +1,44 @@
...
+ * An in-memory cache that stores IDs of cancelled import jobs.
+ */
+@Component
+public class CancelledJobsIdsCache {
+
+  private final Cache<UUID, Boolean> cache;
+
+  public CancelledJobsIdsCache(
+    @Value("${orders.cache.cancelled.jobs.expiration.minutes:1440}") long cacheExpirationTimeMins) {
+    this.cache = Caffeine.newBuilder()
```
BKadirkhodjaev, line 21: I would suggest we refactor this to resemble my latest changes: <https://github.com/folio-org/mod-orders/pull/1110/files#diff-a71e96fd2eb9bd619c3790e739d01d1f8aaff6887c440c9b2f38c7d18e5581d5R47>

Also note that some classes have an issue with using cache, here is a related open ticket that has other closed linked ones attached: <https://folio-org.atlassian.net/browse/MODORDERS-1274>

RuslanLavrov: could you say regarding refactoring do you mean to move caffeine cache creation to a method annotated with @PostConstruct, right?

regarding the issue with caches, it looks like this issue is not relevant for this cache because here we are injecting the expirationTime into the constructor parameter, so at the moment of Caffeine cache creation, the initialized/injected expirationTime is used for Caffeine cache creation

BKadirkhodjaev: The PR cache implementation is alright, it was just my misinterpretation of an unrelated issue with the caffeine cache. Approving.

PR #1285, opened 2026-04-15
[MODORDERS-1415] Do not process inventory when processing a piece and encumbrance change is not possible

damien-git:
## Purpose
- [MODORDERS-1415](https://folio-org.atlassian.net/browse/MODORDERS-1415) - Fix holdings and items creation/deletion logic when budget or ledger restrictions prevent operation
- [MODORDERS-1434](https://folio-org.atlassian.net/browse/MODORDERS-1434) - Batch create pieces: encumbrance is not updated
- [MODORDERS-1435](https://folio-org.atlassian.net/browse/MODORDERS-1435) - Changing a piece format: encumbrance not updated

## Approach
The main difficulty here was that the cost needs to be updated before processing encumbrances, encumbrances need to be processed before inventory so that an error stops the processing before inventory records are created, but updating the po line locations can only be done after updating the inventory.
The solution was to split updating po line quantities into updating the cost and updating the locations. The new general workflow is:
- Update the po line cost
- Process encumbrances
- Process inventory
- Update locations
- Save the po line

To update encumbrances when adding pieces in batch, encumbrances are processed sequentially, modifying the po line record progressively. The po line is now saved once (previously it was saved multiple times in parallel).

Updating a piece now triggers processing encumbrances if the piece format was changed.

## Related PR
- [integration test](https://github.com/folio-org/folio-integration-tests/pull/2279)

---

## Pre-Review Checklist

  - [ ] API/schema changes


SerhiiNosko:
# Code Review: MODORDERS-1415

**Branch:** `MODORDERS-1415`
**Commit:** `d7868e66` — Do not process inventory when processing a piece and encumbrance change is not possible
**Reviewer:** Claude Code
**Date:** 2026-04-16

---

## Summary

This change splits the `poLineUpdateQuantity` method into two separate operations — `poLineUpdateCost` (encumbrance-related) and `poLineUpdateLocations` (inventory-related) — and reorders the piece processing flow so that:

1. **Cost/encumbrance update** happens **before** inventory processing
2. **Location update + PO line save** happens **after** inventory processing

The goal is to avoid inventory processing when an encumbrance change is not possible (package lines, checkin-items lines). A `skipUpdatingPoLine` guard was moved from the flow managers into the base service, and a `lineUpdated` flag was added to `BasePieceFlowHolder` to carry state between the two split phases.

---

## Findings

### 🔥 Critical

#### C1. `poLineUpdateCost` in `PieceCreateFlowPoLineService` always returns `true` — even when nothing meaningful changed

**File:** `src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowPoLineService.java:31-43`

`poLineUpdateCost` unconditionally increments the cost quantity and returns `true`. This means encumbrances will **always** be reprocessed on piece creation, even if the cost update is a no-op edge case. In contrast, the delete counterpart (`PieceDeleteFlowPoLineService.poLineUpdateCost`) correctly returns `false` when quantity cannot be decremented.

This is likely intentional for creation (you always add 1), but it means any caller relying on the boolean to decide whether encumbrances changed will always trigger reprocessing. Worth confirming this is the desired behavior.

---

### ⚠️ High

#### H1. Batch `updateLocationsAndSavePoLineBatch` runs in parallel — possible race condition on shared `PoLine`

**File:** `src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowManager.java:87-93`

```java
private Future<Void> updateLocationsAndSavePoLineBatch(...) {
    var pieceCreationHolderList = createPieceCreationHolderList(holder);
    var futures = pieceCreationHolderList.stream()
      .map(h -> pieceCreateFlowPoLineService.updateLocationsAndSavePoLine(h, requestContext))
      .toList();
    return collectResultsOnSuccess(futures).mapEmpty();
}
```

Each `PieceCreationHolder` in the list shares the **same** `PoLine` object (set via `withPoLineOnly(holder.getPoLineToSave())`). Each call to `updateLocationsAndSavePoLine` mutates the locations on that shared `PoLine` and then calls `saveOrderLine`. Running these in parallel means:

- Multiple concurrent mutations to the same `List<Location>` (not thread-safe)
- Multiple `saveOrderLine` calls with intermediate/inconsistent state

Note that `updatePoLineBatch` correctly uses `chainCall` (sequential execution) for the cost phase. The location phase should likely also use `chainCall` for consistency and safety.

#### H2. `lineUpdated` flag on shared holder may be overwritten in batch flows

**File:** `src/main/java/org/folio/service/pieces/flows/BasePieceFlowUpdatePoLineService.java:54-55`

```java
boolean lineUpdated2 = poLineUpdateLocations(holder);
holder.setLineUpdated(holder.getLineUpdated() || lineUpdated2 || ...);
```

In the batch create flow, each `PieceCreationHolder` is a separate object (good), but they all reference the same `PoLine`. If `updatePoLine` (cost phase) sets `lineUpdated = true` on one holder, and then `updateLocationsAndSavePoLine` is called on a **different** holder for the same line, that second holder's `lineUpdated` starts as `false` — potentially skipping the save even though cost was updated. The `lineUpdated` state is per-holder, but the decision to save should be per-PoLine.

#### H3. Variable naming: `lineUpdated2`

**File:** `src/main/java/org/folio/service/pieces/flows/BasePieceFlowUpdatePoLineService.java:54`

`lineUpdated2` is a poor variable name. Consider `locationsUpdated` or `isLocationsChanged` to clarify intent.

---

### 🟡 Medium

#### M1. Tests do not assert Future result — fire-and-forget calls

**Files:** Multiple test files (e.g., `PieceCreateFlowPoLineServiceTest.java:435-436`, `PieceDeleteFlowPoLineServiceTest.java:763-764`)

```java
pieceDeleteFlowPoLineService.updatePoLine(incomingUpdateHolder, requestContext);
pieceDeleteFlowPoLineService.updateLocationsAndSavePoLine(incomingUpdateHolder, requestContext);
```

The tests call `updatePoLine(...)` and `updateLocationsAndSavePoLine(...)` **without** calling `.result()` on the returned `Future`. Previously the tests called `.result()` which blocks until the Future completes. Without `.result()`, the Future may not have completed before the assertions run. In Vert.x test contexts this can lead to:

- Tests passing even when the underlying operation fails
- Assertions checking state before asynchronous work completes

This affects **all** modified tests in both `PieceCreateFlowPoLineServiceTest` and `PieceDeleteFlowPoLineServiceTest`. The `.result()` calls should be restored, or the two calls should be chained: `updatePoLine(...).compose(v -> updateLocationsAndSavePoLine(...)).result()`.

#### M2. `PieceUpdateFlowPoLineService.poLineUpdateCost` is a no-op stub with FIXME

**File:** `src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowPoLineService.java:49-52`

```java
@Override
public boolean poLineUpdateCost(PieceUpdateHolder holder) {
    // FIXME: See MODORDERS-1435
    return false;
}
```

This is flagged as a known TODO. Ensure MODORDERS-1435 is tracked and prioritized. Since `poLineUpdateCost` returns `false`, the update flow will never process encumbrances through the new `updatePoLine` path — all encumbrance logic for piece updates is effectively disabled. Make sure this is acceptable for the current release scope.

#### M3. Deleted test coverage for `checkinItems=true` scenario in `PieceDeleteFlowManagerTest`

**File:** `src/test/java/org/folio/service/pieces/flows/delete/PieceDeleteFlowManagerTest.java` (removed test `shouldNotUpdateLineQuantityIfPoLineIsNotPackageAndManualPieceCreateTrueAndDeleteHoldingAndItemAndPiece`)

This test was removed from the manager-level test and replaced by a test in `PieceDeleteFlowPoLineServiceTest`. However, the manager-level test verified the **full orchestration flow** (including inventory deletion, title lookup, protection checks) for the `checkinItems=true` case. The new unit test only verifies the PoLine service in isolation. Consider whether integration-level coverage for this scenario is still needed.

#### M4. `skipUpdatingPoLine` check is duplicated between `updatePoLine` and `updateLocationsAndSavePoLine`

**File:** `src/main/java/org/folio/service/pieces/flows/BasePieceFlowUpdatePoLineService.java:33-34, 50-52`

Both methods independently check `skipUpdatingPoLine`. While this is safe (the guard is cheap), it means callers must remember to call **both** methods, and the skip logic is evaluated twice. If the two methods are always called as a pair, consider a single orchestrating method, or document that they must be called together.

---

### 🟢 Low

#### L1. `getLineUpdated()` naming convention

**File:** `src/main/java/org/folio/models/pieces/BasePieceFlowHolder.java:76`

For a `boolean` field, the conventional Java getter name is `isLineUpdated()`, not `getLineUpdated()`. This is a minor style issue but affects readability for other developers.

#### L2. Inconsistent `return true` in `PieceCreateFlowPoLineService.poLineUpdateLocations`

**File:** `src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowPoLineService.java:56-82`

`poLineUpdateLocations` always returns `true` regardless of which branch is taken (even when no location is actually added/updated). The `else` branch where no location exists and `isLocationUpdateRequired` returns `false` was removed — in the old code that branch updated cost only. Now, `poLineUpdateLocations` simply falls through and returns `true`, but it didn't actually modify any locations. This could cause unnecessary `saveOrderLine` calls.

---

## Summary Table

| ID | Severity | Category | File | Description |
|----|----------|----------|------|-------------|
| C1 | 🔥 Critical | Correctness | `PieceCreateFlowPoLineService` | `poLineUpdateCost` always returns `true` |
| H1 | ⚠️ High | Concurrency | `PieceCreateFlowManager` | Parallel location updates on shared `PoLine` |
| H2 | ⚠️ High | Correctness | `BasePieceFlowUpdatePoLineService` | `lineUpdated` flag not shared across batch holders |
| H3 | ⚠️ High | Maintainability | `BasePieceFlowUpdatePoLineService` | Poor variable name `lineUpdated2` |
| M1 | 🟡 Medium | Testing | Multiple test files | Missing `.result()` on Future — tests may pass vacuously |
| M2 | 🟡 Medium | Completeness | `PieceUpdateFlowPoLineService` | `poLineUpdateCost` is a no-op FIXME stub |
| M3 | 🟡 Medium | Testing | `PieceDeleteFlowManagerTest` | Removed integration-level test for `checkinItems=true` |
| M4 | 🟡 Medium | Design | `BasePieceFlowUpdatePoLineService` | Duplicated `skipUpdatingPoLine` guard |
| L1 | 🟢 Low | Style | `BasePieceFlowHolder` | `getLineUpdated()` should be `isLineUpdated()` |
| L2 | 🟢 Low | Correctness | `PieceCreateFlowPoLineService` | `poLineUpdateLocations` always returns `true` |

---

## Overall Assessment

The refactoring achieves its goal of separating cost/encumbrance updates from location/inventory updates, enabling the system to skip inventory processing when encumbrance changes are impossible. The architectural direction is sound.

**Key risks to address before merge:**
1. **M1 (Tests not awaiting Futures)** is the most actionable — tests may be silently passing without verifying anything
2. **H1 (Parallel batch mutations)** should be verified — if the shared `PoLine` is mutated concurrently, data corruption is possible
3. **H2 (lineUpdated flag scoping)** needs verification that the batch flow correctly propagates the "line was updated" signal across all pieces in a batch

📝 *Generated with Claude Code*

src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowManager.java
```diff
@@ -68,18 +71,23 @@ private Future<Void> processInventory(PieceCreationHolder holder, RequestContext
...
+  protected Future<Void> updatePoLineBatch(PieceBatchCreationHolder holder, RequestContext requestContext) {
+    return chainCall(holder.getPiecesToCreate(), piece ->
+      pieceCreateFlowPoLineService.updatePoLineCostAndProcessEncumbrances(createPieceCreationHolder(piece, holder), requestContext));
+  }
+
+  private Future<Void> updateLocationsAndSavePoLineBatch(PieceBatchCreationHolder holder, RequestContext requestContext) {
     var pieceCreationHolderList = createPieceCreationHolderList(holder);
     var futures = pieceCreationHolderList.stream()
-      .map(pieceCreationHolder -> updatePoLine(pieceCreationHolder, requestContext))
+      .map(pieceCreationHolder -> pieceCreateFlowPoLineService.updateLocationsAndSavePoLine(pieceCreationHolder, requestContext))
```
SerhiiNosko: what do you think about sequential update of pol ine here using method chainCall instead of parallel update as it implemented now?

damien-git: Good question, I have not looked into this part much. It is not optimized, the same po line is being saved several times (as before). And because it's done in parallel, it could potentially cause problems, although I haven't seen any in practice. I will try to find a better implementation.

src/main/java/org/folio/service/pieces/flows/BasePieceFlowUpdatePoLineService.java
```diff
@@ -29,16 +30,27 @@ protected BasePieceFlowUpdatePoLineService(PurchaseOrderStorageService purchaseO
...
   }
 
-  public Future<Void> updatePoLine(T holder, RequestContext requestContext) {
-    boolean isLineUpdated = poLineUpdateQuantity(holder);
+  public Future<Void> updatePoLineCostAndProcessEncumbrances(T holder, RequestContext requestContext) {
+    if (skipUpdatingPoLine(holder.getOriginPoLine())) {
+      return Future.succeededFuture();
+    }
+    boolean isLineUpdated = poLineUpdateCost(holder);
+    holder.setLineUpdated(isLineUpdated);
```
SerhiiNosko: this change overrides result of holder.setLineUpdated(true) in [code](https://github.com/folio-org/mod-orders/blob/MODORDERS-1415/src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java#L136) for updating po line receipt status

Suggested change:
```
    boolean costUpdated = poLineUpdateCost(holder);
    holder.setLineUpdated(holder.getLineUpdated() || costUpdated);
    if (costUpdated) {
      return receivingEncumbranceStrategy.processEncumbrances(holder.getPurchaseOrderToSave(), holder.getPurchaseOrderToSave(), requestContext);
    }
```

src/main/java/org/folio/service/pieces/flows/BasePieceFlowUpdatePoLineService.java
```diff
@@ -29,16 +30,27 @@ protected BasePieceFlowUpdatePoLineService(PurchaseOrderStorageService purchaseO
...
-        })
-        .compose(v -> purchaseOrderLineService.saveOrderLine(holder.getPoLineToSave(), getPieceLocations(holder), requestContext));
+      return receivingEncumbranceStrategy.processEncumbrances(holder.getPurchaseOrderToSave(), holder.getPurchaseOrderToSave(), requestContext);
+    }
+    return Future.succeededFuture();
+  }
+
+  public Future<Void> updateLocationsAndSavePoLine(T holder, RequestContext requestContext) {
+    if (skipUpdatingPoLine(holder.getOriginPoLine())) {
+      return Future.succeededFuture();
```
SerhiiNosko: should not return here, since we need to execute purchaseOrderLineService.saveOrderLine when po line is updated with new receipt status

src/main/java/org/folio/service/pieces/flows/update/PieceUpdateFlowManager.java
```diff
@@ -83,8 +83,10 @@ public Future<Void> updatePiece(Piece pieceToUpdate, boolean createItem, boolean
       .compose(aHolder -> basePieceFlowHolderBuilder.updateHolderWithTitleInformation(holder, requestContext))
       .compose(v -> asFuture(() -> defaultPieceFlowsValidator.isPieceRequestValid(pieceToUpdate, holder.getOriginPurchaseOrder(), holder.getOriginPoLine(), holder.getTitle(), 0, createItem)))
       .compose(title -> protectionService.isOperationRestricted(holder.getTitle().getAcqUnitIds(), UPDATE, requestContext))
+      .compose(v -> updatePoLineReceiptStatus(holder, List.of(holder.getPieceToUpdate()), requestContext))
+      .compose(v -> updatePoLineService.updatePoLineCostAndProcessEncumbrances(holder, requestContext))
```
SerhiiNosko, line 87: PUT /orders/pieces/{id} in Module Descriptor should be updated with 2 new module permissions:
"finance.funds.budget.item.get",
"finance.fiscal-years.item.get".

SerhiiNosko (approved):

PR #1291, opened 2026-04-24
[MODORDERS-1439] Changed order status automatic update logic

damien-git:
## Purpose
[MODORDERS-1439](https://folio-org.atlassian.net/browse/MODORDERS-1439) - Cancelling an invoice does not update order correctly

## Approach
Taking the current status of the order into account to decide whether to trigger an automatic status update based on a change to a po line status.

The following rules should be respected:
- A closed order should not be reopened when a line status is changed to final
- A closed order should not be reopened when the initial line statuses are non-final (because the order was not closed automatically)
- A closed order should be reopened when line statuses are changed from final to non-final
- An open order should not be closed when a line status is changed to non-final
- An open order should be closed when a status change makes both line statuses final

Similarly, when receiving or unreceiving a piece, the order status is only updated when it makes sense:
- When a piece is received, the order status is only updated if the order is not closed
- When a piece is unreceived, the order status is only updated if the order is not open

## Related PRs
- [mod-invoice](https://github.com/folio-org/mod-invoice/pull/617)
- [folio-integration-tests](https://github.com/folio-org/folio-integration-tests/pull/2295)

---

## Pre-Review Checklist

  - [ ] API/schema changes

### DIFF
--- a/src/test/java/org/folio/orders/utils/StatusUtilsTest.java
+++ b/src/test/java/org/folio/orders/utils/StatusUtilsTest.java
@@ -12,12 +12,15 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
 import org.folio.CopilotGenerated;
+import org.folio.rest.jaxrs.model.CompositePurchaseOrder;
 import org.folio.rest.jaxrs.model.Piece;
 import org.folio.rest.jaxrs.model.PoLine;
 import org.folio.rest.jaxrs.model.PurchaseOrder;
 import org.folio.service.orders.utils.StatusUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 @CopilotGenerated(partiallyGenerated = true)
 public class StatusUtilsTest {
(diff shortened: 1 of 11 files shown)


src/test/java/org/folio/orders/utils/StatusUtilsTest.java
```diff
@@ -57,6 +60,34 @@ void isStatusChanged_shouldReturnFalse_whenStatusNotChanged() {
     assertFalse(StatusUtils.isStatusChanged(compOrderLine, poLine1));
   }
 
+  @ParameterizedTest
+  @CsvSource(
+    value = {"Closed,Fully Paid,Awaiting Receipt,Awaiting Payment,Awaiting Receipt,false",
+      "Closed,Awaiting Payment,Fully Received,Awaiting Payment,Awaiting Receipt,false",
+      "Closed,Awaiting Payment,Awaiting Receipt,Fully Paid,Awaiting Receipt,false",
+      "Closed,Awaiting Payment,Fully Received,Awaiting Payment,Awaiting Receipt,false",
```
SerhiiNosko: Closed,Awaiting Payment,Fully Received,Awaiting Payment,Awaiting Receipt,false - duplicated row

SerhiiNosko (approved):

src/test/java/org/folio/orders/utils/StatusUtilsTest.java
damien-git: fixed

BKadirkhodjaev (approved):

Saba-Zedginidze-EPAM (approved):

PR #1311, opened 2026-05-18
[MODINVOICE-647] Silently ignore changes to nextPolNumber

damien-git:
## Purpose
[MODINVOICE-647](https://folio-org.atlassian.net/browse/MODINVOICE-647) - Invoice line number is reused after deletion

## Approach
Silently ignore changes to `nextPolNumber` (protected fields are only protected after order is opened).
This does not resolve the UI issue that it should get the latest version of an order after saving a line, before modifying it. But at least it won't affect line numbers anymore.

## Related PRs
- [mod-orders](https://github.com/folio-org/mod-orders/pull/1307)
- and others

---

### **Pre-Review Checklist**

  - [ ] API/schema changes

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderHelper.java
@@ -1,6 +1,7 @@
 package org.folio.helper;
 
 import static org.apache.commons.collections4.CollectionUtils.isEmpty;
+import static org.folio.orders.utils.HelperUtils.NEXT_POL_NUMBER;
 import static org.folio.orders.utils.HelperUtils.PO_LINES;
 import static org.folio.orders.utils.HelperUtils.REASON_CANCELLED;
 import static org.folio.orders.utils.HelperUtils.WORKFLOW_STATUS;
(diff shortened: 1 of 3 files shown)


SerhiiNosko (approved):

src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -338,6 +339,7 @@ public Future<Void> handleFinalOrderItemsStatus(PurchaseOrder purchaseOrder, Lis
 
   public JsonObject validateIfPOProtectedAndOngoingFieldsChanged(CompositePurchaseOrder compPO,
                                                                  JsonObject compPOFromStorageJson) {
+    compPO.setNextPolNumber(compPOFromStorageJson.getInteger(NEXT_POL_NUMBER)); // UIOR-1544 workaround; ignore any change of nextPolNumber
```
SerhiiNosko, line 342: now this looks like a good solution even without this UI ticket in comment

damien-git: I think it would be better to throw an exception when the value is changed, because it should not be changed by the API client and a change will fail to work. But this would make the Trillium UI implementation of "Save&Open" fail, so we have to keep it silent as a workaround. Hopefully we will implement optimistic locking eventually and it will trigger an error that UI will have to fix.

PR #1324, opened 2026-05-28
[MODORDERS-1449] Check invoice lines when opening order to set encumbrance status

damien-git:
## Purpose
[MODORDERS-1449](https://folio-org.atlassian.net/browse/MODORDERS-1449) - Encumbrance is released after un-opening the order with related Paid invoice (release encumbrance = false)

## Approach
When opening an order, check the related invoice lines to determine if encumbrances should be unreleased or released.

## Related PR
- [folio-integration-tests](https://github.com/folio-org/folio-integration-tests/pull/2393)

---

### **Pre-Review Checklist**

  - [ ] API/schema changes

### DIFF
--- a/src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
+++ b/src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
@@ -29,7 +29,8 @@ import org.folio.service.finance.budget.BudgetService;
 import static java.util.stream.Collectors.collectingAndThen;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
-import static org.folio.service.finance.EncumbranceUtils.allowEncumbranceToReleaseOnReopen;
+import static org.folio.rest.acq.model.finance.Encumbrance.Status.PENDING;
+import static org.folio.rest.acq.model.finance.Encumbrance.Status.RELEASED;
 
 public class EncumbranceRelationsHoldersBuilder extends FinanceHoldersBuilder {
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
```diff
@@ -165,9 +166,8 @@ private void mapHoldersToTransactions(List<EncumbranceRelationsHolder> encumbran
           .withAmountExpended(existingEncumbrance.getAmountExpended())
           .withAmountCredited(existingEncumbrance.getAmountCredited())
           .withAmountAwaitingPayment(existingEncumbrance.getAmountAwaitingPayment());
-        // release if encumbrance amountExpended + amountCredited + amountAwaitingPayment > 0 (if we have approved/paid invoices)
-        if (existingEncumbrance.getStatus() == Encumbrance.Status.RELEASED || allowEncumbranceToReleaseOnReopen(existingEncumbrance))
```
SerhiiNosko, line 169: method allowEncumbranceToReleaseOnReopen is not used now so can be deleted

src/main/java/org/folio/service/finance/transaction/PendingToOpenEncumbranceStrategy.java
```diff
@@ -72,4 +77,48 @@ public Future<List<EncumbranceRelationsHolder>> prepareProcessEncumbrancesAndVal
...
+  private Future<Void> withInvoiceWithReleaseEncumbrance(String fiscalYearId, List<EncumbranceRelationsHolder> holders, RequestContext requestContext) {
+    if (holders.isEmpty()) {
+      return Future.succeededFuture();
+    }
+    List<String> polIds = holders.stream()
+      .map(EncumbranceRelationsHolder::getPoLineId)
+      .distinct()
+      .toList();
+    return invoiceLineService.getInvoiceLinesByIdsAndQuery(polIds,
+        ids -> queryToGetRelatedInvoiceLinesByPoLineIds(fiscalYearId, polIds), requestContext)
```
SerhiiNosko: batching is broken here, instead of polIds need to use ids

src/main/java/org/folio/service/finance/transaction/EncumbranceRelationsHoldersBuilder.java
damien-git: It was used in a test, but I should remove that test. I might add another one too.

src/main/java/org/folio/service/finance/transaction/PendingToOpenEncumbranceStrategy.java
damien-git: I don't understand what is broken. We need to retrieve invoice lines by query for batches of pol ids.

SerhiiNosko: The queryFunction is invoked once per sublist, and its List<String> parameter is that sublist (the batch). But in the caller the lambda ignores that parameter and closes over the full polIds.

  So with, say, 30 pol ids → 2 sublists, you get:

  - batch 1: queryFunction([id1..id15]) → but convertIdsToCqlQuery is built from all 30 ids
  - batch 2: queryFunction([id16..id30]) → again built from all 30 ids

damien-git: Ooops, sorry, I get it now.

SerhiiNosko (approved):

PR #1332, opened 2026-06-11
[MODORDERS-1456] Auto-change order status like user

damien-git:
## Purpose
[MODORDERS-1456](https://folio-org.atlassian.net/browse/MODORDERS-1456) - When cancelling an invoice, order is reopened but encumbrance is not created

## Approach
- Split `PurchaseOrderHelper$updateOrder` to separate validation
- Call `updateOrderWithoutValidation()` when changing the order status automatically instead of a custom method which did not create encumbrances as needed
- Fixed a bug with approval by and approval date being reset when an order is opened
- Updated tests

## Related PR
- [folio-integration-tests](https://github.com/folio-org/folio-integration-tests/pull/2437)

## TODO
Investigate a way to pass user permissions for async processing with event handlers.

---

### **Pre-Review Checklist**

  - [ ] API/schema changes

### DIFF
--- a/src/main/java/org/folio/helper/PurchaseOrderHelper.java
+++ b/src/main/java/org/folio/helper/PurchaseOrderHelper.java
@@ -27,6 +27,7 @@ import static org.folio.service.orders.utils.StatusUtils.changeOrderStatusForOrd
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -208,7 +209,7 @@ public class PurchaseOrderHelper {
           throw new HttpException(RestConstants.VALIDATION_ERROR, errors);
         }
         return null;
-      }).compose(v -> updateOrder(compPO, deleteHoldings, requestContext))
+      }).compose(v -> updateOrderWithValidation(compPO, deleteHoldings, requestContext))
       .onSuccess(v -> logger.info("putCompositeOrderById :: Successfully updated order, id: {}", compPO.getId()))
       .onFailure(t -> logger.error("putCompositeOrderById :: Failed to update order", t));
   }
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/helper/PurchaseOrderHelper.java
```diff
@@ -612,4 +618,13 @@ private List<PoLine> getNonPackageLines(List<PoLine> poLines) {
     return poLines.stream().filter(line -> !line.getIsPackage()).toList();
   }
 
+  public void setOrderApprovalDetails(CompositePurchaseOrder compPO, RequestContext requestContext) {
```
SerhiiNosko: approval logic is not agreed fully, correct?

damien-git: From what Dennis said, my change is OK but he's going to look into potentially doing further changes later.

SerhiiNosko: can be private, since used only in this class

src/main/java/org/folio/orders/events/handlers/AbstractOrderStatusHandler.java
```diff
@@ -82,17 +78,12 @@ public void handle(Message<JsonObject> message) {
...
-      .compose(isStatusChanged -> {
-        if (Boolean.TRUE.equals(isStatusChanged)) {
-          return purchaseOrderHelper.handleFinalOrderItemsStatus(purchaseOrder, poLines, initialStatus.value(), requestContext)
-            .compose(aVoid -> purchaseOrderStorageService.saveOrder(purchaseOrder, requestContext))
-            .compose(purchaseOrderParam -> encumbranceService.updateEncumbrancesOrderStatusAndReleaseIfClosed(
-              convert(purchaseOrder, poLines), requestContext));
-        }
-        return Future.succeededFuture();
-      });
+    CompositePurchaseOrder poFromStorage = convert(purchaseOrder, poLines);
```
SerhiiNosko, line 88: po is not actually from storage, can rename to poBeforeStatusChange

SerhiiNosko (approved):

src/main/java/org/folio/orders/events/handlers/AbstractOrderStatusHandler.java
damien-git: it is obtained from storage just above

src/main/java/org/folio/helper/PurchaseOrderHelper.java
damien-git: right

